PR #11, opened 2026-03-26
Add KiCad 10.0 support to plugin installer

Pugnator:
- Add KiCad 10.0 to KICAD_VERSIONS and Python executable candidates
- Detect installed versions dynamically instead of hardcoding 9.0

### DIFF
--- a/install_plugin.py
+++ b/install_plugin.py
@@ -26,7 +26,7 @@ from startup_checks import get_cargo_version
 
 PLUGIN_NAME = "KiCadRoutingTools"
 PLUGIN_DISPLAY_NAME = "KiCad Routing Tools"
-KICAD_VERSIONS = ["9.0", "9.99"]  # 9.0 = stable, 9.99 = nightly
+KICAD_VERSIONS = ["10.0", "9.99", "9.0"]  # Checked in order; 9.99 = nightly
 
 
 def get_kicad_python() -> Path | None:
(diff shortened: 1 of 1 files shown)


drandyhaas:
Claude Code v2.1.84                                                                                                                                                                                                                   
Opus 4.6 (1M context) · Claude Max                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
❯ In the last merge, we added kicad 10.0 support for installation. Will that work for windows only, or all platforms?                                                                                                                            
                                                                                                                                                                                                                                                 
⏺ Bash(git show 9b8104d --stat)                                                                                                                                                                                                                  
  ⎿  commit 9b8104d159dfad80d28eeab9453dfceaefdfc5dd                                                                                                                                                                                             
     Author: Pugnator <ookami@mail.ru>                                                                                                                                                                                                           
     Date:   Thu Mar 26 19:30:41 2026 +0100                                                                                                                                                                                                      
     … +8 lines (ctrl+o to expand)                                                                                                                                                                                                               
                                                                                                                                                                                                                                                 
⏺ Bash(git show 9b8104d)                                                                                                                                                                                                                         
  ⎿  commit 9b8104d159dfad80d28eeab9453dfceaefdfc5dd                                                                                                                                                                                             
     Author: Pugnator <ookami@mail.ru>                                                                                                                                                                                                           
     Date:   Thu Mar 26 19:30:41 2026 +0100                                                                                                                                                                                                      
     … +46 lines (ctrl+o to expand)                                                                                                                                                                                                              
                                                                                                                                                                                                                                                 
⏺ The changes work across all platforms, but with a caveat:
                                                                                                                                                                                                                                                 
  - KICAD_VERSIONS list — updated to ["10.0", "9.99", "9.0"] — this is platform-independent and affects macOS/Linux/Windows equally for plugin directory discovery.                                                                              
  - Windows Python detection — new 10.0 paths were added for finding KiCad's bundled Python. This only affects Windows since macOS/Linux have separate code paths that use shutil.which("python3").
  - Plugin directory detection (get_kicad_plugin_dirs) — the old code hardcoded version == "9.0" as always-included. The new code dynamically checks whether a version is installed, but the install directory check (Path(rf"C:\Program         
  Files\KiCad\{version}")) is Windows-specific. On non-Windows platforms it falls back to checking version_dir.exists(), which is the config/plugin base directory — so it still works, just uses a different heuristic.                         
                                                                                                                                                                                                                                                 
  So it works on all platforms, though the "is this version installed?" detection is most robust on Windows (checks actual install path). On macOS/Linux it relies on the plugin config directory already existing, which should be fine for     
  versions the user has actually run.

PR #25, opened 2026-06-02
Respect KiCad keep-out rule areas in the routing obstacle map

bergmorten:
## What

Teach the grid router to honor KiCad **keep-out rule areas** (`(zone ... (keepout ...))`). Previously these were ignored, so the router could lay tracks/vias straight through a region the board designer marked as off-limits (e.g. an antenna-flange RF clearance).

## Changes

- **`kicad_parser.py`** — add `BoardInfo.keepouts` and `extract_keepouts()`, wired into `parse_kicad_pcb`. Each keepout is parsed to `{polygon, layers:set, tracks_allowed, vias_allowed}`.
- **`obstacle_map.py`** — add `add_keepout_obstacles()`, called from `build_base_obstacle_map`. Blocks track cells per-layer and via placement, gated on the keepout's tracks/vias-allowed flags. A keepout with no layer list applies to all routing layers. Point-in-polygon uses the same vectorized ray-cast as the existing cutout handling.

## Notes

Self-contained; no changes to routing call sites beyond the new obstacle pass. A keepout that allows both tracks and vias is a no-op.

---

## Maintainer review edits (amended onto this branch)

Reviewed against the codebase and amended the commit with the following improvements:

1. **Shared zone-block iterator (de-dup).** `extract_keepouts()` had copy-pasted the zone-scanning + paren-balancing loop from `extract_zones()`. Both now use a single `_iter_zone_blocks(content)` generator, so the scan logic lives in one place.

2. **Clearance is now actually applied.** `track_clear`/`via_clear` were computed but unused — only cell centres strictly *inside* the polygon were blocked, so track/via copper could still poke across the boundary. Now mirrors `_add_cutout_obstacles`: cells inside **or** within the configured clearance of an edge are blocked, keeping the copper (not just the centreline) out of the keep-out.

3. **Single `np.meshgrid`.** The grid was being built by two separate `np.meshgrid` calls (one per axis); now built once and unpacked.

4. **Bug fix — `vias not_allowed` was silently ignored.** The original `(keepout\b(.*?)\)` regex is non-greedy and stopped at the *first* `)`, so for the normal KiCad form `(keepout (tracks not_allowed) (vias not_allowed) ...)` it only ever saw the `tracks` sub-clause — `vias_allowed` always came back `True` and vias were never blocked. Replaced with balanced-paren extraction of the full keepout body.

5. **Docs.** Added a "Keep-out rule areas" feature bullet to `README.md` and listed it as an obstacle type (plus the `add_keepout_obstacles()` step) in `docs/routing-architecture.md`.

### Verification

- `extract_keepouts()` unit-tested against synthetic zones: plural `(layers ...)` and singular `(layer ...)`, and both `tracks`/`vias` flag combinations parse correctly.
- End-to-end against the built Rust `GridObstacleMap`: a single-layer keepout blocks tracks **and** vias inside the polygon and within clearance of the edge on its listed layer only, leaves other layers free, and leaves cells well outside the clearance free.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ A fast Rust-accelerated A* autorouter for KiCad PCB files. Compatible with **KiC
 - **Net ordering strategies** - MPS (crossing conflicts with diff pairs treated as units, shorter routes first using BGA-aware distance; uses segment intersection with MST for non-BGA boards), inside-out (BGA), or original order
 - **MPS layer swap** - When MPS detects crossing conflicts (nets in Round 2+), attempts layer swaps to eliminate same-layer crossings. Tries swapping both the conflicting Round 2 unit and Round 1 unit. Re-runs MPS after swaps to verify conflict resolution
 - **Board edge clearance** - Tracks and vias respect Edge.Cuts boundaries including curved edges (gr_arc), non-rectangular outlines, and interior cutouts. Arcs are automatically linearized into polylines for accurate clearance checking
+- **Keep-out rule areas** - Tracks and/or vias respect KiCad keep-out zones (rule areas with `(keepout (tracks not_allowed) (vias not_allowed) ...)`), blocking only the disallowed item on the keepout's listed layers (or all routing layers if none are listed). Track/via copper is kept clear of the boundary by the configured clearance, e.g. an antenna-flange RF exclusion
 - **BGA exclusion zones** - Auto-detected from footprints, prevents vias under BGAs
 - **Stub proximity avoidance** - Penalizes routes near unrouted stubs, with direction-aware costs (moving away from stubs costs less)
 - **BGA proximity avoidance** - Penalizes routes near BGA edges, with direction-aware costs (moving away from BGA zones costs less)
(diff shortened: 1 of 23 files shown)


drandyhaas:
Hi @bergmorten — thanks for this! I went ahead and amended your commit on this branch with a few maintainer edits (details now in the PR description):

- De-duped the zone-scanning loop into a shared `_iter_zone_blocks()` helper.
- Made the configured clearance actually apply, so track/via **copper** is kept out of the keep-out, not just the centreline.
- Collapsed the double `np.meshgrid` into one.
- **Bug fix:** the `(keepout (.*?))` regex was non-greedy and stopped at the first `)`, so `(vias not_allowed)` was silently ignored and vias were never blocked. Switched to balanced-paren extraction of the full keepout body.
- Added docs (README feature bullet + obstacle type in `docs/routing-architecture.md`).

I verified `extract_keepouts()` with unit tests and ran an end-to-end check against the built Rust `GridObstacleMap`, but that was all synthetic input on my end. Before we merge, could you please re-run this on a **real board with keep-out areas** (ideally one with both `tracks not_allowed` and `vias not_allowed`, and a multi-layer keepout) and confirm the router now respects them as expected? Want to make sure the amended behavior holds on actual KiCad files, not just my fixtures. Thanks!

PR #52, opened 2026-06-10
Claude tab: run AI skills headless from the plugin GUI (first slice of #40)

### LINKED ISSUE #34: Plugin GUI: add a Power Nets tab (patterns + widths), with a "generate Claude prompt" helper
## Summary

Power net routing (`--power-nets` / `--power-nets-widths`) is a major CLI feature with no GUI equivalent. Interactive users are forced into a two-tool workflow: run the CLI for power nets, then the GUI for signal nets.

## Proposed change

Add a **Power Nets** tab to the routing dialog (similar in spirit to the existing Advanced tab in `kicad_routing_plugin/swig_gui.py`):

- A list of pattern → width rows, e.g. `*GND*` → 0.4mm, `*VCC*` → 0.5mm, with add/remove buttons. First matching pattern wins, same as the CLI semantics in `routing_config.py`.
- Alternatively (or additionally): show the board's nets with checkboxes so the user can select specific nets and assign a width per selection, rather than writing glob patterns by hand. The net list with class grouping already exists in the GUI for net selection, so that widget pattern can be reused.
- Enforce the existing rule that power net widths are never smaller than the base track width.
- Persist via `settings_persistence.py`.

## Claude prompt helper

The repo already ships an AI-powered analysis skill (`/analyze-power-nets`, documented in `docs/power-nets.md`) that looks up component datasheets and generates ready-to-use `--power-nets` configurations.

### Preferred: spawn Claude Code directly

The plugin can invoke Claude Code headless and apply the result itself — no copy/paste round-trip:

```
claude -p "/analyze-power-nets /absolute/path/to/board.kicad_pcb — output the final recommendation as --power-nets / --power-nets-widths arguments on the last line" \
  --output-format json --allowedTools "Bash,Read,WebSearch"
```

- Run via `subprocess` with `cwd` set to the plugin install root (skill discovery is working-directory based), on a background thread so the wx GUI stays responsive, with a progress/cancel affordance (datasheet lookups can take minutes).
- Parse the `result` field of the JSON output for the `--power-nets "*GND*" "*VCC*" --power-nets-widths 0.4 0.5` string and fill the tab's pattern/width rows.
- Detect availability first: `claude --version` (installed) and `claude auth status` (exit code 0 = logged in).

### Fallback: copy prompt to clipboard

When the `claude` CLI isn't installed or authenticated, fall back to a **"Copy Claude prompt"** button that puts the same prompt on the clipboard for the user to paste into an interactive Claude Code session, plus a "Paste result" field that parses the returned argument string back into the tab.

### Working directory requirement

The skill is project-scoped: Claude Code only discovers `/analyze-power-nets` when launched from the KiCadRoutingTools directory (where `.claude/skills/` lives). So both the spawned subprocess and the manual fallback instructions must start there, and the board path must be **absolute** (the board file is generally not inside the repo):

```
cd /path/to/KiCadRoutingTools   # the plugin's install root
claude
> /analyze-power-nets /absolute/path/to/board.kicad_pcb
```

The plugin can compute both paths at runtime: its own install root from `os.path.dirname(__file__)` and the board path from pcbnew.

### PCM installs don't ship the skills

`package_pcm.py` whitelists what goes into the PCM zip (`PLUGIN_FILES_KEEP` / `PLUGIN_DIRS_KEEP`, lines ~54–71): `kicad_routing_plugin/`, `rust_router/`, and the root `.py` modules. `.claude/` is **not** included, so users who installed via the Plugin and Content Manager have no skill to invoke — even though the scripts the skill drives (`analyze_power_paths.py`, `list_nets.py`) *are* shipped.

Two options:

1. Add `.claude` to `PLUGIN_DIRS_KEEP` so PCM installs ship the skills, and have the spawned subprocess / generated prompt use the PCM plugins directory.
2. Make the prompt self-contained: inline the essential skill instructions in the prompt text (the heavy lifting lives in the shipped Python scripts anyway), so it works from any directory and doesn't depend on skill discovery at all.

Option 1 keeps a single source of truth for the skill logic; option 2 is more robust for users who have never cloned the repo. They can also be combined (prefer the skill when present, fall back to the inline prompt).

## Related

- #39 (Claude button for /find-high-speed-nets), #40 (Claude tab for /plan-pcb-routing). The CLI detection, subprocess runner, and result parsing should be shared plumbing across all three.

### LINKED ISSUE #39: Plugin GUI: Claude Code button for the /find-high-speed-nets skill (GND via distance)
## Summary

Companion to #34. The repo also ships the `/find-high-speed-nets` skill (`.claude/skills/find-high-speed-nets/`), which looks up component datasheets to classify nets by speed tier, estimates max frequencies and rise times, and recommends `--gnd-via-distance` values for GND return via placement. Like the power-nets skill, it has no GUI presence.

## Proposed change

Add a Claude Code button to the **Planes** tab (`kicad_routing_plugin/planes_gui.py`), next to the GND via controls, since the skill's main actionable output is the `--gnd-via-distance` recommendation:

- **"Ask Claude for GND via distance"** (or similar) button.
- Preferred behavior: spawn Claude Code headless and apply the result directly (see #34 for the invocation pattern):

  ```
  claude -p "/find-high-speed-nets /absolute/path/to/board.kicad_pcb — report the recommended --gnd-via-distance value on the last line as GND_VIA_DISTANCE=<value>" --output-format json --allowedTools "Bash,Read,WebSearch"
  ```

  run via `subprocess` with `cwd` set to the plugin install root (skill discovery is working-directory based), on a background thread so the wx GUI stays responsive. Parse the `result` field from the JSON output and fill the GND via distance field.
- Fallback when the `claude` CLI isn't installed/authenticated (detect with `claude auth status`, exit code 0 = logged in): copy the prompt to the clipboard for manual pasting, as in #34.

## Notes

- Datasheet lookups via WebSearch can take a few minutes — the button needs a progress/cancel affordance.
- Same packaging caveat as #34: `.claude/` is not currently included in the PCM zip (`package_pcm.py` `PLUGIN_DIRS_KEEP`), so either ship it or inline the skill instructions in the generated prompt.
- The full speed-tier report (not just the one number) is useful context — consider showing it in a scrollable text dialog with the recommendation pre-applied.

### LINKED ISSUE #40: Plugin GUI: Claude tab — generate /plan-pcb-routing prompt, run the plan, route the board
## Summary

The `/plan-pcb-routing` skill (`.claude/skills/plan-pcb-routing/`) is the project's flagship AI workflow: it analyzes the board, identifies fanout needs (BGA/QFN/QFP/PGA), detects diff pairs and length-matching groups, categorizes power nets, and produces a step-by-step routing plan. Today it's CLI-only (README "Option B").

Add a **Claude** tab to the plugin dialog that drives this skill end-to-end: spawn Claude Code headless to produce the plan, parse the plan into GUI parameters, and then **the GUI executes the routing steps itself, in-process on the live board**.

## Execution model: Claude plans, the GUI routes

Claude is used only for analysis and planning — it never runs routing commands itself. The plugin parses the suggested steps and runs them through its existing in-process routing machinery (the same code paths the Fanout / Differential / Config / Planes tabs already use).

Why this split:

- **No live-board conflict.** The CLI scripts modify the `.kicad_pcb` file while pcbnew holds the board in memory. The plugin's own tabs already route the live board in-process, so executing the plan through them sidesteps the file-vs-editor problem entirely (no temp-copy/import dance, no reload).
- **Minimal permissions.** Claude's headless run only needs `Read`, `Bash` for the read-only analysis scripts (`list_nets.py`, `analyze_power_paths.py`), and `WebSearch` for datasheets — never write access to the board.
- **Reviewable in familiar terms.** The plan lands in the GUI as parameter values on the existing tabs (or a step list referencing them), so the user reviews/tweaks settings in controls they already know instead of auditing shell commands.

### Structured plan output

Parsing free-form CLI suggestions out of prose would be brittle. Instead, the prompt should instruct the skill to emit the plan as machine-readable JSON, e.g.:

```json
{
  "steps": [
    {"action": "fanout", "kind": "bga", "component": "U1", "params": {"via_size": 0.45, "via_drill": 0.2}},
    {"action": "route_diff", "nets": ["*lvds*"], "params": {"track_width": 0.15, "diff_pair_gap": 0.15}},
    {"action": "route", "nets": ["*"], "params": {"power_nets": ["*GND*", "*VCC*"], "power_nets_widths": [0.4, 0.5]}},
    {"action": "route_planes", "nets": ["GND"], "params": {"plane_layers": ["In2.Cu"], "add_gnd_vias": true}}
  ]
}
```

This maps cleanly onto the GUI because the CLI and GUI already share one parameter space (`routing_defaults.py` / `GridRouteConfig`): each step is essentially "which tab + which field values". Validate the JSON against the known parameter names/ranges (the GUI ranges already exist in `routing_defaults.py`) and reject/flag anything unknown. Optionally also display the equivalent CLI commands for transparency/reproducibility.

### Invocation

```
claude -p "/plan-pcb-routing /absolute/path/to/board.kicad_pcb — do not execute anything; output the plan as JSON per the schema below: ..." \
  --output-format json --allowedTools "Read,WebSearch,Bash(python3 list_nets*),Bash(python3 analyze_power_paths*)"
```

- `subprocess` with `cwd` = plugin install root (skill discovery is working-directory based), on a background thread with progress/cancel (datasheet lookups take minutes).
- Detect availability first: `claude --version` / `claude auth status` (exit 0 = logged in); fall back to copy-prompt-to-clipboard as in #34.
- The board should be saved first so the analysis sees current state.

### UX flow

1. **Plan** — user clicks "Plan routing with Claude"; streaming/progress shown; plan arrives and is rendered as an ordered, editable step list with the proposed parameter values (and "apply to tab" so users can inspect each step in its native tab).
2. **Review** — user edits parameters, reorders, or unchecks steps.
3. **Execute** — GUI runs the checked steps sequentially in-process, reusing the per-tab progress reporting and cancellation that already exist.

### Optional: iterate on failures

After execution, failed nets / DRC results could be fed back to Claude by resuming the session (`claude -r <session_id> -p "these nets failed: ... suggest parameter adjustments as JSON"`), producing a revised plan for another review/execute round. This keeps the verify-and-retry loop, but with the GUI still doing all board mutation.

## Individual skills on the Claude tab (not just the full plan)

The full plan-everything flow is the headline, but the tab should also expose the focused skills individually — often you don't want a whole routing plan, just one analysis. With #42–#46 implemented, the repo has six skills; the tab becomes a small launcher:

| Button | Skill | Output handling |
|--------|-------|-----------------|
| Plan & route board | `/plan-pcb-routing` | Full plan → review → execute flow above |
| Review routed board | `/review-routed-board` | Show the pass/fail sign-off report in a results pane; offer "diagnose failures" follow-up when connectivity fails |
| Diagnose last failures | `/diagnose-routing-failures` | Needs the last run's log — the GUI should capture each routing run's output to a session log file precisely so this button has input. Parsed retry parameters can be applied to the relevant tab with one click |
| Identify diff pairs | `/identify-diff-pairs` | Confirmed pairs populate the Differential tab's net selection; suspected pairs shown for confirmation |
| Recommend stackup | `/recommend-stackup` | Report shown; resulting `--impedance`/plane-layer values offered as one-click fills |
| Analyze power nets | `/analyze-power-nets` | Fills the Power Nets tab (#34) |
| Find high-speed nets | `/find-high-speed-nets` | Fills `--gnd-via-distance` on the Planes tab (#39) |

Shared plumbing for all buttons (one implementation, reused): CLI detection (`claude --version` / `claude auth status`), subprocess runner with cwd = plugin root, background thread + progress/cancel, streaming log pane, copy-prompt-to-clipboard fallback when the CLI is unavailable.

### Recommend diagnosis in the failure popup

Independent of the Claude tab's buttons: when any routing run from any tab finishes with failed nets, the completion popup should mention the option — e.g. *"3 nets failed to route. The Claude tab's 'Diagnose failures' can analyze the log and suggest a targeted retry."* (Or the equivalent copy-prompt instructions when the `claude` CLI isn't installed.) This is the natural discovery point — users hit it exactly when they need the skill. Requires the per-run log capture mentioned above.

## Notes

- Same packaging caveat as #34/#39: `.claude/` must ship in the PCM zip (add to `PLUGIN_DIRS_KEEP` in `package_pcm.py`) or the prompts must inline the skill instructions.
- The skill itself may need a "plan-only, JSON output" mode added to `.claude/skills/plan-pcb-routing/SKILL.md` so the CLI workflow (where Claude does execute commands) and the GUI workflow (plan only) both stay supported.
- Related: #34 (power nets prompt helper), #39 (high-speed nets button), #42–#46 (the skills themselves). The Claude tab hosts the shared plumbing that #34/#39 reuse.

### LINKED ISSUE #53: New skill + Planes tab button: Claude recommends net -> plane-layer mappings
## Summary

The Planes tab requires the user to decide which nets get planes and on which copper layers — a judgment call that depends on the stackup, the power tree, and signal-integrity needs (GND reference adjacency for high-speed layers, plane capacitance between rails, splitting multiple rails across one layer, etc.). We already have the per-field Claude helpers from #34/#39; this issue adds the same treatment for plane assignments.

### 1. New skill: `/recommend-plane-mappings` (`.claude/skills/recommend-plane-mappings/`)

Analyzes a board and recommends net → plane-layer assignments:

- Read the stackup (copper layer count and which inner layers are available) and flag when the board has no inner layers to give planes to (2-layer boards → recommend pours/zones on F.Cu/B.Cu or suggest a 4-layer stackup, cross-referencing `/recommend-stackup`).
- Identify plane-worthy nets: GND plus the power rails (reuse the `/analyze-power-nets` approach — pad counts, datasheet current estimates — rather than name matching alone).
- Assign layers with SI rationale: GND adjacent to the primary signal layers for return paths, VCC/rail planes paired with GND for interplane capacitance, multiple low-current rails sharing a split layer where that's sound.
- Existing zones/assignments on the board should be respected/mentioned, not silently duplicated.

### 2. Machine-readable output

End the reply with one line the GUI can parse, e.g.:

```
RESULT=GND:In1.Cu;VCC|+3V3:In2.Cu
```

(assignment groups separated by `;`, nets within a group joined by `|` — matching `PlaneAssignmentPanel`'s `(nets_list, layer)` tuples — with the full reasoning in the report body). Multi-layer targets can repeat the net on another group.

### 3. Planes tab button

- **"Ask Claude for plane mappings"** next to the plane assignment panel in `kicad_routing_plugin/planes_gui.py`.
- Reuse the existing plumbing: `ClaudeSkillDialog` (live transcript, cancel), the Claude tab's model/effort selections via `get_claude_params`, and `extract_result_line` parsing.
- On success, validate (layers exist on the board, nets exist, no empty groups) and populate the assignment panel via `set_assignments()` — replacing or merging with existing rows (prompt the user if rows already exist).
- Unusable output is rejected and logged, fields untouched (same convention as #34/#39).

## Notes

- Depends on the PR #52 plumbing (`ClaudeSkillRunner` / `ClaudeSkillDialog`).
- The skill should be useful standalone from the CLI too (like the other skills), so write it board-path-first and document it in `docs/claude-skills.md`.
- Plane creation itself stays in-process via the existing Planes tab machinery — Claude only plans (same execution split as #40).

drandyhaas:
## Summary

Closes #40, closes #34, closes #39, closes #53.

First slice of #40 (groundwork for #34 / #39): a new **Claude** tab in the routing dialog that spawns Claude Code headless, streams live progress into the GUI, and parses a machine-readable result back into a field.

- **New `kicad_routing_plugin/claude_gui.py`** — `ClaudeTab` panel with a single test button (**Test: Recommend Stackup**) that runs `claude -p "/recommend-stackup <board>" --output-format stream-json --verbose` on a background thread (wx stays responsive), with pulse gauge, elapsed time, and working Cancel.
- **Live transcript** — each stream-json event renders as it arrives: a startup header (Claude Code version, model, cwd, discovered skills), tool calls (`-> Bash: Parse board stackup`), one-line tool results (`[ok]` / `[x]`), and assistant prose. stderr is drained on a side thread to avoid pipe-buffer deadlock.
- **Machine-readable result** — the prompt asks the skill to end with `RESULT=<n>`; the tab extracts it into a "Parsed result" field, proving the pattern #34/#39 will use to fill real GUI controls.
- **CLI discovery** — handles GUI-launched KiCad missing the shell PATH (checks `~/.claude/local`, `~/.local/bin`, homebrew paths); button disabled with install instructions when `claude` isn't found. `cwd` is the plugin root so `.claude/` skill discovery works (skills ship in the package since 619fd74).

## Test plan

- [x] Parser/renderer unit tests against real CLI event shapes (success, error, non-JSON, launch failure, missing fields)
- [x] Live end-to-end run through `_run_claude` (wx stubbed): events streamed, `RESULT=42` parsed
- [x] Full skill run against `kicad_files/lvds_converter_dualclk.kicad_pcb`: complete stackup report, `RESULT=4` parsed
- [x] Tested interactively in pcbnew: live transcript and parsed result confirmed working (2m47s run)

## Next steps (follow-up PRs)

1. #39: "Ask Claude for GND via distance" button on the Planes tab (`/find-high-speed-nets`)
2. #34: `/analyze-power-nets` button filling the Power Nets/Widths fields
3. #40 proper: `/plan-pcb-routing` JSON plan → editable step list → in-process execution


drandyhaas:
Added #39 on this branch: refactored `claude_gui.py` into a reusable `ClaudeSkillRunner` + `ClaudeSkillDialog`, and the Planes tab now has an **Ask Claude** button next to Max Distance that runs `/find-high-speed-nets` and fills the GND via distance field (clamped to the GUI range, enables the GND-vias checkbox, logs the action). Live-tested: the skill recommended `RESULT=3.0` for the LVDS sample board and the value lands in the control.

drandyhaas:
Added #34 on this branch: an **Ask Claude** button next to the Power Nets field on the Basic tab runs `/analyze-power-nets` (obeying the Claude tab's model/effort), parses `RESULT=--power-nets <patterns> --power-nets-widths <widths>` with validation, and fills the Power Nets / Power Widths fields. Live-tested on the LVDS sample board: 134s run with TI datasheet lookups returned `--power-nets "/VCC" "/GND" --power-nets-widths 0.5 0.5`, which parses and applies cleanly. The fields/persistence/min-width rule from the issue already existed; this adds the missing Claude helper.

drandyhaas:
Added the main #40 integration: **Plan Routing with Claude** on the Claude tab runs `/plan-pcb-routing` headless and gets back a JSON step plan. The plan fills each step's parameters into its native tab (Basic/Differential/Fanout/Planes) for review, and shows as a checkable step list. **Run Selected Steps** executes them sequentially through the tabs' own in-process machinery (`claude_plan.PlanExecutor`), re-applying each step's selection, switching tabs as it goes, and updating the list with live marks (`>` running, `[ok]` done, `[FAIL]` aborts). Stop halts between steps; completed steps uncheck so a stopped run resumes with the remainder.

Live-validated on the LVDS sample board: the real skill produced a clean 4-step plan (GND plane on B.Cu → diff pairs /CLK,/DATA @ 0.2/0.25 → route 22 nets with /VCC@0.5 power → plane re-pour) whose pair names exactly matched the GUI's own `find_differential_pairs` output, applying with zero validation warnings.

Issue #193, opened 2026-06-24
route_diff: support --layer-costs (per-layer routing cost), matching route.py / route_planes


drandyhaas:
## Summary
`route.py` and `route_planes.py` accept `--layer-costs` to bias routing toward/away from specific copper layers (e.g. `F.Cu=1 In1.Cu=3 ...`). `route_diff.py` has no equivalent — the flag is rejected (`unrecognized arguments`) and the differential-routing engine has no layer-cost support at all.

## Current state (verified)
- `route.py --help` → has `--layer-costs` (threaded to the engine at `route.py:1517`).
- `route_planes.py --help` → has `--layer-costs`.
- `route_diff.py --help` → **no** `--layer-costs`.
- `grep layer_cost diff_pair_routing.py` → **no matches** (the diff engine never takes/uses layer costs).

So this is a feature gap, not just an unexposed CLI flag.

## Motivation
On 4–8 layer boards you often want diff pairs to *prefer* certain layers — keep high-speed pairs on an inner layer adjacent to a reference plane (impedance/return path), or steer them off a congested `F.Cu`. Today there's no way to express that for diff pairs; they take whatever layer the router picks. A stress worker reached for `route_diff … --layer-costs …` by analogy with `route.py` and hit the wall.

## Proposed
Add `--layer-costs` to `route_diff.py` and thread it through `batch_route_diff_pairs` into the per-layer cost mechanism the single-ended path already uses (the Rust `GridRouter` supports per-layer proximity/cost, so the diff path can reuse it).

Per CLAUDE.md "keep CLI and GUI routing in sync", a new engine flag must be wired through **both** front-ends: the `route_diff.py` argparse layer **and** the GUI call site (`kicad_routing_plugin/differential_gui.py`), its config dict, the options panel, and `settings_persistence.py` — plus a default matching `route.py`'s (`F.Cu=1`, inner=`3`-style).

## Severity
Enhancement / route-quality — no correctness bug; adds control that exists for single-ended and plane routing but not diff pairs.

PR #207, opened 2026-06-26
feat(layer-costs): -1 = forbidden layer (obstacle-only, no routed copper); fixes #206

### LINKED ISSUE #206: Through-via barrels not checked against copper on non-routing (intermediate) layers in subset-layer routing
## Summary

When routing on a **subset** of copper layers (e.g. `route.py --layers F.Cu B.Cu` on a ≥4‑layer board), the engine can place a **through‑via** whose barrel passes the intermediate layers — but the obstacle map only blocks the via on the *routing* layers. Copper on the **non‑routing intermediate layers is never rasterized**, so a through‑via barrel can land on (short) an inner signal trace (e.g. `In2.Cu`).

## Root cause

In `obstacle_map.py`, `build_base_obstacle_map`:

- `num_layers = len(config.layers)` and the grid is built over **routing layers only**.
- The segment loop skips copper whose layer isn't in `config.layers`:
  ```python
  layer_idx = layer_map.get(seg.layer)
  if layer_idx is None:
      continue            # copper on a non-routing layer is never added
  ```
- `_add_via_obstacle` iterates `for layer_idx in range(num_layers)` — a placed through‑via is only ever marked on the routing layers, never on the layers its barrel actually spans (`via.layers` is parsed in `kicad_parser.py` but unused here).

So on a subset‑layer route the engine is blind to intermediate‑layer copper for via placement.

This is the **layer‑set analogue** of the already‑fixed via‑vs‑copper issues (#123, #174): those eliminated via/track overlaps on the **routing** layers; the intermediate‑layer barrel case remains because the obstacle map is built over `config.layers` only.

## Reproduction (deterministic, obstacle‑map level)

Build a 4‑layer board (`F.Cu, In1.Cu, In2.Cu, B.Cu`) with a foreign‑net track on `In2.Cu` and an identical control track on `F.Cu`, then build the base obstacle map for `--layers F.Cu B.Cu` and ask `is_via_blocked` over each track:

```
F.Cu  control via blocked : True     <- routing-layer copper correctly blocks vias
In2.Cu trace via blocked  : False    <- BUG: a through-via may be placed over the inner trace
In1   POUR  via blocked   : False    <- correct (a plane clears around foreign vias via antipad)
```

(A minimal `Segment`/`PCBData` script reproduces this directly against the Rust `GridObstacleMap`; happy to attach.) On a real 10‑layer board this surfaces end‑to‑end as KiCad DRC via↔inner‑trace shorts, e.g. `Via [..._P] (F.Cu–B.Cu) ↔ Spoor [..._N] on In2.Cu`.

## Proposed fix (offering a PR)

Add a pass that mirrors the via‑keepout path of `add_rule_area_keepout_obstacles`, but sourced from real copper: for each **segment** on a copper layer **not** in `config.layers` (and not a net being routed), block via cells along it via `add_blocked_vias_batch` (vias are global, so this is layer‑agnostic). Wire it into `build_base_obstacle_map` and the windowed builder; gate with `config.check_intermediate_via_collisions` (default on) + a `--no-intermediate-via-check` opt‑out.

Deliberately **segments only, not filled zones** — a copper pour clears around a foreign via (antipad) on re‑fill, so a via through an inner GND/PWR plane is not a short; blocking those would make via placement impossible on plane‑heavy boards.

Verified on a real 6‑layer board (`orangecrab_ext_pll`): with all layers routed the fix is **byte‑identical** to stock (inert); on a `F.Cu/B.Cu` subset it is **purely additive** (stock ⊆ fixed, 0 cells removed) and adds exactly the inner‑trace via keep‑outs. PR incoming.

edgehero:
## What

Adds a forbidden-layer mode to `--layer-costs`: a value of **`-1`** for a layer means *"keep this layer in the obstacle map (its copper blocks via barrels, and through-vias may span it) but place **no routed track copper** on it."* Any other value keeps the existing `[1.0, 1000]` cost-multiplier meaning.

This re-scopes the PR to the approach you suggested on #206 (thanks for the steer). Intended usage — pass **all** board copper layers in `--layers` for full obstacle awareness, and forbid the inner ones via cost:

```
--layers F.Cu In1.Cu In2.Cu B.Cu --layer-costs 1.0 -1 -1 3.0
```

Tracks route only on F.Cu/B.Cu; In1/In2 stay obstacles and via spans.

## Why this supersedes the original obstacle-map patch

The first version of this PR added `add_intermediate_copper_via_obstacles` to make a **subset**-layer route safe. As you noted, passing **every** layer in `--layers` already makes `build_base_obstacle_map` rasterize all copper (including the global via keep-out), so a through-via barrel can't short an inner trace — the same #206 protection, natively. The special-case becomes redundant (its `inter = copper_layers - layers` set is empty when all layers are passed), so this PR drops it and solves #206 through the layer-cost mechanism instead.

## Why a Rust change, not just a high cost

The move-cost path is unclamped, so a large *positive* cost is only a **soft** avoid — A* still routes onto the layer when cornered. A hard "never place copper here" needs the successor skipped:

- **`rust_router/router.rs` + `pose_router.rs`** — skip any A* successor that would end a via on (G1) or lateral-move/turn on (G2) a layer whose cost `< 0`. A via is one direct edge to its target layer, so through-vias keep **spanning** a forbidden middle layer for free.
- Sanitize the sentinel out of `min_layer_cost` (start penalty + admissibility heuristic) and the via layer-transition delta, so a negative cost can never underflow the cost math. Applied across `route_multi`/`route_with_frontier` and `route_pose`/`route_pose_with_frontier`.

## Python

- `routing_constants.FORBIDDEN_LAYER_COST = -1`.
- `routing_config.get_layer_costs()` emits `-1` **verbatim** (not `int(-1*1000) = -1000`, which the engine would read as a *discount*).
- `route.py` / `route_diff.py` / `route_planes.py`: `--layer-costs` validation accepts `-1`; `--help` updated.

## Tests

`tests/test_forbidden_layer_grid.py` + `tests/test_forbidden_layer_pose.py`: a hard ban (zero copper on the forbidden layer, where a control proves the cost bias *would* have used it), a through-via that spans the forbidden middle layer, and a forbidden target/source that is unreachable / lays no copper. Built from source (`build_router.py --from-source`) and run green; the existing `test_pose_layer_costs.py` (#193 biasing) still passes.

Verified end-to-end on a real barrel-vs-inner-copper repro (a F.Cu→B.Cu net whose via must clear a foreign trace on an inner layer): with `--layers F.Cu In1.Cu In2.Cu B.Cu --layer-costs 1.0 -1 -1 1.0` the via detours exactly to the clearance edge of the inner copper and no track lands on In1/In2; the old subset route drops the via straight onto the inner trace.

Fixes #206.

### DIFF
--- a/route.py
+++ b/route.py
@@ -86,7 +86,7 @@ from routing_common import (
 import routing_defaults as defaults
 import re
 from terminal_colors import RED, RESET
-from routing_constants import DEFAULT_4_LAYER_STACK, POWER_NET_EXCLUSION_PATTERNS
+from routing_constants import DEFAULT_4_LAYER_STACK, POWER_NET_EXCLUSION_PATTERNS, FORBIDDEN_LAYER_COST
 
 # Import Rust router (startup_checks ensures it's available and up-to-date)
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'rust_router'))
@@ -270,12 +270,14 @@ def batch_route(input_file: str, output_file: str, net_names: List[str],
         else:
             layer_costs = [1.0 if layer == 'F.Cu' else 3.0 for layer in layers]
 
-    # Validate layer costs are in range [1.0, 1000]
+    # Validate layer costs: -1 = forbidden (no copper placed; still an obstacle),
+    # otherwise a multiplier in [1.0, 1000].
     for i, cost in enumerate(layer_costs):
-        if cost < 1.0 or cost > 1000:
+        if cost != FORBIDDEN_LAYER_COST and (cost < 1.0 or cost > 1000):
             layer_name = layers[i] if i < len(layers) else f"layer {i}"
             from routing_exceptions import ConfigurationError
-            raise ConfigurationError(f"Layer cost for {layer_name} must be between 1.0 and 1000, got {cost}")
+            raise ConfigurationError(f"Layer cost for {layer_name} must be -1 (forbidden) or "
+                                     f"between 1.0 and 1000, got {cost}")
 
     costs_str = ', '.join(f"{layers[i]}={layer_costs[i]}x" for i in range(min(len(layers), len(layer_costs))))
     print(f"  Layer costs: {costs_str}")
@@ -1331,8 +1333,11 @@ For differential pair routing, use route_diff.py:
 
     # Layer preference options
     parser.add_argument("--layer-costs", nargs="+", type=float, default=[],
-                        help="Per-layer cost multipliers (1.0-1000, default: F.Cu=1.0, others=3.0). "
-                             "Order matches --layers. Example: --layer-costs 1.0 5.0")
+                        help="Per-layer cost multipliers (1.0-1000, default: F.Cu=1.0, others=3.0), "
+                             "or -1 = FORBIDDEN: the layer stays an obstacle (its copper blocks vias) "
+                             "and through-vias may span it, but NO track copper is placed on it. "
+                             "Order matches --layers. Example (route only F.Cu/B.Cu, In1/In2 obstacle-"
+                             "only): --layers F.Cu In1.Cu In2.Cu B.Cu --layer-costs 1.0 -1 -1 3.0")
 
     # Debug options
     parser.add_argument("--debug-lines", action="store_true",
--- a/route_diff.py
+++ b/route_diff.py
@@ -220,11 +220,13 @@ def batch_route_diff_pairs(input_file: str, output_file: str, net_names: List[st
     # --layer-costs is given; unlike route.py there is no 2-layer F.Cu/B.Cu default.
     if not layer_costs:
         layer_costs = [1.0] * len(layers)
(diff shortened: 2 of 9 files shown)


drandyhaas:
Merged into `main` manually (rebased to keep linear history, your authorship preserved) as commit 59656d8, with a follow-up 47cc15a that relaxes the CLI/GUI validation to accept **any** negative `--layer-costs` value as forbidden (folded to the canonical `-1`), since the Rust engine already treats every negative identically — plus a tooltip/help refresh and the rust crate bump to 0.17.1.

Reviewed and verified before merge: byte-identical routing geometry when no negative cost is passed (feature inert), forbidden layers carry zero routed copper while through-vias span them, and `-1/-2/-1000/-999999` all behave identically. Thanks @edgehero — clean PR, and the `-1`-as-hard-skip insight (vs a large finite cost) was the right call.

Issue #243, opened 2026-06-30
Route congested-BGA multipoint diff pairs via per-leg hybrid (CK1) — layer scheme + leg-endpoint TODOs


drandyhaas:
## Context

A multipoint differential pair whose coupled escape can't **launch** from a congested BGA is currently deferred to single-ended, even when a clean coupled route exists. Concrete case: **butterstick `/FPGA-DDR3L/CK1`** (DDR3 bank, chain `U4:J19/K19 -> U12:J7/K7 -> R71:2`). The standard multipoint leg re-escapes the *coupled pair* from the pads, but the DDR3 bus through-via column (CS0 right of J19, ODT0 right of K19) leaves only 0.35 mm — a coupled pair needs ~0.48 mm — so "no valid source position." Each net escapes fine *individually* (the fanout already did), so the pair should route each leg like a **2-terminal hybrid**: a coupled middle + single-ended escape legs.

A hand-routed proof that it's feasible (and **not** congested near U12) is in `~/Downloads/butterstick_diffpair_final.kicad_pcb`: both CK1 nets run **coupled on B.Cu** (P y=105.6, N y=105.9, 0.3 mm gap) in an **arc** from U4 down-and-around to ~(151.5, 105.7) *below* U12, then each net **single-ends up** to its split U12 stub tip (P tip 151.8,107.875; N tip 153.375,107.875 — 1.575 mm apart).

## Work done (branch `worktree-multipoint-leg-hybrid`, base main `f675731`, head `c0f0d1b`)

Foundation that **works for the coupling half**:
- `_route_direct_coupled_middle` gains `endpoints=(src,tgt)` (route one leg's two terminals as a 2-terminal hybrid).
- `_terminal_stub_endpoint` (diff_pair_multipoint): launch from the fanout **stub tips** (clear of the BGA) — the pad midpoint is buried inside the BGA, blocked on every layer.
- "couple as far as possible" relaxation: shorten the coupled middle toward the other end until it reaches open copper.
- multipoint `_route_chain_attempt`: on leg failure, fall back to the per-leg hybrid.

With this the **coupled middle routes** (U4 end works). The pair still defers because of the two TODOs below. Existing hybrid + multipoint tests pass (`test_watchy_diff_hybrid_escape`, `test_multipoint_diff_route`, `test_diff_pair_route`).

## TODO 1 — layer scheme (for BOTH 2-terminal and multipoint-leg hybrid)

[...]

Issue #224, opened 2026-06-27
route/repair tracks laid below clearance to a foreign component pad (PAD-SEGMENT)


drandyhaas:
## Summary

From the 45-board no-LLM stress redo (HEAD `05b22a8`), **20 PAD-SEGMENT violations**
remain that are not covered by #221 (route_diff vias), #222 (fanout cap-pad vias) or
#223 (fanout track width): routed signal/reconnect tracks — and GND-plane repair
tracks — placed below clearance to a **different net's component pad**.

## Evidence (DRC at routed clearance)

| board | count | example |
|---|--:|---|
| ottercast_audio | 12 | `Pad:GNDA (MK1.1) <-> Seg:Net-(C63-Pad2)` (0.136mm), `Pad:Net-(C3-Pad1) <-> Seg:Net-(R81-Pad1)` |
| ddr5_testbed | 4 | `Pad:LBDQ (J1.209) <-> Seg:GND` (0.069mm) — GND-plane track grazing a connector pad |
| hackrf_one | 2 | `Pad:/mcu/usb/power/U0_TXD (P25.5) <-> Seg:/mcu/usb/power/P2_13` |
| cynthion | 2 | `Pad:/AUX Port/AUX PHY/CLK (U11.21) <-> Seg:.../AUX_PHY.~{RESET}` |

These are different-net track-to-pad clearance violations on outer layers (F.Cu).
Two distinct producers:
- **route.py / reconnect** signal tracks routed too close to a neighbouring pad
  (hackrf, cynthion, ottercast's C3/R81).
- **plane repair** (route_disconnected_planes) GND tracks grazing a foreign pad
  (ddr5 `Seg:GND <-> LBDQ`, ottercast `GNDA` mounting pad).

## Why it matters

- The router's obstacle map should treat a foreign pad as a keep-out at the working
  clearance; these show it isn't always (overlaps up to 0.136mm).
- Small count, but a real different-net clearance class distinct from the via and
  width issues already filed.

## Suggested direction

Ensure the pad keep-out (foreign pads, including mounting/test pads like MK1) is
inflated by the working clearance in both the signal-route and plane-repair obstacle
maps, the same way via/track clearances are. Note the per-pad duplication in the
report (same pad/segment counted ~4×) — likely one violating segment split across
grid cells; worth confirming the dedup in check_drc as well.

## Relationship

- #221 / #222 — via-clearance classes; this is the **track-to-pad** class.
[...]

PR #245, opened 2026-06-30
feat(#196): end-to-end length matching for AC-coupled differential pairs

### LINKED ISSUE #196: route_diff: AC coupled diff pairs split by series caps are length matched per side, not end to end
### Summary
When a differential pair is AC coupled through series capacitors, it is split into two separate base named pairs, one on each side of the caps. `route_diff` pairs them by base name and length matches each side independently. It has no concept that the two pairs are one electrical signal, so the end to end skew of the coupled pair (what the receiver actually sees) is neither reported nor matched.

This is the common case for PCIe, USB3, SATA and similar links, where every high speed pair has series DC blocking caps.

### Minimal repro
A synthetic board: two 2 pin connectors and two 0402 caps, one pair split across the caps so the P side is `J1.1 -> C1 -> J2.1` and the N side is `J1.2 -> C2 -> J2.2`. Build with KiCad bundled python (point `LIB` at your footprints dir):

```python
import pcbnew
LIB = "/usr/share/kicad/footprints"
HDR = pcbnew.FootprintLoad(LIB + "/Connector_PinHeader_2.54mm.pretty", "PinHeader_1x02_P2.54mm_Vertical")
CAP = pcbnew.FootprintLoad(LIB + "/Capacitor_SMD.pretty", "C_0402_1005Metric")
b = pcbnew.BOARD(); b.SetCopperLayerCount(2)
def place(fp, ref, x, y):
    f = pcbnew.FOOTPRINT(fp); f.SetReference(ref)
    f.SetPosition(pcbnew.VECTOR2I(pcbnew.FromMM(x), pcbnew.FromMM(y))); b.Add(f); return f
J1 = place(HDR, "J1", 10, 15); C1 = place(CAP, "C1", 18, 13)
C2 = place(CAP, "C2", 18, 17); J2 = place(HDR, "J2", 26, 15)
def net(n):
    ni = pcbnew.NETINFO_ITEM(b, n); b.Add(ni); return ni
N = {k: net(k) for k in ("DPA_P", "DPA_N", "DPB_P", "DPB_N")}
J1.Pads()[0].SetNet(N["DPA_P"]); J1.Pads()[1].SetNet(N["DPA_N"])
C1.Pads()[0].SetNet(N["DPA_P"]); C1.Pads()[1].SetNet(N["DPB_P"])  # P side caps
C2.Pads()[0].SetNet(N["DPA_N"]); C2.Pads()[1].SetNet(N["DPB_N"])  # N side caps
J2.Pads()[0].SetNet(N["DPB_P"]); J2.Pads()[1].SetNet(N["DPB_N"])
for x0, y0, x1, y1 in [(5,5,31,5),(31,5,31,25),(31,25,5,25),(5,25,5,5)]:
    s = pcbnew.PCB_SHAPE(b); s.SetShape(pcbnew.SHAPE_T_SEGMENT)
    s.SetStart(pcbnew.VECTOR2I(pcbnew.FromMM(x0), pcbnew.FromMM(y0)))
    s.SetEnd(pcbnew.VECTOR2I(pcbnew.FromMM(x1), pcbnew.FromMM(y1)))
    s.SetLayer(pcbnew.Edge_Cuts); b.Add(s)
pcbnew.SaveBoard("cap_chain.kicad_pcb", b); print("wrote cap_chain.kicad_pcb")
```

then:

```
python route_diff.py cap_chain.kicad_pcb out.kicad_pcb \
  --nets DPA_P DPA_N DPB_P DPB_N --layers F.Cu B.Cu \
  --track-width 0.15 --diff-pair-gap 0.25 --diff-pair-intra-match --no-fix-polarity
```

Observed (the two sides are matched separately):

```
Analyzing layer swaps for 2 diff pair(s)...
[1/2] Routing diff pair DPA
[2/2] Routing diff pair DPB
Intra-pair P/N length matching
    P/N intra-pair: already matched (delta=0.059mm <= 0.1mm)
    P/N intra-pair: already matched (delta=0.059mm <= 0.1mm)
"routed_diff_pairs": ["DPA", "DPB"]
```

DPA (the J1 to caps side) and DPB (the caps to J2 side) are each matched to their own intra pair tolerance. There is no match of the concatenated P path (`DPA_P + DPB_P`) against the N path (`DPA_N + DPB_N`).

### Where it bites on a real board
On a 10 layer carrier, a PCIe TX pair runs SO-DIMM to two AC caps to M.2 in a tight corridor. route_diff matched each side but could not fit meanders on either:

```
P/N intra-pair: P=76.441mm, N=77.960mm, delta=1.519mm, adding meanders to P
P/N intra-pair: could not fit meanders
P/N intra-pair: P=14.341mm, N=14.877mm, delta=0.535mm, adding meanders to P
P/N intra-pair: could not fit meanders
```

The end to end skew came out at 2.054 mm, which is exactly the sum of the two per leg deltas (1.519 + 0.535). Because matching is per side, it never considered putting the compensating length on whichever side had room.

### Suggested approach
The chain is auto detectable from the board, the same way diff pairs already are. A 2 pad component (the AC cap) bridging DPA_P and DPB_P, plus another bridging DPA_N and DPB_N, identifies the series coupled pair. Detecting that and length matching the concatenated P versus N path (free to place the compensating meander on either side) would manage the real end to end skew. Related to #175.

### Versions
KiCadRoutingTools at `a017e74` (main, the #195 fix), KiCad 10.0, Windows.

edgehero:
## Sorry up front, this one's bigger than I'd like 🙏

I normally try hard to keep PRs small and surgical, and I know a large diff is more to review. I genuinely couldn't find a smaller way to do #196 *correctly*: detecting the AC-coupling topology, matching the concatenated path end-to-end, and keeping the CLI / GUI / settings in parity are interdependent, so a partial version would be either wrong or misleading. To make it as reviewable as possible I've split it into single-purpose commits and kept the whole feature **provably inert when the flag is off**. Thanks for bearing with me.

## What this does

Fixes #196. When a differential pair is AC-coupled through series DC-blocking caps it splits into two base-named pairs : `DPA_P/DPA_N` (driver→caps) and `DPB_P/DPB_N` (caps→receiver). `route_diff` length-matches each side's P/N independently and has no concept that they're one signal, so the end-to-end skew the receiver actually sees :`len(DPA_P)+len(DPB_P)` vs `len(DPA_N)+len(DPB_N)` is neither reported nor matched.

This adds the industry-standard **extended-net / XNet** treatment behind a new `--ac-couple-match` flag (off by default):

1. **Detect** the cap chain (`diff_xnet.py`): a 2-pad cap bridging `A_P↔B_P` plus a partner bridging `A_N↔B_N` joins two pairs into one XNet. Conservative symmetric bridges only, signal nets only, **DNP/no-pop caps treated as open circuits and skipped**, multi-cap chains supported, asymmetric junctions warned and left to per-side behavior.
2. **Match end-to-end** (`apply_ac_coupled_length_matching`): measure the concatenated P vs N conductor (polarity/stub-aware), then place the compensating meanders on whichever segment has room, preferring the segment where it also reduces local intra-pair skew and **report the end-to-end skew** in `JSON_SUMMARY` as `ac_coupled_xnets`. For its member pairs it supersedes per-side intra-pair matching so the two passes don't fight.
3. Reuses the existing meander engine via a shared `_lengthen_net_with_meanders()` core refactored out of `apply_intra_pair_length_matching()` (behavior unchanged). no parallel matching subsystem.

Wired through the CLI (`--ac-couple-match`), the Differential GUI tab + `settings_persistence`, with docs in `length-matching.md` and `api-routing-config.md`.

## ⚠️ One unrelated fix is bundled in (happy to split it out)

While testing I hit a **pre-existing import-breaking bug on `main`**: `pcb_modification.py` (from `2148f33`, `fix(#224)`) annotates a parameter `List[Dict]` but never imports `Dict`, so **`route_diff` fails to import under any Python** (`NameError: name 'Dict' is not defined`). CI didn't catch it because it doesn't run the functional tests. The first commit here is the one-word fix; nothing else works without it. Glad to move it to its own PR if you prefer.

## Inertness (byte-identical when off)

With the flag off, every new path is skipped and the only un-gated change (the intra-pair refactor) is logic-preserving. I verified the routed **geometry is byte-identical** to clean `main` + the `Dict` fix by parsing both outputs and comparing segment/via multisets — MD5 is meaningless here since KiCad emits fresh random track UUIDs each run.

## Test results

Run under a Python with the router + numpy/scipy/shapely:

```
test_ac_coupled_match            PASS   detection ×7 + concat/polarity measurement + integration + control
test_diff_pair_route             PASS   5/5 pairs
test_diff_pair_detection         PASS
test_diff_layer_costs_response   PASS   (routes pairs → exercises the shared meander refactor)
test_multipoint_diff_route       PASS
run_doc_examples                 27 run, 5 failed   (all pre-existing missing-fixture cases; was 6 before
                                                    the Dict fix — this PR recovers one)
```

On the issue's repro board (`kicad_files/cap_chain.kicad_pcb`, built by the script in #196), `--ac-couple-match` now surfaces the end-to-end skew `main` never reported:

```
AC-coupled XNet: DPA+DPB (coupled via C1, C2)
AC-couple DPA+DPB: end-to-end P=20.089mm, N=19.974mm, delta=0.115mm (coupled via C1, C2)
JSON_SUMMARY: {... "ac_coupled_xnets": [{"nets": "DPA+DPB", "skew_mm": 0.1148, "bridges": ["C1", "C2"]}]}
```

On this near-symmetric synthetic board the residual is a fixed ~0.115 mm pad/stub asymmetry just over the 0.1 mm tolerance, so it correctly reports "could not fit" rather than faking a match — the multi-mm skew in the issue came from real-board congestion. The meander path itself is the same one intra-pair uses, exercised by the routing tests above.

## Background / references

The end-to-end "extended net" approach and the signal-integrity rationale are well established; the design follows these:

- **Altium — Defining High-Speed Signal Paths with xSignals** — an xSignal is two nets joined by a series component (incl. capacitors); Length / Matched-Length rules span the whole path. https://www.altium.com/documentation/altium-designer/defining-high-speed-signal-paths-with-xsignals
- **Cadence / OrCAD X — XNet (eXtended Net)** — a net joined to another across a series passive; Constraint Manager bundles the segment lengths into one value. https://resources.pcb.cadence.com/blog/2024-orcad-x-create-remove-xnet-step-by-step-guide-cadence
- **TI SPRAAR7E — High-Speed Interface Layout Guidelines** — intra-pair skew budgets (5 mil for USB3/SATA/PCIe). https://www.ti.com/lit/an/spraar7e/spraar7e.pdf
- **Samtec, DesignCon 2024 — "Comparing the Different Metrics of Intra-Pair Skew in Tracking Channel Performance"** — residual skew converts differential energy to common mode; why end-to-end skew is the receiver-relevant metric; caution against blindly zeroing it. https://suddendocs.samtec.com/notesandwhitepapers/samtec-dc24-paper-comparing-the-different-metrics-of-intra-pair-skew-in-tracking-channel-performance.pdf
- **Intra-Pair Skew Propagation Graph (ISPG)**, arXiv 2026 (preprint) — total channel skew = sum of per-segment contributions, i.e. the formal basis for concatenating across the caps. https://arxiv.org/abs/2605.08570
- **Obstacle-Aware Length-Matching Routing for Any-Direction Traces**, DAC 2024 — confirms academic length-matchers are purely geometric and don't model cap-split topology, so the XNet has to be a constraint layer above the meandering. https://arxiv.org/abs/2407.19195
- KiCad 8/9/10 has no native cross-series-component tuning (KiCad 10 added delay tuning, not xnets — tracked at https://gitlab.com/kicad/code/kicad/-/issues/18118), so this is built into the tool.

## Known limitations / follow-ups

- Length-only (mirrors intra-pair); a delay-aware variant and routing the pair *coupled through* the cap break (via the #243 hybrid) are natural follow-ups.
- Exactly-2-pad caps only — 4-pad arrays / multi-pair packages fall back to per-side matching + a warning.
- Pad-escape length isn't counted (inherited limitation from intra-pair matching).

### DIFF
--- a/diff_xnet.py
+++ b/diff_xnet.py
@@ -1,0 +1,145 @@
+"""
+AC-coupled differential-pair (XNet) detection -- issue #196.
+
+A high-speed differential pair (PCIe / USB3 / SATA / ...) is commonly split into
+two separately-named pairs by series DC-blocking capacitors:
+
+    DPA_P --[C1]-- DPB_P        (the full P conductor: driver -> caps -> receiver)
+    DPA_N --[C2]-- DPB_N        (the full N conductor)
+
+``route_diff`` sees ``DPA_*`` and ``DPB_*`` as two independent pairs and
+length-matches each side's P/N on its own, so the end-to-end skew the receiver
+actually sees (the whole P path vs the whole N path) is never matched. This
+module detects that topology -- the industry "extended net" / XNet -- so the
+matcher can treat the concatenated P and N conductors as one length budget.
+
+Detection is a pure, side-effect-free read of the parsed board.
+"""
+from dataclasses import dataclass, field
+from typing import Dict, List, Tuple
+
+from kicad_parser import PCBData
+from routing_config import DiffPairNet
+
+
+@dataclass
+class AcCoupledXNet:
+    """A differential signal whose P and N conductors are each split across one
+    or more series 2-pad parts (DC-blocking caps), joining 2+ ``DiffPairNet``s
+    into one end-to-end pair.
+
+    Attributes:
+        members: the per-segment pairs that make up the chain (2 or more).
+        bridge_refs: component references of the series parts joining them.
+    """
+    members: List[DiffPairNet]
+    bridge_refs: List[str] = field(default_factory=list)
+
+    @property
+    def p_net_ids(self) -> List[int]:
+        return [m.p_net_id for m in self.members]
+
+    @property
+    def n_net_ids(self) -> List[int]:
+        return [m.n_net_id for m in self.members]
+
+    @property
+    def base_names(self) -> List[str]:
+        return [m.base_name for m in self.members]
+
+
+def find_ac_coupled_xnets(
+    pcb_data: PCBData,
+    diff_pairs: Dict[str, DiffPairNet],
+) -> Tuple[List[AcCoupledXNet], List[str]]:
+    """Detect AC-coupling XNets among already-detected ``diff_pairs``.
+
+    A junction between two pairs A and B is recognized only when BOTH polarities
+    are bridged by a distinct 2-pad, non-DNP part (``A.P<->B.P`` AND
+    ``A.N<->B.N``) -- the conservative, symmetric case that unambiguously
+    identifies one electrical signal split by series DC-blocking caps. A 2-pad
+    part with a pad on GND/power (a decoupling cap) is ignored automatically
+    because that pad's net is not a differential-pair member, and a no-pop (DNP)
+    part is ignored because it is an open circuit. Single-sided / asymmetric
+    junctions are skipped and reported as warnings, so the caller keeps today's
+    per-side behavior for them.
+
+    ``diff_pairs`` is the dict returned by ``find_differential_pairs``; its keys
+    are used as stable pair identifiers.
+
+    Returns ``(xnets, warnings)``. Pure read of ``pcb_data``; mutates nothing.
+    """
+    # Map every complete diff-pair member net -> (pair_key, 'P'|'N').
+    net_to_member: Dict[int, Tuple[str, str]] = {}
+    for pair_key, pair in diff_pairs.items():
+        if not pair.is_complete:
+            continue
+        net_to_member[pair.p_net_id] = (pair_key, 'P')
+        net_to_member[pair.n_net_id] = (pair_key, 'N')
+
+    # Per unordered pair-of-pairs, record which polarities are bridged by a 2-pad
+    # part: junctions[(keyX, keyY)] -> {'P': ref, 'N': ref}.
+    junctions: Dict[Tuple[str, str], Dict[str, str]] = {}
+    for ref, fp in pcb_data.footprints.items():
+        if fp.dnp or len(fp.pads) != 2:
+            continue
+        n1, n2 = fp.pads[0].net_id, fp.pads[1].net_id
+        if not n1 or not n2 or n1 == n2:
+            continue
+        m1, m2 = net_to_member.get(n1), net_to_member.get(n2)
+        if m1 is None or m2 is None:
+            continue  # a side is not a diff-pair net (e.g. GND/VCC decoupling cap)
+        (key1, pol1), (key2, pol2) = m1, m2
+        if key1 == key2 or pol1 != pol2:
+            continue  # within one pair, or a cross-polarity (swapped) bridge
+        jkey = (key1, key2) if key1 < key2 else (key2, key1)
+        junctions.setdefault(jkey, {})[pol1] = ref
+
+    # A coupling edge requires BOTH polarities bridged (symmetric); warn otherwise.
+    edges: List[Tuple[str, str, List[str]]] = []
+    warnings: List[str] = []
+    for (kx, ky), pols in junctions.items():
+        if 'P' in pols and 'N' in pols:
+            edges.append((kx, ky, [pols['P'], pols['N']]))
+        else:
+            only = next(iter(pols))
+            warnings.append(
+                f"AC-couple: pairs '{kx}' and '{ky}' are bridged on the {only} side "
+                f"only (via {pols[only]}); not treating as one signal "
+                f"(asymmetric coupling -- per-side matching kept)."
+            )
+
+    if not edges:
+        return [], warnings
+
+    # Union-find over pair keys connected by coupling edges -> XNet groups
+    # (handles chains of 2+ caps, e.g. A--B--C, as one group).
+    parent: Dict[str, str] = {}
+
+    def find(x: str) -> str:
+        parent.setdefault(x, x)
+        root = x
+        while parent[root] != root:
+            root = parent[root]
+        while parent[x] != root:
+            parent[x], x = root, parent[x]
+        return root
+
+    for kx, ky, _refs in edges:
+        parent[find(kx)] = find(ky)
+
+    group_keys: Dict[str, List[str]] = {}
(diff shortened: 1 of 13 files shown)


drandyhaas (commented):
Reviewed and merging — clean 3-way merge with current `main`, all tests pass here (`test_ac_coupled_match`: detection ×7 + measurement + integration + control; `test_diff_pair_route` 5/5; `test_diff_pair_detection` 16+6), full CLI↔GUI parity, and the feature is properly gated/inert when off. The `Dict` import fix you bundled is already on main (#247) so it no-ops after merge — no harm.

Two minor follow-up notes (non-blocking, left for your discretion):

1. **No "don't make it worse" guard in `apply_ac_coupled_length_matching`.** `apply_intra_pair_length_matching` reverts when `new_delta >= delta`; the AC-couple pass always applies meanders and reports the resulting skew. Since meanders only *add* length, minimum-feasible bumps could overshoot `want` and push total-S past total-L, worsening end-to-end skew vs. doing nothing. In practice it's bounded (small deltas → tiny/zero bumps, as the synthetic board's "could not fit" shows) and the residual is reported honestly — but mirroring intra-pair's revert would make the contract consistent.

2. **Interaction with `--length-match` groups is unconsidered.** If a user puts XNet member nets in a length-match group *and* enables `--ac-couple-match`, group matching runs first and AC-couple meanders on top. It measures current state so it won't double-count, but the two can fight. Worth a one-line caveat in the docs or a follow-up.

Both are edge cases already gestured at under Known limitations. Thanks for the thorough writeup and references.

PR #247, opened 2026-06-30
Fix NameError: import Dict in pcb_modification.py

trapperl:
## Bug
Importing `pcb_modification` raises `NameError`, which takes down `route.py` (it imports `pcb_modification` at module load) and every other importer:

```
$ python3 route.py kicad_files/interf_u_unrouted_placed.kicad_pcb out.kicad_pcb
  File ".../pcb_modification.py", line 1263, in <module>
    def cleanup_plane_taps_grazing(pcb_data: PCBData, all_new_segments: List[Dict], ...
NameError: name 'Dict' is not defined. Did you mean: 'dict'?
```

`cleanup_plane_taps_grazing()` is annotated `all_new_segments: List[Dict]`, but the module imports only `List, Optional, Tuple` from `typing` (line 9). With no `from __future__ import annotations`, the signature is evaluated at import time, so the module can't be imported at all.

## Fix
Add `Dict` to the `typing` import — one line.

## Verification
- `python3 -c "import pcb_modification"` now succeeds (was raising `NameError`).
- `python3 route.py kicad_files/interf_u_unrouted_placed.kicad_pcb out.kicad_pcb` routes end-to-end (787 track segments + 82 vias written).

### DIFF
--- a/pcb_modification.py
+++ b/pcb_modification.py
@@ -6,7 +6,7 @@ self-intersecting or redundant segments.
 """
 
 import math
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 from kicad_parser import PCBData, Segment, Via
 from routing_utils import pos_key, POSITION_DECIMALS, into_pad_frame_point


drandyhaas:
Confirmed valid — merging. One clarification on the repro for the record:

The bug is real: `pcb_modification.py:1263` annotates `all_new_segments: List[Dict]` but `Dict` is never imported (line 9 imports only `List, Optional, Tuple`), and there's no `from __future__ import annotations`. The one-line fix is correct.

However, the PR body's repro is environment-dependent. On **Python ≤ 3.13** (eager annotation evaluation), `import pcb_modification` crashes at `def` time exactly as described — breaking `route.py` and every importer. But on **Python 3.14+** the import *succeeds*, because PEP 649 defers annotation evaluation; the `NameError` only surfaces when something forces it (e.g. reading `cleanup_plane_taps_grazing.__annotations__`). So `python3 -c 'import pcb_modification'` will not reproduce on 3.14.

This still matters: the project targets KiCad's bundled Python (3.9–3.12), where the import genuinely fails — so the fix is needed. Accurate framing is 'fails on Python ≤ 3.13, latent on 3.14.' Wording nit only; merging.

drandyhaas:
Follow-up: swept the whole repo for this class of bug and hardened against it.

**Scan result:** after this fix, there are **zero** remaining *bare* (def-time-evaluated) annotation names that are undefined/unimported. The `List[Dict]` here was the only one. A broad scan surfaced ~54 other "undefined name in annotation" hits, but all are harmless string forward-refs (`'GridObstacleMap'`, `'Pad'`, etc.) that are never evaluated at `def` time.

**Hardening (commit 9b88098 on main):** added `from __future__ import annotations` to all 93 annotated modules that lacked it. With deferred annotations, every annotation is stored as a string and never evaluated at runtime (the codebase never calls `get_type_hints`/introspects `__annotations__`), so a missing typing import can no longer break `import` on **any** Python — closing this bug class structurally rather than one file at a time. It also makes the latent-on-3.14 / breaks-on-≤3.13 split moot.

Verified: `compileall` clean across the repo, all key entrypoints import, and `route.py` routes a board end-to-end. `cleanup_plane_taps_grazing.__annotations__['all_new_segments']` is now the string `'List[Dict]'` instead of raising.

PR #251, opened 2026-06-30
Fix TypeError in place_route_loop when a multipoint net fails to route

trapperl:
## Bug
`place_route_loop.py` crashes right after the first route on any board that has an unrouted multipoint net:

```
Round 0: routing initial placement...
  failures=26 iterations=2,453,233 vias=514
Traceback (most recent call last):
  File ".../place_route_loop.py", line 80, in nets_to_refs
    nid = name_to_id.get(name)
TypeError: unhashable type: 'dict'
```

`run_route()` builds `failed_nets` from `failed_single` (net-name **strings**) **plus** `failed_multipoint` — but those entries are **dicts** `{net_name, failed_pads}`. `nets_to_refs()` then calls `name_to_id.get(name)` (and `main()`'s `net_weights` comprehension does `name_to_id[n]`), using a dict as a dict key → `TypeError: unhashable type: 'dict'`. The repair loop dies before it can move anything.

## Fix
Extract `net_name` from the multipoint dicts so `failed_nets` is uniformly net-name strings.

## Verification
- **Before:** crashes entering round 1 with the traceback above (207-part board, 26 failures after round 0).
- **After:** runs past round 0 into the placement-repair quench — no traceback; reaches the part-move + re-route step.

### DIFF
--- a/place_route_loop.py
+++ b/place_route_loop.py
@@ -55,7 +55,11 @@ def run_route(pcb_file: str, routed_file: str, route_args: str, log_file: str):
     summary = json.loads(m.group(1))
 
     failed_nets = list(summary.get('failed_single', []))
-    failed_nets += list(summary.get('failed_multipoint', []))
+    # failed_multipoint entries are dicts {net_name, failed_pads}; keep just the
+    # name so failed_nets is uniformly net-name strings (downstream uses them as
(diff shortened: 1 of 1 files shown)


edgehero:
try pulling from main instead of the release ( @drandyhaas you probably release today again right? with this bugfix )

PR #262, opened 2026-07-01
fix: add missing typing imports and fix Path | None syntax for Python…

andredubovski:
Fix Python 3.9 typing compatibility (KiCad's bundled Python)

KiCad ships Python 3.9 on macOS and Windows. Three issues prevented install_plugin.py and the plugin from loading on Python 3.9:

- Path | None union syntax (PEP 604) in install_plugin.py and package_pcm.py — only valid Python 3.10+, replaced with Optional[Path]
- Missing from typing import statements in 8 files (Dict, List, Set, Mapping, Type, cast, Optional) — these names have never been builtins and require explicit imports

No logic changes; annotation-only fixes.

**Test:
Rebuilt on my Mac and was able to use KRT**

Files changed:
- package_pcm.py: Path | None → Optional[Path] (×2) + add Optional import
- kicad_parser.py: sort typing imports (Counter was already in collections)
- kicad_routing_plugin/board_swaps.py: add Type
- kicad_routing_plugin/claude_gui.py: add Set
- kicad_routing_plugin/settings_persistence.py: add Set (after docstring)
- kicad_routing_plugin/swig_gui.py: add List
- obstacle_map.py: add Mapping
- pcb_modification.py: add Dict
- plane_obstacle_builder.py: add cast

### DIFF
--- a/kicad_parser.py
+++ b/kicad_parser.py
@@ -6,7 +6,7 @@ import re
 import math
 import json
 from dataclasses import dataclass, field
-from typing import Dict, List, Tuple, Optional
+from typing import Dict, List, Optional, Tuple
 from pathlib import Path
 
 
--- a/kicad_routing_plugin/board_swaps.py
+++ b/kicad_routing_plugin/board_swaps.py
@@ -15,6 +15,7 @@ This module deliberately avoids wx so it can be used headless.
 """
 
 from routing_utils import pos_key
+from typing import Type
 
 
 def _to_mm_key(pcbnew, point):
--- a/kicad_routing_plugin/claude_gui.py
+++ b/kicad_routing_plugin/claude_gui.py
@@ -19,6 +19,7 @@ import threading
 import subprocess
 
 import wx
+from typing import Set
 
 PLUGIN_DIR = os.path.dirname(os.path.abspath(__file__))
 ROOT_DIR = os.path.dirname(PLUGIN_DIR)
--- a/kicad_routing_plugin/settings_persistence.py
+++ b/kicad_routing_plugin/settings_persistence.py
@@ -3,6 +3,7 @@ KiCad Routing Tools - Dialog Settings Persistence
 
 Handles saving and restoring dialog settings between invocations.
 """
+from typing import Set
 
 
 def get_dialog_settings(dialog):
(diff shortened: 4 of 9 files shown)


drandyhaas:
Thanks for this, and for caring about the KiCad-3.9 case — it's a real constraint (KiCad does bundle Python 3.9 on macOS/Windows), so it's worth getting right.

I dug into this against a real Python 3.9.21 interpreter, and I don't think these changes are actually fixing the load failure you hit. A few things:

**The plugin already loads clean on 3.9.** I tested three ways on `main` *without* this PR:
- AST scan of the whole tree: no PEP 604 (`X | Y`) unions in live annotations, no `match` statements, nothing 3.10+.
- `py_compile` of every tracked `.py` under 3.9.21: all clean.
- Importing the full load chain under 3.9.21 (engine + all `*_gui` modules + `action_plugin`): every module imports fine.

**The added `typing` imports don't correspond to real usages.** They match method calls and docstring text rather than type annotations, e.g.:
- `board_swaps.py` `Type` → `item.Type()` (a pcbnew method)
- `claude_gui.py` / `settings_persistence.py` `Set` → `widget.Set(...)` (a wx method)
- `swig_gui.py` `List`, `obstacle_map.py` `Mapping`, `plane_obstacle_builder.py` `cast` → only appear in docstrings/comments ("List of…", "ray cast")

None of these are used as typing names, so the imports are harmless but don't change loading. (If one *were* genuinely missing from a live annotation, it'd `NameError` on every Python version, not just 3.9.)

**The two that touch real patterns are also moot:**
- `pcb_modification.py` `Dict` — already imported on `main`, and the file has `from __future__ import annotations`, so the annotation is never evaluated anyway.
- `package_pcm.py` `Path | None` → `Optional[Path]` — that file also has `from __future__ import annotations`, so the union never executes at runtime; and it's a maintainer packaging script, not loaded by KiCad.

So I suspect the thing that actually broke for you was something else — a stale/missing Rust binary for your platform, a leftover `.pyc`, or an install-path issue that the rebuild cleared. **Could you paste the verbatim traceback KiCad's plugin console showed when it failed to load?** That'll point us straight at the real cause, and I'm happy to fix that properly.

andredubovski:
Thanks for taking a look. In attempting to reproduce the issue, I

1. installed a clean KiCAD 10.0.4 on a fresh macOS user account
2. cloned the repo
3. built and installed KRT

I ran into an unrelated installation issue - KiCAD hadn't fully initialized - but that is moot here.

I was not able to reproduce my bug, so I'll close this PR.

Issue #337, opened 2026-07-07
check_drc false negatives on urchin: track overlapping a foreign pad (32 kicad-cli shorting_items) not flagged


drandyhaas:
## The dangerous direction: kicad-cli finds real copper contact check_drc misses

urchin (0707 wave final, `check_drc` = 0 at routed clearance 0.127): kicad-cli reports **32 shorting_items + 10 clearance + 2 hole_clearance**, present since the 0705 baseline (kdrc 39→49).

Verified geometrically at the first site (108.9, 72.1):
- `R_ROW2` F.Cu segment `(106.7,72.1)->(108.9,72.1)` ends ~20µm INSIDE pad D31.2's copper (`Net-(D31-Pad2)`, 0.95×2.29 pad at (109.417,72.30)) — real different-net overlap, `check_drc` silent.
- Additionally suspicious: an F.Cu segment ends and a B.Cu segment begins at exactly (108.9,72.1) — check whether a via actually exists there (my scan didn't query vias); if not, that's a missing-via layer change (writer/router bug) on top of the overlap.
- D31.2 exists as TWO stacked SMD pads (one F.Cu, one B.Cu entry, same number, drill=0) — reversible-footprint style; a pad-shape/duplication modeling issue in check_drc's pad-segment check is one suspect (bbox/rotation/dual-entry handling), the #264 net-name matching family is another.

Repro: `python3 tests/stress/kicad_drc_compare.py <wave>/set1/urchin/<final> ` at clearance 0.127; first KICAD-ONLY shorting item at (108.9,72.1).

This class needs check_drc parity BEFORE trusting near-zero scoreboards: 42 real-looking items on one board that we grade as clean.

drandyhaas:
**Scope broadened by the full-corpus kicad-only taxonomy** (all 0707-wave boards with kicad_only > 0, each finding sub-classified):

- copper_edge_clearance 347 → #338; footprint/pad overrides 551 → #326; board-setup hole 5 (#327-adjacent, tiny).
- `<no net>` overlap artifacts (~24: upduino 8, watchy 4, ottercast/AE1 2, neo6502 2, …) — stacked no-net footprint pads in the INPUT boards, benign, same family as the known AE1 case.
- **Real check_drc false negatives — NOT just urchin (~14+ items on 4+ boards):**
  - urchin: 32 shorting + 10 clearance (original report; track-inside-foreign-pad verified).
[...]

Issue #319, opened 2026-07-05
Multipoint tap: restore the net's own ripped copper when a tap ultimately fails (soft-joint source)


drandyhaas:
## Problem
`route_multipoint`'s phase-3 tap routing rips the net's OWN copper to retry a
blocked tap (`try_phase3_ripup` / `_reroute_phase3_ripped_nets` in
phase3_routing.py). When the retry ultimately FAILS, it restores the ripped
BLOCKER nets (phase3_routing.py:692-704, `restore_net`) but does NOT roll back the
net's own committed main-route + partial-tap copper -- and in particular it does
not restore a clean connecting segment it deleted while re-tapping.

Result: a segment that used to join two pieces of the net is gone, and the pieces
are left held together only by a sliver of cap overlap -- a SOFT JOINT (check_drc
`segment-endpoint-gap`).

## Concrete case (butterstick DQ5)
In the fanout/diff output, DQ5's escape is a clean chain with a connecting segment
`(143.125,106.55)->(143.2,106.6) w=0.1` (coincident endpoints). The signal-step
multipoint tap deleted that segment while re-tapping, the tap failed 2 edges, and
it never restored the segment -- leaving `(143.125,106.55)` and `(143.2,106.6)`
0.09mm apart with a 0.0045mm cap overlap (a soft joint). Traced across steps
(fanout/diff = clean, signal = gapped).

## Desired fix
When a multipoint tap attempt ultimately fails (falls back / abandons), restore
the net's own copper to its pre-tap state (or at least never delete a
connecting segment without restoring an equivalent), so the net is left as
connected as it was before the tap -- no soft joint created. Mirror the
`restore_net` obstacle-cache dance for ref-count consistency (#208/#309).

`close_soft_joints` currently papers over the resulting gap with a tiny bridge,
but not creating the gap in the first place is cleaner.

## Related
- close_soft_joints + check_drc `segment-endpoint-gap` (0aac72e).
- Diff-pair soft-joint source (#318); component-based multipoint (#317).

drandyhaas:
## Update — root cause is NOT rip-restore; it's the cleanup passes' connectivity definition (ties into #320)

Investigated the butterstick DQ5 case end-to-end (reproduced at the exact stress commit `842a024`, before `close_soft_joints` existed). The original framing here — "restore the net's own ripped copper when a tap fails" — turned out to be a **misdiagnosis**. Findings:

### What actually creates the DQ5 gap
[...]

Issue #188, opened 2026-06-23
Custom comb-shaped exposed pads (MOSFET source/drain) modeled as solid bounding box, boxing in adjacent pads (bitaxe Q1/Q2)


drandyhaas:
**Found by:** stress test set 1, `bitaxe_ultra` (run dir `runs_set1/bitaxe_ultra`).

## Symptom

Every routing/connectivity failure on this board is in the Q1/Q2 MOSFET power-converter section — 4 signal/power nets + 5 GND plane pads left unconnected, all reporting `no rippable blockers found`:

- `Net-(Q1-G)` — **fully unrouted** (gate pad, 2 pads)
- `/Power/SW` — 3 Q1 drain pads disconnected (91.09–93.65, 77.31)
- `Net-(Q2-G)` — 1 Q2 pad (99.61, 82.86)
- `/Power/VIN` — 1 U8 pad (85.50, 82.09)
- `GND` plane — 5 pads (U8.11, Q2.1-3, C34.2)

The board was routed at fine grid (`--grid-step 0.05`) and `--clearance 0.1` — i.e. this is **not** the coarse-grid fine-pitch case of #95 (which is closed). The small pads here are at ~1.3 mm pitch, not fine pitch.

## Root cause: custom comb-shaped pads are modeled as a solid bounding box

Q1 and Q2 are `CSD17311Q5` MOSFETs whose large exposed source/drain pad is a **custom comb/finger polygon**, e.g. Q2 pad 5 (`/Power/SW`) in the board:

```
(pad "5" smd custom
  (size 0.01 0.01)
  (options (clearance outline) (anchor rect))
  (primitives (gr_poly (pts
    (xy 0 0) (xy 0 4.56) (xy -5 4.56) (xy -5 3.898) (xy -4.32 3.898)
    (xy -4.32 3.2647) (xy -5 3.2647) ... )   ; notched fingers
)))
```

The actual copper has **notches/channels between the fingers** (the `-4.32` vs `-5` jogs), and `(clearance outline)` means clearance follows that polygon — not the bounding box.

But our pipeline throws the polygon away:

- `kicad_parser._custom_pad_extent()` reads the primitives only to compute a **bounding-box** `size_x × size_y` (Q1.5 → `9.13 × 10.01`, Q2.5 → comb's bbox).
[...]

PR #388, opened 2026-07-13
Fix custom-pad obstacles, --bus Windows crash, duplicate stub connectors, and input-stub deletion in the dead-end sweep

casbrbr-beep:
Four fixes found while running route.py as the deterministic bulk router inside an automated PCB pipeline (KiCad 10, Windows). Each was reproduced before fixing and carries a test; the kit-dev-coldfire-xilinx_5213 reference route is byte-identical before/after the last three (the first changes obstacle modeling by design).

## 1. Custom-pad obstacle modeling + via-via clearance rounding (6079816)
Pads built from primitives (`custom` shape) contributed no copper to the obstacle map, so the router could route straight through them; via-to-via clearance also rounded down at grid quantization. Custom pads now rasterize their primitives into the obstacle grid, and via-via clearance rounds up.

## 2. `--bus` crash on Windows consoles (510317a)
`single_ended_loop.py:192` printed U+2192 arrows in the bus-detection output; under the stock Windows console encoding (cp1252) that raises `UnicodeEncodeError` and kills the run. Repro: any `--bus` invocation with `PYTHONIOENCODING=cp1252`. Fix: ASCII `->` (the arrow the rest of the progress output already uses). Test: an AST scan pinning all string constants in the two bus modules to ASCII.

## 3. Duplicate coincident endpoint-connector segments (654681e)
Routing from off-grid stub free ends emitted two identical ~0.02 mm connector segments at one stub tip (both `route_multipoint_main` and `route_multipoint_taps` re-emitted the same original-to-grid connector through `_path_to_segments_vias`). On a 6-net synthetic fanout-stub board: 5 duplicate pairs. Fix: `route_multipoint_taps` filters emitted segments against the net's accumulated copper (exact twins only, either orientation, same layer/width/net). Test covers exact/reversed/float-noise twins plus layer/width/net non-twins.

## 4. Dead-end sweep deleting the input file's own copper (16d8658)
The final dead-end sweep pruned authored INPUT stubs it considered unused, silently deleting pre-existing board copper (our pipeline authors escape stubs as launch points, so this mattered). Fix: input-file segments pass to `prune_dead_end_segments` as never-removed anchors; only the run's own copper is prunable. The old behavior stays available as `prune_original=True`, which `clean_plane_copper` keeps using (it sweeps copper the plane tools themselves just wrote). Deliberate behavior change: pre-existing dead-end copper in the input now survives a run; `check_orphan_stubs.py` still reports it.


drandyhaas:
Thanks for these, Cas — all four were real, reproduced, and tested, which made triage easy. Since this PR was opened, `main` moved on and three of the four have been overtaken, so I've extracted the one fix that's still novel and am closing this PR.

**Landed: fix 3 (duplicate coincident endpoint connectors).** This is the one still-live-on-`main` defect. I reproduced it on a real repo board — routing `kicad_files/interf_u_unrouted.kicad_pcb`, `route_multipoint_taps` emits 17 exact twins and **3 distinct coincident-duplicate keys (VCC/GND, ~27–33 µm connectors) survive into the written output file**; nothing downstream dedupes them, exactly as you described. Your `_drop_segments_already_present` filter drives it to 0 with no other change (962→958 segments). I've pulled it (plus `tests/test_duplicate_endpoint_connector.py`) into a dedicated commit on `main` crediting you as co-author.

**Superseded / already handled (not taken):**
- **Fix 1 — custom-pad obstacle.** `main` now models custom pads via `pad.polygons` (issues **#188 + #337**): it rasterizes the *real comb/finger copper outline* — keeping the finger channels routable — in **both** the obstacle map **and** DRC, and already models "anchor **UNION** primitives" (#337). The axis-aligned bbox here is coarser (over-blocks the channels) and DRC-blind, so taking it would regress relative to `main`.
- **Fix 2 (via-via rounding).** `main` refactored via-via expansion to a float (`via_via_mm * inv_step`), which is more precise than rounding up to a whole grid cell.
- **Fix 2 (bus ASCII arrows).** Already fixed independently on `main` — `single_ended_loop.py` prints `->` and the bus modules are ASCII-clean.
- **Fix 4 — dead-end sweep deleting input copper.** The concern is legitimate, but `main`'s `sweep_dead_ends` has since grown zone-credit and soft-joint-bridge protection (#319) and copper-graphic force-keep (#337), and we've just landed a related identity-based rip fix (#389, "never remove input copper by signature-twin"). The "don't delete authored input stubs" case is being folded into that line of work directly on current `main` rather than via the pre-refactor patch here.

Closing as superseded — with fix 3 carried forward and credited. Really appreciate the thorough repros and tests on all four; that's what let us confirm exactly which one still mattered. 🙏

Issue #84, opened 2026-06-12
route.py multipoint: orphan dead-end stubs and same-net self-crossings on successfully routed nets


drandyhaas:
Found by the real-world-board stress test (`tests/stress/`, `/stress-test-router` skill). Corpus boards are downloaded/prepared per `tests/stress/README.md`; board names below refer to that corpus.

**Affected boards (orphan stubs):** megadesk (3), splitflap_driver (6), piantor (8), urchin (6), smartknob_base, bitaxe_ultra (48 incl. failed-net stubs) — `check_orphan_stubs.py` flags dead-end tap/branch segments on nets that routed 100% and pass DRC+connectivity.

**Affected boards (same-net self-crossings):** urchin (+8 DRC: ROW0 x3, R_ROW2 x2, COL2/R_ROW x3), smartknob_base (10, e.g. /TMC_VL nets), haxophone (1) — `SEGMENT-CROSSING-SAME-NET` violations introduced by the router on multipoint nets; baselines were 0.

Both patterns point at multipoint tap/topology merging: taps that approach an existing trace, land, and leave the superseded branch in place; and later legs crossing earlier legs of the same net. Electrically harmless but they are DRC noise, copper clutter, and (for the stubs) antenna risk. The guide-corridor docs already note self-crossing as a known quality gap for waypoint concatenation — this is the same class of issue in ordinary multipoint routing.

**Expected:** post-route cleanup pass per multipoint net: trim dead-end branches (no pad/via at endpoint) and merge/eliminate self-crossing segments.

drandyhaas:
Run-2 orphan-stub counts across the corpus (post-fix, all boards else improved): megadesk 3, piantor 4, splitflap 6, urchin 7, haxophone 8, upsy_desky 21, smartknob 32, bitaxe 33, castor 34, lumenpnp 68, neo6502 120. neo6502 detail: predominantly unused qfn_fanout stubs left when nets routed away from the stub or failed — fanout stubs of failed/abandoned nets (11 on +3.3V alone) are never cleaned up. A post-route dead-end sweep would clear these on every board.

drandyhaas:
Additional evidence from the second full stress run (current `main`): orphan dead-end stubs are **pervasive across the corpus**, not board-specific, and persist on nets that `check_connected` reports as fully connected with `check_drc` clean. Per-board orphan-stub counts from this run:

| board | orphan stubs | connectivity | DRC delta |
|---|---|---|---|
[...]

PR #392, opened 2026-07-14
route.py: `--net-clearances` with per-obstacle cross-class pricing; `--keep-input-copper`; chord short jogs against connection-width artifacts

casbrbr-beep:
Hi Andy,

Thank you for the #388 review, happy to contribute. Your newer polygon-pad and float-radius work covers the rest better than my versions did. This branch is rebased onto your current main. Below are three more changes, looking forward to hear your thoughts.

**1. `--net-clearances` CLI + per-obstacle cross-class pricing** (73c84e0)

The engine has carried a `net_clearances` map since the GUI grew
per-netclass support, but the CLI had no way to supply it, and both
obstacle-map builders applied it as one global maximum (`max` over ALL map
values), which over-blocks every routed net at the largest foreign class.
KiCad's pairwise rule is `max(classA, classB)`, so:

- `route.py --net-clearances <json path>` (`{net name -> class clearance
  mm}`), resolved against the parsed board's nets and passed to
  `batch_route`. Live case this fixes for headless runs: a pre-placed
  POWER_HI (0.25) via seen while routing a Default (0.15) group was priced
  at 0.15, and a freshly routed via landed 0.2487 from it: a genuine KiCad
  DRC error the router could not see.
- Both builders now price each pre-placed obstacle at
  `max(routing-side floor, that obstacle net's own class clearance)`, with
  the floor maxed over the ROUTED nets only (a foreign class can no longer
  inflate it). Inert when no map is supplied: byte-identical behaviour.
- `net_rescue` sub-nominal rungs keep the map minus the rescued net's own
  entry: the neck-down still works, while foreign obstacles keep their
  class clearance (dropping the whole map under-blocked exactly the
  cross-class pair the map exists to protect).
- `--net-clearances` is registered in `manifest_to_plan.IGNORE_FLAGS` (a
  board-specific path; the GUI derives the same map live from netclasses).
- Known boundary, deliberately not extended: the per-net obstacle caches
  still price in-run foreign copper at `config.clearance`. Our chain routes
  per net-class group, so the base map covers it; extending the caches is a
  separate discussion if you want it.
- Tests: `tests/test_cross_class_clearance.py` (6 checks: inert baseline,
  cross-class block, no-over-block probe, mm-exact off-grid center).

**2. `--keep-input-copper`** (7c50306)

Chained flows (fanout then route, or an external stage that authors escape
stubs and pre-routed nets) need the input file's own copper to survive a
run verbatim, including dead-end stubs of nets the run FAILED to route,
which the #84 sweep deletes by design. New opt-in flag, default off (#84
semantics unchanged): input copper is never removed, split, or re-bent by
any post-route cleanup pass; it still anchors every degree / T-junction /
grazing / connectivity model, and this-run copper is cleaned normally.
Candidacy is restricted at source in each pass (no strip-list
post-filtering, which would desync the #319 board == write-model
contract). Threaded per CLAUDE.md: both argparse fronts, engine kwargs,
GUI checkbox + settings persistence + `reset_params_to_defaults`, parity
tables. Not enabled on the plane fronts (their file-round-trip runs with
`results=[]`, where everything reads as input).
Tests: `tests/test_keep_input_copper.py` (13 checks incl. a default-run
guard so the flag can never silently become default-on).

**3. Chord short 135° jogs** (70cf62c)

A grid jog (straight + diagonal, each ~one track-width) routinely lands its
outer cap EXACTLY tangent to adjacent same-net copper; whether the copper
union then has a micro-void is float-dust, and KiCad's `connection_width`
checker reports a sub-minimum web around the void. Measured live: one jog
flagged at 0.079 mm although the copper is provably ≥ 0.2 mm everywhere by
polygon erosion; the identical shape 70 mm away grades clean. A
detector would inherit the same float-borderline, so `cleanup_pipeline`
now chords every short jog (arms ≤ 1.5 track-widths, 135°, at least one
this-run arm) across its far endpoints. The chord's sliver triangle cannot
legally contain foreign copper, so it cannot create a clearance violation
by construction; verified live (the flagged board re-grades 0
`connection_width`, no new violations of any type). Additive-only,
deterministic, in the additive tail before `close_soft_joints`. One heads-up:
unlike the two flags above, this pass is always-on, so it changes default
output copper (micro-segments wherever short jogs occur). Happy to gate it
behind a flag instead if you prefer the default untouched.
Tests: `tests/test_chord_short_jogs.py` (12 checks).

**Suite status:** `tests/run_all.py --fast` 108 passed / 30 integration
skips / 1 pre-existing failure (`test_enable_used_layers`: missing fixture
board, fails identically on clean main). Parity gates:
`test_manifest_plan_parity` 35 flag-checks 0 mismatches,
`test_cli_postpass_coverage` 0 failures (no new CLI post-pass; the new
cleanup passes are engine-side, both fronts inherit them).

**Drive-by observation** (not in this PR): `fab_tiers.py`'s U+26A0 warning
prints crash cp1252 Windows consoles (same class as the fixed bus arrows);
your documented `-X utf8` invocation works around it, subprocess pipes need
`PYTHONIOENCODING=utf-8`.


drandyhaas:
## Merging — cross-class clearance is the correct behavior; the remaining gap is characterized and deferred

Landing this. Rationale, from a stress sets 1–5 replay + a deep dive on the hardest multi-class board (cparti_fpga):

### pr392 is *more correct* than main

Graded against the **designer's own netclasses** (SMA 0.35 / USB 0.2 / 50_ohm 0.125), cparti_fpga three ways:

| | fully routed? | clearance DRC @ SMA=0.35 | disconnected |
|---|---|---|---|
| **Human original** | yes (6487 traces) | **0** | 2 |
| **main (pre-merge)** | yes | **168 violations** (`netclass 'SMA' 0.3500; actual …`) — hidden by clamping SMA→0.1 in the output `.kicad_pro` | 9 |
| **pr392** | 35 nets short | **0** | 44 |

The human proves the board **is** routable at spec. main only "completes" by rewriting the netclass rule in the output to hide 168 real clearance violations — exactly the behavior `--net-clearances` + "stop clamping netclasses" removes. Across sets 1–5 the DRC census is unchanged vs main; pr392 simply stops shipping copper that violates the designer's own clearances.

### The remaining completion gap is router *planning*, not correctness — and not a cheap tune

The nets pr392 leaves short all cluster at U1 (FPGA) / U3 (SRAM) pin-fields sitting *under* the SMA band. The human threads them with **1–2 vias each** — escape the pad, drop once to a clean B.Cu, run the whole span underneath, via back up — keeping SMA on F.Cu only so B.Cu stays a clear "cross-under highway." pr392's greedy per-net A* burns **5–11 vias/net**, thrashes all 4 layers, and fragments the next net's escape.

I tested the obvious knobs on the signal step (via_cost↑, near-forbid the plane layers). **All made it worse:**

| config | vias | disconnected @ signal step |
|---|---|---|
| baseline (pr392) | 833 | **19** |
| via_cost 150 | 594 | 23 |
| inner-layers 100× | 1129 | 23 |
| both | 838 | 30 |

Penalizing vias cuts the count (833→594) but *starves* the strategic cross-under drop, so more nets fail; forbidding the plane layers just removes routing resource. The human's low via count is an **emergent property of coordinated escape planning + directional layer roles**, not something forceable with cost weights. That's an algorithmic (search/ordering) improvement — filed as deferred router work, not a blocker for this PR.

### Landing

Merged to main as a `--no-ff` merge; the chord-jog experiment is reverted and deferred to its own issue. GUI parity is wired through (net-clearances threaded to the plugin call sites + parity tests). New tests: `test_cross_class_clearance.py`, `test_cross_class_inrun.py`, `test_keep_input_copper.py` — all green.

Issue #441, opened 2026-07-18
Residual KiCad-confirmed DRC across the corpus (post-#439): ~20 real defects


drandyhaas:
After #439 (corpus A/B: KiCad DRC −1527, check_drc −2164, both engines), the whole 5-set corpus leaves **~20 KiCad-confirmed real defects** (baseline-subtracted vs each board's unrouted input, graded at the routed/fab floor via `kicad-cli pcb drc --refill-zones`). These are the genuine router-introduced items still worth chasing. (The larger `check_drc` tally is dominated by known grader-only false positives on icepi/ecp5 — tracked separately.)

### The real residuals, by board

| board | set | count | type | detail |
|---|---|---|---|---|
| **zynq_ad9364** | 4 | 1 | `shorting_items` | **copper short GND ↔ NetR2_2** @ (99.80, −76.47) — the one real short |
| picodvi | 4 | 5 | `clearance` | 0.052–0.082 mm vs 0.09 floor, cluster @ (211.5, 107.8) |
| vfo_ctrl | 4 | 5 | `hole_clearance` | **actual 0.000 mm** — via/hole barrels touching, @ ~(134–140, 108–120) |
| core1106_cam | 5 | 4 | `copper_edge_clearance` | sub-0.20 mm edge grazes |
| ulx3s | 2 | 3 | `copper_edge_clearance` | **actual 0.000 mm** — copper on the board outline @ (188.08, 112.22) |
| icepi_zero | 4 | 1 | `clearance` | 0.053 mm vs 0.09 floor @ (153.30, 101.58) |
| ghoul | 5 | 1 | `hole_clearance` | 0.222 mm vs 0.30 mm @ (104.80, 57.10) |

**By type:** clearance 6 · hole 6 · edge 7 · short 1.

### Priority
1. **zynq short** — the only connectivity-breaking defect; a genuine GND/NetR2_2 short. Highest priority.
2. **vfo_ctrl hole overlaps (0.0 mm)** and **ulx3s copper-on-outline (0.0 mm)** — hard geometric overlaps, not marginal grazes; likely fanout/via placement and edge keep-out gaps.
3. **picodvi / icepi clearance** — sub-floor copper-copper grazes (~0.05 mm under a 0.09 floor); candidates for the graze-nudge passes.

### Repro
[...]

Issue #70, opened 2026-06-11
Fine grid steps produce sub-cell clearance DRC violations (obstacle expansion floors where pad/track clearance needs ceil)


drandyhaas:
## Symptom

Routing at `--grid-step 0.05` produces pad-segment DRC violations with sub-cell overlaps (all displayed overlaps ≤ 0.046mm, i.e. less than one 0.05mm cell). Found while benchmarking for #36 on `kicad_files/kit-dev-coldfire-xilinx_5213.kicad_pcb`:

| Configuration | check_drc violations |
|---|---|
| Input board (unrouted) | 0 |
| Routed at 0.1mm | 17 |
| Routed at 0.05mm (before #69) | 207 |
| Routed at 0.05mm (after #69) | 67 |

Repro:

```bash
python3 route.py kicad_files/kit-dev-coldfire-xilinx_5213.kicad_pcb /tmp/kit-005.kicad_pcb \
  --nets "/*" "Net-*" GNDA --track-width 0.2 --clearance 0.2 --via-size 0.5 --via-drill 0.4 \
  --hole-to-hole-clearance 0.3 --layers F.Cu In1.Cu In2.Cu B.Cu --grid-step 0.05 \
  # (+ the test_kit_route.py options)
python3 check_drc.py /tmp/kit-005.kicad_pcb --clearance 0.2 --hole-to-hole-clearance 0.3
```

## Suspected cause

Track/pad clearance expansions are converted to cells with `GridCoord.to_grid_dist` (floor), e.g. `obstacle_cache.py` `expansion_grid_by_layer` and the equivalents in `obstacle_map.py`. Flooring under-blocks by up to one cell, so a route can legally (per the grid) sit up to one cell closer to a pad than the real clearance allows. Via-related clearances already use `to_grid_dist_safe` (ceil) for exactly this reason. At 0.1mm the under-blocking is usually swallowed by geometry slack; at 0.05mm it shows up as the tiny overlaps above.

## Suggested fix (not yet attempted)

Use ceiling (`to_grid_dist_safe`) for pad/track clearance expansion as well. Notes for whoever picks this up:

- This makes obstacles slightly more conservative, so it can reduce routability on tight boards and will change routed outputs (no bit-identical regression possible — verify with DRC + connectivity instead).
- Worth re-running the #36 benchmark matrix (interf_u + kit at 0.1/0.05) to quantify the routability cost.
- Decided in #36 discussion to defer this for now.

[...]

PR #443, opened 2026-07-19
Fix via copper clearance from existing drill holes

floze-the-genius:
## Summary

- reserve via cells by the larger of drill-to-drill and via-copper-to-hole clearance
- apply the shared rule to both signal-routing and plane-routing initial via maps
- add a regression for an h2h-legal cell whose via annulus still violates KiCad hole clearance

This addresses the zero-distance `hole_clearance` residual class reported for `vfo_ctrl` in #441. It intentionally does not close the umbrella issue, which tracks several independent boards and defect classes.

## Why

Initial via maps only expanded existing drills by `existing_drill/2 + new_via_drill/2 + hole_to_hole`. A larger via annulus could therefore touch an existing hole even though the two drill edges met the hole-to-hole floor. Newly placed session vias already used the larger copper/drill reservation, so initial and incremental maps disagreed.

## Validation

- `python3 tests/run_all.py --fast --timeout 300 drill_hole npth_hole gnd_via shared_via plane_via via_reuse bare_pad` (7 passed)
- `python3 tests/gui_parity/test_manifest_plan_parity.py`
- `python3 tests/gui_parity/test_cli_postpass_coverage.py`
- `python3 -m py_compile obstacle_map.py plane_obstacle_builder.py tests/test_drill_hole_keepout.py`
- `git diff --check`

### DIFF
--- a/obstacle_map.py
+++ b/obstacle_map.py
@@ -1160,17 +1160,21 @@ def _add_polygon_edge_obstacles(obstacles: GridObstacleMap, polygons,
 
 def block_via_cells_near_drills(obstacles: GridObstacleMap,
                                  drill_holes, via_drill: float,
-                                 hole_to_hole_clearance: float, grid_step: float):
-    """Block via-placement cells within the hole-to-hole drill minimum of each
-    drill hole.
+                                 hole_to_hole_clearance: float, grid_step: float,
+                                 via_size: float = 0.0,
+                                 copper_to_hole_clearance: float = 0.0):
+    """Block via-placement cells within the required clearance of each drill.
 
     A via placed on the grid sits at its cell's real center; block the cell when
-    that center is within the required center-to-center distance of the REAL
-    drill center, tested in mm -- NOT as a floored/quantized integer-cell disk.
+    that center is within the larger of the drill-to-drill and via-copper-to-hole
+    center distances from the REAL drill center, tested in mm -- NOT as a
+    floored/quantized integer-cell disk.
     Flooring the radius (or centering the disk on the quantized drill cell) lets
     a via land a sub-cell inside the hole-to-hole minimum (issue #70 / #125:
     PAD-DRILL-VIA-DRILL at the default 0.1mm grid). Being exact in mm avoids both
     the under-block (a real fab violation) and the over-block (lost routability).
+    Reserving only the via drill also lets a larger via annulus touch an existing
+    hole while its drill still clears (issue #441, vfo_ctrl hole_clearance).
 
     Shared by the signal router (add_drill_hole_obstacles) and route_planes
     (_add_drill_hole_via_obstacles) so both enforce the keepout identically.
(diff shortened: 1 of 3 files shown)


drandyhaas:
## Review: correct and cleanly built, but I'd gate the merge on an A/B — the cited residual no longer reproduces at HEAD

**Geometry — sound.** `required_dist = max(drill/2 + via_drill/2 + h2h, drill/2 + via_size/2 + copper_to_hole)` is the right model: it reserves for both drill-to-drill *and* the via **annulus** (outer copper) against the existing drill edge. The underlying bug is real — the initial via map reserved only `via_drill/2`, so a larger via ring could touch an existing hole while the two drills still cleared h2h, disagreeing with the incremental map (which already reserves copper). Exact-mm test, consistent with the existing #70/#125 approach. `config.clearance` as the copper-to-hole value is a reasonable (slightly conservative) proxy.

**Quality — high.** Shared helper keeps the signal and plane routers in parity (both call sites updated — matches our CLI/GUI parity rule); the regression test is tightened to 5/5 with good boundary cases (asserts the h2h-legal-but-copper-violating cell is now blocked *and* that a cell beyond both floors stays free, so over-blocking is guarded); backward-compatible defaults; Python-only.

**Applies cleanly despite the base drift.** `obstacle_map.py` moved since this branched (ef3e0f9, "441-hole-clearance", merged to main) — but that change added a *separate* `block_track_cells_near_drills` (the TRACK-to-hole class) and left `block_via_cells_near_drills` byte-identical to this PR's base. So #443 (VIA-to-hole) is logically independent and complementary; no conflict.

**Necessity — this is the concern.** The PR is motivated by `vfo_ctrl`'s "zero-distance `hole_clearance` residual," but I can't reproduce that at HEAD. Grading a current HEAD-routed `vfo_ctrl` (`step5c_retry`):

- `check_drc`: **23 violations, all `segment-board-edge`** (the FG net near the outline) — **0 hole-related**.
- `kicad-cli pcb drc`: **4 violations, all `copper_edge_clearance`** — **0 `hole_clearance`**.

So the via-copper-to-hole class this PR targets isn't present on its own cited board anymore (likely already resolved by the merged #441 work, or this route just doesn't place the offending via). The fix closes a genuine but currently **latent** initial-vs-incremental consistency gap rather than an active corpus defect.

**Risk — activation broadening.** The signal call-site guard goes from `hole_to_hole > 0` to `hole_to_hole > 0 or clearance > 0`, so the via-near-drill keep-out now fires on essentially every board. In practice the effect is usually modest (h2h is rarely 0; the common case is a small keep-out growth where copper-to-hole is the tighter constraint), but it's a corpus-wide routing change that can cost completion in dense drill fields.

**Recommendation.** Because it changes routing, I'd hold merge until it's run through an A/B (it would contaminate one we currently have in flight). Gate on:
1. zero completion regression across the corpus,
2. no newly-manufactured DRC, and
3. **at least one board where it removes a kicad-cli-confirmed `hole_clearance` violation** — the crux, since the cited board no longer demonstrates one. If nothing in the corpus reproduces a real removed violation, this is a latent-only fix and the over-block trade isn't clearly justified.

Happy to queue the #443 A/B as a follow-on and, as part of it, scan the corpus for any board with a real via-copper-to-hole `hole_clearance` to settle gate 3.

PR #432, opened 2026-07-17
Emit frontier blocker attribution as a structured JSON_SUMMARY 'blockers' key (#409)

edgehero:
Closes #409.

## What

**Commit 1, emit (report-only).** The latest non-empty `analyze_frontier_blocking` result per net is recorded on a new `RoutingState.frontier_blocking` accumulator (last-wins, capped at 10 entries to match the printed "Top blockers" list) at the 8 single-ended-path analysis sites: `single_ended_loop` first analysis + N-retry re-analysis, `try_phase3_ripup` + re-analysis, `_retry_victim_main_with_ripup` + re-analysis, and the `reroute_loop` single-net branch + re-analysis. At summary time, nets still failed at END of run get their record serialized as the additive `blockers` key, following the `intentional_edge_band_nets` conventions from #408: omitted when empty, wrapped in a defensive try/except. The identical list is attached to `return_results` for the GUI front (and `_empty_results_data` extended to match, per its #382 E5 exact-keys contract). The diff-pair path is untouched; `pair_reports` in route_diff's summary already carries `blocking_nets`.

The recorder is pure bookkeeping after existing analyses and is never read by routing decisions.

Real sample (glasgow_revC, `--nets "/IO_Banks/*"`, trimmed):

```json
"failed_single": ["/IO_Banks/QA4", "/IO_Banks/QB4"],
"blockers": [
  {"net": "/IO_Banks/QB4", "stage": "single_ended",
   "blocked_by": [
     {"net": "/IO_Banks/DB5", "blocked_count": 71, "unique_cells": 71,
      "track_cells": 71, "via_cells": 0,
      "near_target_cells": 10, "near_source_cells": 31},
     {"net": "/IO_Banks/QB6", "blocked_count": 49, "unique_cells": 49, ...}
   ]},
  {"net": "/IO_Banks/IO_Buffer_B/Z1", "stage": "phase3", "blocked_by": [...]}
]
```

Semantics: `blocked_by` keeps the analysis's rip-priority order, capped at 10 per net with a `more` count when truncated; `stage` is one of `single_ended` / `phase3` / `phase3_victim` / `reroute`. A net whose failures never produced a frontier analysis has no entry (QA4 above), and static copper (pre-existing nets, pads, zones) is not attributable, exactly as in the printed diagnostics. Line-size cost is ~150 bytes per blocker, bounded by the cap.

[...]

Issue #348, opened 2026-07-08
Phase-3 rip-up lateral trades: victim loses a different pad than the ripper gains (1-for-1 swaps ship)


drandyhaas:
Split from #347 (full diagnosis in its comments: the ghoul deep-dive). The phase-3 rip machinery can trade sideways: SCK ripped MATRIX_CS and took their shared U5 corridor (+1/−1); k10 ripped +3V3 and the victim's re-route couldn't re-thread the J3 connector field (+1/−1). The #85 arbitration deliberately counts only TOTAL strandings (its comment records that vetoing partial regressions measured worse corpus-wide), so 1-for-1 swaps ship.

Candidate levers, each needing its own corpus A/B:
1. Count a victim's NEWLY-failed pads (vs its pre-rip `saved_failed_count`, already computed in `_reroute_phase3_ripped_nets`) in `pads_lost`, keeping the strict `lost > gained` veto — or flipping equality to 'abandon' (fixes ghoul, but contradicts the measured #85 rationale; measure it).
2. Victim re-route effort parity: the original +3V3 J3 connection came from a 1.06M-iteration orphan-pad fallback; the victim's failed tap edge died at 15k iterations. Give victim re-routes the same fallback ladder before declaring failure.
3. Blocked-cell attribution: MOSI (a this-run, rippable net) walled the sealed pocket but `analyze_frontier_blocking` attributed only pads/stubs (the pocket boundary), so no N=2 rip was attempted — attribute the DECISIVE wall, not just the nearest copper (same blind spot as core1106's TRIG_GPIO vias).

Repro: ab_347_0708/new/ghoul (MATRIX_CS U5 pad 98.42,67.75; +3V3 J3.74 137.29,71.33) vs base wave. Related: #347, #85, #189.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012HY47S9XjHpmTDMmoL3BM5

Issue #213, opened 2026-06-26
bga/qfn fanout: escape vias violate clearance to neighbouring pads (esp. cap pads); cap-placement remedy incomplete on dense boards (PAD-VIA)


drandyhaas:
**Combined issue** (folds in #222, which was the cap-pad subset of this same mechanism).

## Summary

`bga_fanout` / `qfn_fanout` drop escape vias that violate clearance to **neighbouring
component pads** — most often **decoupling-cap pads** adjacent to the BGA/SDRAM. This is
the **2nd-largest net source of remaining DRC corpus-wide** in the 45-board no-LLM stress
redo (HEAD `05b22a8`): **fanout introduces +252 PAD-VIA after its own cap-placement
cleanup**, concentrated on the dense FPGA/SDRAM boards fanout exists to serve.

The post-fanout cap-placement optimizer (#110, #128) *mostly* fixes this by nudging the
offending cap clear — **but under congestion, when there's no clear spot to move the cap
to, the violating escape via is left in place** and survives to the final board.

## Evidence (DRC at routed clearance 0.1)

| board | final PAD-VIA | example |
|---|--:|---|
| daisho | **75** | `Pad:/front_end/FE_VCCIO_C (C72.1) <-> Via:.../D49`, overlap 0.014 mm |
| orangecrab | **15** | `Pad:P1.35V (C68.1) <-> Via:RAM_BA0`, `Pad:P2.5V (C55.1) <-> Via:RAM_CAS#`, ... (+6 VIA-SEGMENT) |

Representative daisho violations (escape via ~0.5 mm diagonally from a cap pad pair,
grazing both by ~0.014 mm at 0.1 mm clearance):
```
Pad:/front_end/FE_VCCIO_C (C72.1) <-> Via:/fpga_front_end_bank_c/D49   overlap 0.014mm
  Pad (120.00,66.05)  Via (119.50,66.50)
Pad:GND (C72.2)       <-> Via:/fpga_front_end_bank_c/D49   overlap 0.014mm
  Pad (120.00,66.95)  Via (119.50,66.50)
```

## Step attribution (daisho DRC by step)

`bga_fanout U1 → 93`, `bga_fanout U2 → 120`, unchanged through diff/signal/planes,
`route_disconnected_planes → 75` (plane-repair clears 45 of the 120 but cannot clear the
fanout escape vias). So the residual 75 are all fanout-origin escape vias grazing cap pads.

## Suggested directions

[...]

PR #455, opened 2026-07-21
quench: cap swap displacement + prune candidate scans (#430); fix the loop's steering (#458)

### LINKED ISSUE #430: quench: swaps bypass --max-displacement (swapped parts strand outside their own nudge disk); per-candidate scans are O(all parts)
`docs/placement-optimization.md` already flags the symptom in passing: *"same-footprint swaps are not displacement-capped (one resistor traveled 32 mm via swap)"*, and suggests a cap or flag for parts where location is meaningful (line 396). Reading `placement/quench.py` for the #407 follow-ups, the mechanics are a bit worse than a missing nicety: the swap phase breaks the tool's documented contract, has a second-order interaction with the nudge phase, and sits next to a perf ceiling that the block-scale direction will hit. Three items, one file.

## 1. Swaps bypass the `--max-displacement` contract

`place_optimize.py:8-12` states the core safety property:

> Each part stays within --max-displacement of its original position, so the output is "your placement, nudged" - implicit constraints encoded in the original placement survive.

The nudge phase enforces this (`_candidate_positions`, `quench.py:337-355`, hard radius check at :347). The swap phase does not: `quench.py:463-518` evaluates the pair cost and calls `apply_move` directly, with no distance check against either part's seed. Given the experiments' own dose-response finding (3 mm cap is the sweet spot, 10 mm collapses the placement), an uncapped move channel is not a cosmetic gap: `--max-displacement` is the dominant safety knob, and swaps sidestep it entirely.

(Deliberately *not* a bug: the swap phase also skips `candidate_valid`. That part is fine, since a same-footprint pose exchange is occupancy-identical and can't create new overlap. Noting it so nobody "fixes" it.)

## 2. Second-order effect: a swapped part strands outside its own nudge disk

`_candidate_positions` generates candidates around `seed_x/seed_y`, which is correct for enforcing the contract. But after an uncapped swap, the part may now sit *outside* its own candidate disk. On the next pass, its nudge evaluation compares the current position against candidates clustered around the stale seed, so a single "nudge" acceptance can teleport it the full seed-distance back in one move, with halo/crossing deltas evaluated across the whole jump. The nudge phase's small-local-move invariant silently stops holding for any part that has ever been swapped.

## 3. Per-candidate scans are O(all parts), and the fix already exists in-repo

`candidate_valid` (`quench.py:269-282`) and `part_geometry_cost` (:257-267) each loop over **all** parts, and they run for every candidate: roughly pi * (max_disp/step)^2 grid points times up to 4 rotations, per movable part, per pass. `placement/fanout_clearance.py:425-455` solved the identical problem with exact spatially-pruned neighbour lists. Its own comment: *"excluding it is exact, not an approximation. This turns the per-candidate hard_blocked / penalty loops from O(all parts) into O(handful), the dominant cost on dense boards (#213 profiling)."* The same reach argument holds verbatim for the quench: a part moves at most `max_displacement` from its seed, so anything whose seed gap already exceeds that reach can never interact with it.

## Ask

1. Enforce the cap on swaps: skip a swap pair when either part would land farther than `max_displacement` from its seed. If free swaps of generic passives are worth keeping, gate them behind an explicit `--swap-max-displacement` (default = `max_displacement`) so the contract violation is opt-in rather than silent.
2. With capped swaps, item 2 disappears by construction (the disk always contains the part). If an uncapped mode survives, re-seed a part on swap so its candidate disk follows it.
3. Port the pruned neighbour-list pattern into `candidate_valid` / `part_geometry_cost`.

Why now rather than someday: this is #110's "improve it" bullet in its cheapest form, and the block-scale direction from #407 wants denser candidate grids and larger displacement windows, exactly the regime where both the O(all-parts) scans and silent uncapped teleports bite hardest.

Happy to PR any or all three.

### LINKED ISSUE #458: place_route_loop: tally ignores reconciliation recoveries, failed-net weighting is nearly inert, and the #455 swap cap widens with every rejected round
Four gaps in how the loop steers the quench. They compound: the loop overstates failures, barely biases the cost toward the nets it cares about, and widens the one cap that should probably stay fixed. Quench line refs are against 17d4670 (PR #455 head).

1. Stale tallies. The loop reads the first JSON_SUMMARY in the log (place_route_loop.py:52-54). route.py runs a final reconciliation pass by default exactly when failures exist, prints a second JSON_SUMMARY plus recoveries, and says in so many words that the run's full tally is the earlier summary plus those recoveries (route.py:1838-1935). So the loop counts recovered nets as failed, quenches parts anchoring nets that are actually solved, and better() (place_route_loop.py:105-109, 215-223) compares inconsistent tallies across rounds.

2. Near-inert failed-net weighting. net_weights only scales airwire length (quench.py:300-305); crossings are counted unweighted (:316-318). At the loop defaults (length_weight 0.3, crossing_penalty 30), --failed-net-weight 3.0 raises a length coefficient from 0.3 to 0.9 against a crossing coefficient of 30. The mechanism the whole loop is documented around barely changes any decision. Weighting a crossing by the max of the two involved nets' weights would make it real.

3. Swap cap widening. The quench call (place_route_loop.py:187-199) passes neither swap_max_displacement nor allow_rotations/allow_swaps, and no flags exist for them. Since the swap cap defaults to max_displacement (#455), the 1.5x widening on a rejected round widens swap teleports in lockstep: a 3mm base reaches about 15mm after a few rejections, which is the #430 failure mode coming back through the loop. Suggest holding the swap cap at the base value unless explicitly raised, and plumbing the three knobs.

4. Rotation semantics on the same knob surface. --no-rotate only gates nudge rotations (quench.py:512-513); an accepted swap still exchanges rotations (:608-609), so a 0-degree part can end at 90/270 or even 45 via a rotated partner. And rotation candidates require orig_rot % 90 == 0, so a 45-degree-seeded part never tries any rotation at all, while a part that swap-inherits 45 can only be nudged back onto 90-degree multiples, never moved while keeping its current 45.

Drive-by if someone touches the file: route.py is invoked by bare relative path with an unchecked return code (place_route_loop.py:46-49), so running from any other cwd reports "produced no JSON_SUMMARY" instead of the real error.

edgehero:
Fixes drandyhaas/KiCadRoutingTools#430
Fixes drandyhaas/KiCadRoutingTools#458

## The three defects, and their fixes

**1. Swaps bypassed `--max-displacement`.** Same-footprint swaps exchanged part poses with no displacement check at all — only the cost gain mattered — violating the documented "each part stays within `--max-displacement` of its original position" guarantee. Fix: a new `swap_max_displacement` kwarg on `quench()` (default `None` = `max_displacement`; `ValueError` if above it or negative). The swap pair loop now rejects any swap that would leave either part beyond the cap from its **own** seed, checked before the expensive cost evaluations; the verbose pass line reports `swap-capped=N`. Exposed on `place_optimize.py` as `--swap-max-displacement` (can tighten the cap for swaps specifically, never exceed it); `place_route_loop.py` inherits the capped behavior through the default (note: the loop's 1.5x widening on rejected rounds widens the swap cap in lockstep; tracked with the loop's other steering gaps in #458). Two hardening guards ride along: same-name footprints with differing per-ref courtyard bounds no longer swap, and a part inheriting its partner's non-90° rotation now gets exact `_rotate_local_bounds` entries instead of `rect()`'s silent rot-0 fallback.

**2. Parts stranded outside their nudge disk.** A swap used to be able to teleport a part beyond `--max-displacement`, where every subsequent nudge candidate (generated around the immutable seed) was out of reach — the part was stranded, unrecoverable by later passes. Now impossible by construction: the swap cap ≤ `max_displacement` keeps every part inside its seed-centered candidate disk at all times. Seeds stay immutable and no re-seed machinery was added — see design decisions below.

**3. O(all-parts) candidate scans.** `candidate_valid` and `part_geometry_cost` scanned every part per candidate pose (O(parts²) per pass). New `QuenchState.build_neighbor_lists(travel_budget)` ports the pruned-neighbor pattern from `fanout_clearance` (#213 profiling): per-part lists built once from union-of-rotations seed rects, axis-wise overlap test with margin `budget_a + budget_b + max(clearance, halo_a + halo_b)`. Because a part's live position provably stays within the budget of its seed, the pruning is **exact**, not approximate — a dedicated test enforces bit-identical results with and without the lists. Wired up as `build_neighbor_lists(max_displacement + grid_step)`.

## Design decisions

- **Strict cap rather than cap-plus-repair.** The repo's own experiments (docs/placement-optimization.md) found ~3 mm displacement the sweet spot on both boards tested and uncapped movement actively harmful to the placement's macro structure. So the cap is airtight for every move type, and swap-driven large perturbations are deliberately not "repaired" back — they simply never happen.
- **No re-seeding.** Seeds stay immutable; re-seeding after a swap would let displacement accumulate across passes and silently reintroduce the unbounded-drift failure mode. Large-perturbation exploration is deferred to the #407 block-relocation direction, where it belongs.
- **Swaps still skip `candidate_valid`.** Safe because of the occupied-space invariance the new identity guard enforces: two parts may only swap when their courtyard bounds are identical, so the set of occupied rectangles is unchanged by the exchange.

## Verification

- **9 new tests** (`tests/test_quench_swap_cap.py`, `tests/test_quench_neighbor_lists.py`): improving-but-over-cap swap rejected, in-cap swap accepted with exchanged positions, `swap_max_displacement` tightening, locked parts never swap, kwarg validation, no-stranding invariant on a full `interf_u` run, pruned-vs-brute-force bit-exact parity (`candidate_valid` + `part_geometry_cost` + full-run placements), and a white-box guard that the seed rects union over all `bounds_by_rot` entries including non-90° seed rotations. All pass, plus the existing `tests/test_fanout_clearance.py` (12/12).
- **End-to-end** on `kicad_files/interf_u_unrouted.kicad_pcb` at `--max-displacement 3`: `check_drc.py` reports `NO DRC VIOLATIONS FOUND!` before and after; `check_connected.py` reports `FOUND 110 ISSUES / Unrouted nets (110)` on both input and optimized output (counts compared, not file hashes, per CONTRIBUTING).
- **Timing** with `python3 place_optimize.py kicad_files/kit-dev-coldfire-xilinx_5213.kicad_pcb <out> --max-displacement 5 --max-passes 2` (160 footprints): ~110 s wall pre-fix → ~69 s post-fix (68.7 s / 69.6 s over two runs), ~1.6×. The win grows as `--max-displacement` shrinks.

## Possible follow-up (pre-existing, not addressed here)

`QuenchState._net_points` iterates the `net_refs` **set**, so airwire point ordering — and through MST tie-breaking, quench output — varies across processes with `PYTHONHASHSEED`. Same-process runs are deterministic (the parity tests rely on this). Now tracked in #457 (which also covers a second set-iteration site inside the swap evaluation).

---

## Also closes #458: the four ways the loop steers the quench badly

#458 was filed against this PR's head. One of its items is a regression this
PR would otherwise ship: `place_route_loop.py` widens `--max-displacement`
1.5x after every rejected round and never passed `swap_max_displacement`,
which defaults to `max_displacement`. So the cap this PR adds would have
widened in lockstep, turning a 3mm budget into 15.2mm after four rejections,
which is exactly the #430 stranding failure coming back through the other
front end. Fixing it here rather than in a follow-up seemed right, and the
other three items plus the drive-by came along since they are the same
subsystem. Six commits, each self-contained; happy to split any of them out.

**1. `place_route_loop`: hold the swap cap at the base displacement.** The
loop now passes `swap_max_displacement` explicitly and pins it to the base
value for the whole run, so widening the local nudge search can never become
a long-range swap. The three knobs the loop never plumbed are real flags with
the same names `place_optimize.py` uses: `--swap-max-displacement` (validated
at the argparse layer, mirroring `place_optimize.py`, rather than letting the
value reach quench and raise from inside a round), `--no-rotate`,
`--no-swap`, plus `--verbose` so the `swap-capped=N` diagnostic this PR adds
is reachable from the loop at all. A negative `--max-displacement` is
rejected too: passing the swap cap explicitly would otherwise turn it into a
traceback where it used to be silently inert.

**2. `place_route_loop`: count reconciliation recoveries in the round tally.**
`route.py` runs its end-of-run reconciliation pass (#348) exactly when the
first pass left failures, and prints a second `JSON_SUMMARY` scoped to the
retried nets. The loop read the first one, so a net that pass recovered still
counted as failed: the loop then quenched parts anchoring nets that were
already solved, and `better()` compared tallies taken at different points in
the run. `merge_route_summaries` now takes failure state from the **last**
summary and **sums** effort across all of them.

Failure state from the last summary is exact rather than a delta: every net
with a nonzero failure term is in the retry set by construction, and the
sub-run re-derives each retried net's pad counts over all of that net's pads
from the final-board union-find (route.py:1840), so those numbers are
absolute. Summing pad counts would double-count. Effort sums because both
passes are work the placement cost the router; taking iterations from the
last summary would make a badly failing candidate look cheap, since the
reconciliation pass only re-routes a handful of nets. On a single summary,
which is what a clean run produces, the reduction is exactly the old
behaviour.

Two cases the case walk turned up, both pinned by tests. A reconciliation
that raises **after** printing its summary claims recoveries the board write
may never have committed, since route.py prints the summary before writing;
failure state falls back to the first pass, which is what is definitely on
disk. And coverage-gate nets have no routed result, so their pads never reach
`multipoint_pads_total` and they weighed zero even though they ship at broken
copper; each now counts 1. That second one matters most on the last summary,
because those are nets the reconciliation itself broke through its rip
escalation, and without it the loop can read `failures=0` on a board shipping
disconnected copper and stop. **It is four lines and droppable if you would
rather not change that count.**

**3. `place_route_loop`: run route.py by absolute path and check its exit
code** (the drive-by in #458). A bare relative `route.py` only worked when
the caller's cwd happened to be the repo root; from anywhere else the child
exited 2 with "no such file", the code was discarded, and the operator got
"route.py produced no JSON_SUMMARY". Two related holes closed while there:
the log is read as UTF-8 with `errors='replace'` (the bare `open()` decoded
with the locale default, so a cp1252 locale died on the first non-ASCII glyph
and lost the round), and both `RuntimeError`s carry the last 15 log lines. No
`cwd=` override, deliberately: route.py resolves its own assets from
`__file__`, and relative paths inside `--route-args` must keep resolving
against the operator's cwd.

**4. `quench`: price each crossing by the heavier of the two nets' weights.**
`net_weights` scaled airwire length only, and length is the small term. On
kit-dev-coldfire at the loop's own knobs the objective splits 3.5% length,
95.7% crossings, 0.9% halo and edge, so `--failed-net-weight 3.0` raised a
length coefficient from 0.3 to 0.9 against a crossing coefficient of 30:
about 6 out of 66000 for a net of median airwire length, a fifth of one
crossing. The mechanism the loop is documented around was numerically
invisible.

`max(w_a, w_b)` rather than sum, product or mean. It is neutral at all-ones,
so an unweighted board is untouched and `--crossing-penalty` keeps meaning
"the cost of an ordinary crossing"; it prices a conflict by the most
important net it obstructs; and it is the only rule under which the flag's
number is literally the multiplier. `_count_crossings_np` already broadcasts
net ids into an `(n,m)` `same_net` matrix, so this needs no new data, just a
weight matrix on the same broadcast. Both crossing helpers return
`(count, weighted)`: the cost uses the weighted figure, the count stays a raw
int, so `total_cost()['crossings']` and the per-pass banner are unchanged.
About 2% on the crossing kernel, weighted runs only.

**Disclosure:** the same numeric `--failed-net-weight` now has roughly 30x
more leverage, because the weightable share of the objective went from 3.5%
to essentially all of it. Loop defaults are unchanged and no second knob is
added.

**5. `quench`: `--no-rotate` must stop swaps from exchanging rotations.**
`allow_rotations` gated the nudge phase's candidate list and nothing else,
but a swap exchanges full poses. On the two-part fixture in the new test,
`--no-rotate` used to return C1 at 90 and C2 at 0: both parts rotated,
neither by a rotation move. Swaps are now restricted to pairs that already
share a rotation when rotations are off, which also keeps the exchange
rotation-neutral and so **strengthens** the occupied-space invariance this
PR relies on to skip `candidate_valid`. The check sits before the cap check
so `swaps_skipped` keeps counting cap rejections only.

This is a contract defect rather than a geometry one: because a swap
exchanges full poses between parts with identical courtyards, the set of
poses on the board is unchanged and no overlap was ever produced. What was
wrong is that a user who typed `--no-rotate` got a board where the part he
placed at 0 sits at 90. **Honest cost:** the flag now also declines
mixed-angle same-footprint pairs, which on real boards is most of them.
Equal-angle pairs as a share of all same-footprint pairs: kit-dev-coldfire
625/2160, orangecrab_ext_pll 559/2132, tigard 216/628, glasgow_revC
1901/4792, interf_u 7/21. The default path is untouched, since the guard is
dead code whenever rotations are allowed.

**6. `quench`: rotation candidates relative to the part's own angle.**
Candidates were the four axis rotations, offered only when the **seed**
rotation was a multiple of 90. So a part placed at 45 degrees never tried a
rotation at all, and a part that inherited 45 from a swap partner was offered
only 0/90/180/270, i.e. its only way to move was to snap onto an axis.
Candidates are now the 90-degree lattice through the part's current angle,
plus the lattice through its seed angle when a swap moved it off that one,
which is strictly additive: nothing reachable before becomes unreachable.

This is a no-op on any board whose footprint rotations are all multiples of
90, which is 20 of the 22 boards in `kicad_files` and all three used by the
placement tests. Candidates are built in `ROTATIONS` order from a base of
`rot % 90`, so such a part gets `ROTATIONS` itself: same values, same order.
The order is asserted because the caller keeps the first strict minimum, so a
reordered list silently changes which pose wins a tie; the two tempting
rewrites that break it (a set literal, and using `part.rot` instead of
`part.rot % 90` as the base) are called out in the code.

`_Part` now records the whole lattice of a non-orthogonal seed, which is what
keeps `build_neighbor_lists` exact **without touching its code**: it already
unions `bounds_by_rot` over the movable same-footprint group, and with the
full lattice recorded that union is `{seed angle + 90k}` over the group,
precisely the closure of what a swap can hand over and a nudge can then
rotate within. Measured pruning cost on the four non-orthogonal parts in
`kicad_files` (glasgow_revC R9 at 45, rp2350_fpga_eensy_prePlane R9/C28/D1 at
315): zero, because their courtyards are origin-symmetric and the four
lattice boxes are congruent. Worst case for an off-centre courtyard is a
doubling of the box area, scoped to one footprint group.

### Verification (#458 commits)

- **Byte-identity on real boards.** With `PYTHONHASHSEED` pinned (quench
  output varies across processes, #457), a full `quench()` run on
  `interf_u_unrouted` returns byte-identical placements at `91a90b2` and at
  the head of this PR, at three different seeds; same result on
  `orangecrab_ext_pll` (165 parts). So commits 4 to 6 above change nothing on
  an unweighted, all-orthogonal board, which is every board in the test
  fixtures. Passing an explicit all-ones weight for every net also returns
  the same placement as passing none.
- **The existing quench tests pass unmodified**: `test_quench_swap_cap.py`
  6/6 and `test_quench_neighbor_lists.py` 4/4, including the bit-exact
  pruning parity test and the group-rotation closure test.
- **Three new test files**, all in the `--fast` lane (no child process, so
  `run_all.py`'s source sniffing keeps them out of the integration set):
  - `tests/test_458_loop_steering.py`, the first test this file has ever had.
    It replaces the routing step, the quench and the writer with recorders,
    and asserts that four rejected rounds walk `max_displacement`
    3 / 4.5 / 6.75 / 10.125 while `swap_max_displacement` stays 3.0. The
    captured kwargs are bound against the real `quench` signature so a
    renamed keyword fails there instead of silently doing nothing. The tally
    half drives `run_route` against canned two-pass logs: recoveries not
    counted, effort summed, blockers still unioned across both passes, a
    single-summary run unchanged, a net the reconciliation itself broke
    reported and weighed, the aborted-reconciliation fallback, and the
    sub-run-printed-nothing degradation.
  - `tests/test_458_quench_net_weights.py`. The decisive one is
    `test_weighting_flips_the_optimizer_choice`: one board, one set of knobs,
    only `net_weights` differs, and the optimizer's answer changes from "stay
    at the seed" to "move". Before the fix both runs returned `[]`. Also
    pinned: all-ones reproduces the integer count exactly with `==` on the
    floats, the weighted halving is exact, `max` and not sum or product, and
    `stats['crossings']` is still an unweighted int.
  - `tests/test_458_quench_rotations.py`. The discriminating pair for
    `--no-rotate` is one board and one set of knobs differing only in the two
    parts' seed angles: the mixed-angle pair is refused, the equal-angle pair
    still swaps. For the lattice, the neighbour-list test derives its
    threshold from `_rotate_local_bounds` rather than hard-coding it, and
    uses an off-centre courtyard on purpose, since for the origin-symmetric
    courtyards of ordinary passives every lattice box is congruent and only
    an asymmetric one can tell the two closures apart.
- `tests/run_all.py --fast`: 176 passed, 1 failed, 33 skipped. The baseline at `91a90b2` in the
  same environment is 173 passed, 1 failed, 33 skipped: the three new files
  are the delta, and `test_506_507_plan_movie_tools.py` fails identically on
  both sides, so it is pre-existing here and not a delta from these commits.
- **Real-board loop run** on `kicad_files/kit-dev-coldfire-xilinx_5213.kicad_pcb`:
  the loop runs end to end with the new flags and the cap visibly holds.
  `max_disp` widened 3.0 / 4.5 / 6.8 across the rejected rounds while
  `swap cap=3.0mm` never moved, and `--verbose` shows the `swap-capped=N`
  line, a diagnostic the loop previously had no way to display. Failures 43
  to 39 over five rounds.

  I ran the same chain three ways to see what the change actually does. Both
  the first two produce a byte-equivalent round-0 route (identical checker
  counts on both sides), so the only difference is the steering. The third
  is this PR's code with `--failed-net-weight 1.0`, which makes the new
  crossing weighting a proven no-op while every other fix stays active:

  | final board | failures | check_connected | SEG-SEG | VIA-SEG | PAD-SEG | PAD-VIA |
  |---|---|---|---|---|---|---|
  | `91a90b2` | 40 | 44 issues | 73 | 20 | 21 | 14 |
  | this PR, weight 3.0 | 39 | **41 issues** | 103 | 40 | 75 | 17 |
  | this PR, weight 1.0 | 39 | **41 issues** | 84 | 62 | 39 | 16 |

  Those violation counts are graded at `--clearance 0.2`. That is the wrong
  floor to judge these boards by, and it took me a while to see it: on this
  chain the router escalates and reports `min_clearance_used: 0.1` on **all
  three** runs, so a 0.2mm grade is counting copper the router deliberately
  laid finer. **Graded at 0.1mm, all three final boards are
  `NO DRC VIOLATIONS FOUND`.** Checking the severity of every 0.2mm-graded
  violation confirms it: the worst overlap is exactly 0.1000mm, i.e. a
  minimum actual copper-to-copper gap of 0.1000mm, and that worst case is
  **identical on the baseline board, on this PR's board, and on the shared
  round-0 input**. Nothing touches on any of them. Every violation sits in a
  90 micron band between 0.100 and 0.190mm.

  So the counts are a measure of how much sub-0.2mm copper each run ended up
  with, and they are volatile: VIA-SEGMENT goes 20 / 40 / 62 and PAD-SEGMENT
  21 / 75 / 39 across the three runs with no monotone relationship to the
  weighting. Round to round within a single run they swing just as hard (the
  baseline's own last re-route took it from 170 total rows down to 130 while
  moving 5 parts). Of the incidents that appear in this PR's final round, 20
  of 25 are more than 10mm from any of the 4 parts that round moved, and the
  largest cluster sits on an unmoved part 12 to 19mm away. The mechanism is
  which global routing solution the router happens to land on, not the
  placement. With one run per configuration I cannot rule out that the
  steering systematically produces grazier routing, and I am not going to
  claim otherwise from n=1; what I can say is that no board in this
  experiment has copper closer than the 0.1mm the router itself chose, and
  connectivity improved on both of this PR's runs.

  Two findings from that experiment that are worth more than the DRC numbers:

  **The tally is what changes the trajectory, not the crossing weighting.**
  Re-scoring the identical round-1 candidate board under both semantics: the
  old reading gives 42 failures against a 42-failure incumbent and accepts on
  a 0.44% iteration margin; the new reading gives 44 against 43 and rejects.
  That single flip sends the two runs down different paths, and everything
  downstream follows. Isolating the quench objective with identical inputs
  and a pinned `PYTHONHASHSEED`, the crossing weighting moves 1 part by
  0.50mm at a 3mm cap and up to 3 parts at a 6.8mm cap: real, but secondary.
  The weight-1.0 run above reaches the same 39 failures and 41 issues as the
  weight-3.0 run, so on this board the weighting bought no connectivity
  either. Worth knowing why: 43 of 276 scored nets are weighted here, and
  82.9% of all crossings touch at least one of them, so the weighting is
  close to a global scale factor rather than a selective bias. It should
  matter more on a board with a handful of failures, which is the regime the
  loop is documented for.

  **A disclosure about the new tally I only found by measuring.** On this
  board the reconciliation pass recovers nothing; it re-derives the same 43
  still-open nets as `42 single-ended + 1 multipoint with 1 open pad` where
  the first pass called them `33 single-ended + 10 multipoint with 9 open
  pads`. Both readings are describing the same 43 nets, but the new one
  counts each reclassified net as 1 instead of by its pad deficit, so
  `failures` reads 1 to 2 higher and is **less** sensitive to partial
  progress on a multipoint net than it used to be. I think the new number is
  the more honest one (43 nets are open, and the old formula only undercounts
  because it prices multipoint nets per pad and single-ended nets per net),
  but the loss of partial-progress sensitivity is real, and it is exactly the
  granularity problem #432's pad-pair tallies exist to solve. Reworking
  `better()` onto that metric is the part of #458 I have deliberately not
  done here.

- `git merge-tree` against current `main` (`e8c72fe`): still conflict-free.

### Decisions worth stating, so they are not re-proposed

- **No `--no-final-reconcile` flag on route.py.** It would let the loop skip
  the reconciliation during search rounds, but `route.py` is in both
  `PARITY_SURFACES` and the GUI parity `CLI_MAINS`, so a new engine flag
  drags in the whole GUI thread-through checklist plus both gates, and the
  benefit is nil: that pass is a fraction of a second against rounds of
  millions of iterations, and skipping it during search would make the loop
  optimise against a metric the final board is not graded on, which is #458
  by another route.
- **The crossing weighting is severable.** It is one commit (`d880c81`) and
  dropping it leaves the other five intact. Measured above it is the
  secondary effect, not the driver of the trajectory change, and on this
  board it bought no connectivity, so I would understand wanting it split
  out and evidenced on a board in the loop's documented regime first.
- **No second crossing-weight knob** and no change to the loop's default
  weights. One flag, one meaning; changing `--length-weight` or
  `--crossing-penalty` would also invalidate the experiment tables in
  docs/placement-optimization.md.
- **No `--max-passes` on the loop** (quench's default already matches
  `place_optimize.py`) and **no absolute ceiling on the widening**
  (`--rounds` already bounds it; an unbounded nudge search is a different
  failure mode).
- **Position-only swaps rejected** as the fix for `--no-rotate`. Putting a
  wide rectangle where a tall one was breaks the occupied-space invariance
  exactly when it matters, and since swaps skip `candidate_valid` the
  resulting courtyard overlap would ship silently. Making it safe costs a
  `candidate_valid` call per part plus an explicit A-against-B clearance
  test, and the result is a new move type rather than a fix to the flag.

### Notes for review

- **Parity cost.** `place_route_loop.py` is in neither `PARITY_SURFACES` nor
  the GUI parity `CLI_MAINS`, so commits 1 to 3 add none. `placement/` is
  already in this PR's audit scope, and `kicad_routing_plugin/` contains no
  quench call site at all, so that audit stays a formality.
- **Overlap with #432.** That PR also edits `run_route`. Commit 2 leaves
  `place_route_loop.py` lines 56 to 76 byte-identical, so #432's hunk applies
  at an offset with no conflict whichever lands first. One thing to fix at
  that point: under a last-wins merge #432's `pad_pairs_connected` and
  `pad_pairs_total` would become reconcile-subset-scoped. The right merge,
  and it goes in one place inside `merge_route_summaries`, is to take
  `pad_pairs_total` from the first summary (the full-scope denominator) and
  add the per-net deficit recovered, using `pad_pairs_open`'s per-net counts,
  since a net fully recovered in pass 2 is simply absent from pass 2's
  `pad_pairs_open`.
- **Known gap, not fixed here.** `failed_multipoint` entries from the
  non-multipoint multi-pad path are also zero-weight in the loop's
  `failures`, and nothing in the summary distinguishes them from the ones
  that do carry pad counts. Unchanged by these commits; happy to file it.
- **The docs table.** The kit-dev-coldfire round table in
  docs/placement-optimization.md was recorded under the old tally, so its
  failure counts are pre-reconciliation and its iteration counts cover the
  first pass only. The exact command chain behind those numbers was not
  recorded with them, so I annotated the table rather than substituting
  numbers from a chain that may not be the same one. Say the word and I will
  re-measure it under whatever chain you consider canonical.

### DIFF
--- a/docs/placement-optimization.md
+++ b/docs/placement-optimization.md
@@ -206,8 +206,9 @@ precedent), which is simpler and may capture most of the value:
   into the move generator so every perturbation is feasible, rather than
   penalizing violations):
   - respect `locked` flags (already parsed by `placement/parser.py`)
-  - `--max-displacement` radius from the seed position, so output reads as
-    "your placement, nudged" and inherited human constraints survive
+  - `--max-displacement` radius from the seed position — enforced for every
+    move type, swaps included — so output reads as "your placement, nudged"
+    and inherited human constraints survive
   - courtyard non-overlap via the existing rect machinery
   - decaps tethered to their IC (group membership or hard radius)
   - rotation disabled per-class where assembly conventions matter
@@ -273,7 +274,8 @@ that's the documented failure mode of wirelength-driven placement.
 
 Implemented as `place_optimize.py` + `placement/quench.py`: greedy quench
 (nudge within `--max-displacement` of the seed, 90° rotations with correct
-pad-angle rewriting, same-footprint swaps), cost = `length_weight`·airwire
+pad-angle rewriting, displacement-capped same-footprint swaps), cost =
+`length_weight`·airwire
 length + `crossing_penalty`·crossings + pin-count-scaled halo + soft edge
 margin. Test board: `interf_u` (25 parts, PGA120 + bus connectors, 2 layers),
 pipeline `route_planes` → `bga_fanout U9` → `route.py` with the
@@ -354,7 +356,10 @@ Refined conclusions:
 - `--max-displacement` is the dominant safety knob: small caps keep the
   human's macro structure (bus corridors, cluster geometry) intact by
   construction, which is the entire value of seeding from a human placement.
-  ~3 mm was the sweet spot on both boards tested.
+  ~3 mm was the sweet spot on both boards tested. The cap is now airtight:
+  no move type — swaps included — can take a part beyond it, and
+  `--swap-max-displacement` can tighten (never exceed) it for swaps
+  specifically.
 - The big halos backfire on dense boards: with `--halo-coef 0.5` the
   144-pin Xilinx demands a 6.5 mm halo that a dense board cannot satisfy, so
   the halo gradient dominates everything and scatters the layout. Modest
--- a/docs/placement-optimization.md
+++ b/docs/placement-optimization.md
@@ -143,7 +143,10 @@ crossing counter is the non-differentiable version of the same quantity —
 fine, since SA doesn't need gradients. The theory behind crossing counts is
 topological routing (SURF, DAC 1991): zero airwire crossings on a layer means
 a planar ratsnest, i.e. routable on that layer in the topological sense; every
-crossing costs a via pair or a detour.
+crossing costs a via pair or a detour. Because a crossing is a conflict
+*between two nets*, the natural way to prioritize one net is to price each
+crossing by the larger of the two nets' weights: the cost of a conflict then
+tracks the most important net it obstructs.
 
 For reference, RUDY (the standard VLSI congestion proxy) smears each net's
 expected wire area uniformly over its bounding box —
@@ -193,7 +196,9 @@ precedent), which is simpler and may capture most of the value:
   - *nudge*: translate within a window that shrinks as temperature drops;
     optionally bias toward the part's optimal region (median of connected
     nets' bounding boxes — O(pins) to compute)
-  - *rotate*: 0/90/180/270
+  - *rotate*: 0/90/180/270, relative to the part's own placement angle (a
+    45°-placed part rotates on the 45/135/225/315 lattice, so deliberate
+    non-orthogonal placement is preserved rather than snapped to the axes)
   - *swap*: same-footprint pairs (footprint-identical R/C swaps are free
     wins; mixed-size swaps are usually illegal anyway, so restrict by
     footprint compatibility)
@@ -212,6 +217,8 @@ precedent), which is simpler and may capture most of the value:
   - courtyard non-overlap via the existing rect machinery
   - decaps tethered to their IC (group membership or hard radius)
   - rotation disabled per-class where assembly conventions matter
+    (`--no-rotate`, which also restricts same-footprint swaps to equal-angle
+    pairs, since a swap exchanges full poses)
 - **Cost**: Δ(crossings)·penalty + Δ(airwire length) + density/halo term —
   all incrementally updatable (only airwires touching the moved part's nets
   change). Mostly already in `rust_placer`; extending the scorer to evaluate
@@ -277,7 +284,10 @@ Implemented as `place_optimize.py` + `placement/quench.py`: greedy quench
 pad-angle rewriting, displacement-capped same-footprint swaps), cost =
 `length_weight`·airwire
 length + `crossing_penalty`·crossings + pin-count-scaled halo + soft edge
-margin. Test board: `interf_u` (25 parts, PGA120 + bus connectors, 2 layers),
+margin. Both airwire terms take optional per-net weights: a net's length is
+multiplied by its weight, and each crossing is priced at the larger of the
+two crossing nets' weights, so an unweighted board is untouched
+(max(1, 1) = 1). Test board: `interf_u` (25 parts, PGA120 + bus connectors, 2 layers),
 pipeline `route_planes` → `bga_fanout U9` → `route.py` with the
 `tests/test_interf_u.py` arguments. Router iterations ≈ effort; vias and
 completion are the quality metrics.
@@ -359,7 +369,8 @@ Refined conclusions:
   ~3 mm was the sweet spot on both boards tested. The cap is now airtight:
   no move type — swaps included — can take a part beyond it, and
   `--swap-max-displacement` can tighten (never exceed) it for swaps
-  specifically.
+  specifically. The router-in-the-loop widening below widens **nudges only**:
+  the swap cap holds at its base value for the whole run.
 - The big halos backfire on dense boards: with `--halo-coef 0.5` the
   144-pin Xilinx demands a 6.5 mm halo that a dense board cannot satisfy, so
   the halo gradient dominates everything and scatters the layout. Modest
@@ -372,18 +383,34 @@ Refined conclusions:
 The proxy-only results above motivated closing the loop: use the *router's
 own failure diagnostics* to decide what to move. Each round:
 
-1. Route the board; parse the JSON summary (failed nets, iterations) and the
-   frontier blocking analysis (which nets wall off each failed route).
+1. Route the board; parse the run's JSON summaries and the frontier blocking
+   analysis (which nets wall off each failed route). Note the plural:
+   `route.py` runs an end-of-run reconciliation pass whenever the first pass
+   left failures, and prints a second summary scoped to the retried nets.
+   The failure lists in that last summary are the still-open set, while
+   router effort adds across both passes.
 2. Build the movable set: parts owning pads of the **failed nets** (move the
    endpoint out of the congested pocket) and of the **blocker nets** (move
    the anchor so the blocking wall re-routes) — excluding high-pin-count
    parts (`--max-target-pins 40`: moving a resistor that anchors a blocker
    is low-risk; dragging a 144-pin QFP is how placements get destroyed).
-3. Micro-quench only those parts, with failed nets' airwires weighted 3×.
+3. Micro-quench only those parts, with the failed nets weighted 3×
+   (`--failed-net-weight`) — both their airwire *length* and any *crossing*
+   they take part in, the latter priced at the larger of the two nets'
+   weights. Weighting length alone was near-inert: at the loop's own knobs
+   the crossing term is ~96% of the objective, so scaling a length
+   coefficient from 0.3 to 0.9 against a crossing coefficient of 30 moved a
+   typical failed net by less than the cost of a fifth of one crossing.
 4. Re-route. Accept only if (failures, then iterations) improves; otherwise
-   revert and widen the displacement cap 1.5× for the next attempt.
+   revert and widen the displacement cap 1.5× for the next attempt — the
+   nudge radius only. The swap cap stays at its base value, so a widened
+   round cannot become a long-range swap.
 
-Result on kit-dev-coldfire from the hand placement:
+Result on kit-dev-coldfire from the hand placement (recorded before #458;
+`failures` was then the pre-reconciliation count and `router iterations` the
+first pass only, so both columns would read slightly differently today: fewer
+failures where the reconciliation pass recovers a net, and more iterations
+because that pass's own effort is now counted):
 
 | round | action | failures | router iterations | vias |
 |---|---|---|---|---|
@@ -397,11 +424,12 @@ less router effort — by moving only resistors/caps/jumpers** (49 parts
 total; the ICs and connectors never moved). The reject-and-revert mechanism
 did real work: two of four candidate placements made things worse and were
 discarded, which is exactly the proxy-blindness the loop exists to catch.
-Costs: +36 vias, and same-footprint swaps are not displacement-capped (one
-resistor traveled 32 mm via swap — fine for generic pull-ups, but worth a
-cap or flag for parts where location is meaningful). DRC: the same
-PAD-SEGMENT micro-overlap artifact the router produces on the baseline (16
-occurrences) appears somewhat more often on the repaired board (30).
+Costs: +36 vias. (This run predates the swap displacement cap — one resistor
+traveled 32 mm via an uncapped swap, which is what #430 fixed. Swaps are now
+capped like every other move, and the loop holds that cap at its base value
+while it widens the nudge radius, #458.) DRC: the same PAD-SEGMENT
+micro-overlap artifact the router produces on the baseline (16 occurrences)
+appears somewhat more often on the repaired board (30).
 
 This validates the core hypothesis of the whole investigation: **proxies
 propose, the router disposes.** Wall-clock cost was ~5 routing runs
--- a/place_optimize.py
+++ b/place_optimize.py
@@ -6,10 +6,12 @@ Usage:
 
 Starts from the current (hand- or AI-made) placement and applies a greedy
 quench: small nudges, 90-degree rotations, and same-footprint swaps that
-reduce airwire length + crossings + a whitespace (halo) penalty. Each part
-stays within --max-displacement of its original position, so the output is
-"your placement, nudged" - implicit constraints encoded in the original
-placement survive. Locked footprints never move.
+reduce airwire length + crossings + a whitespace (halo) penalty. Every move
+type keeps each part within --max-displacement of its original position -
+same-footprint swaps obey the same cap (or the tighter
+--swap-max-displacement) - so the output is "your placement, nudged" and
+implicit constraints encoded in the original placement survive. Locked
+footprints never move.
 
 See docs/placement-optimization.md for the background research.
 """
@@ -40,6 +42,10 @@ Examples:
     parser.add_argument("--max-displacement", type=float, default=10.0,
                         help="Max distance a part may move from its original "
                              "position in mm (default: 10)")
+    parser.add_argument("--swap-max-displacement", type=float, default=None,
+                        help="Max distance a swap may move each part from its "
+                             "original position in mm; must not exceed "
+                             "--max-displacement (default: equal to it)")
     parser.add_argument("--step", type=float, default=1.0,
                         help="Candidate grid step in mm (default: 1.0)")
     parser.add_argument("--grid-step", type=float, default=defaults.GRID_STEP,
@@ -85,6 +91,13 @@ Examples:
 
     args = parser.parse_args()
 
+    if args.swap_max_displacement is not None:
+        if args.swap_max_displacement < 0:
+            parser.error("--swap-max-displacement must be >= 0")
+        if args.swap_max_displacement > args.max_displacement:
+            parser.error("--swap-max-displacement must not exceed "
+                         "--max-displacement")
+
     if args.output_file is None:
         base, ext = os.path.splitext(args.input_file)
         args.output_file = base + '_optimized' + ext
@@ -97,6 +110,7 @@ Examples:
         pcb_data,
         pcb_file=args.input_file,
         max_displacement=args.max_displacement,
+        swap_max_displacement=args.swap_max_displacement,
         step=args.step,
         grid_step=args.grid_step,
         clearance=args.clearance,
--- a/place_route_loop.py
+++ b/place_route_loop.py
@@ -121,7 +121,13 @@ def main():
                         help="Max repair rounds (default: 5)")
     parser.add_argument("--max-displacement", type=float, default=3.0,
                         help="Initial displacement cap per round in mm "
-                             "(default: 3; widened 1.5x after a rejected round)")
+                             "(default: 3; widened 1.5x after a rejected "
+                             "round - nudges only, never swaps)")
+    parser.add_argument("--swap-max-displacement", type=float, default=None,
+                        help="Displacement cap for same-footprint swaps in mm; "
+                             "must not exceed --max-displacement and is NOT "
+                             "widened between rounds (default: the initial "
+                             "--max-displacement)")
     parser.add_argument("--max-target-pins", type=int, default=40,
                         help="Don't move parts with more connected pins than "
                              "this (default: 40)")
@@ -142,10 +148,26 @@ def main():
     parser.add_argument("--grid-step", type=float, default=defaults.GRID_STEP)
     parser.add_argument("--ignore-nets", nargs="+", default=None)
     parser.add_argument("--lock", nargs="+", default=None)
+    parser.add_argument("--no-rotate", action="store_true",
+                        help="Disable rotation moves")
+    parser.add_argument("--no-swap", action="store_true",
+                        help="Disable same-footprint swap moves")
+    parser.add_argument("--verbose", "-v", action="store_true",
+                        help="Print each accepted quench move and the "
+                             "per-pass swap-capped count")
     parser.add_argument("--work-dir", default=None,
                         help="Directory for intermediate files "
                              "(default: alongside output)")
     args = parser.parse_args()
+
+    if args.max_displacement < 0:
+        parser.error("--max-displacement must be >= 0")
+    if args.swap_max_displacement is not None:
+        if args.swap_max_displacement < 0:
+            parser.error("--swap-max-displacement must be >= 0")
+        if args.swap_max_displacement > args.max_displacement:
+            parser.error("--swap-max-displacement must not exceed "
+                         "--max-displacement")
 
     work = args.work_dir or os.path.dirname(os.path.abspath(args.output_file))
     os.makedirs(work, exist_ok=True)
@@ -160,6 +182,14 @@ def main():
           f" vias={best['vias']}")
 
     max_disp = args.max_displacement
+    # The swap cap is pinned to the BASE displacement and never widened (#458).
+    # A rejected round widens the NUDGE radius, a local search around each
+    # part's own seed that a real re-route then validates. Widening the swap
+    # cap in lockstep turns a 3mm budget into a 15mm teleport after four
+    # rejections, which is the #430 stranding failure coming back through the
+    # loop. Swaps get one fixed budget for the whole run.
+    swap_cap = (args.max_displacement if args.swap_max_displacement is None
+                else args.swap_max_displacement)
 
     for rnd in range(1, args.rounds + 1):
         if best['failures'] == 0:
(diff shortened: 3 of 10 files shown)


edgehero:
Follow-up commit 17d4670: post-push review caught a soundness gap at the pruning frontier — an accepted swap inserts a partner's non-90° rotation into `bounds_by_rot` *after* `build_neighbor_lists` computed its union boxes, so on boards mixing 90° and non-90° seeds within one movable same-footprint group the inherited rect could protrude up to (√2−1)·half-extent beyond the pruning box. Seed boxes now union over the whole movable group's rotation set (locked parts contribute nothing — they never swap). Covered by the new `test_neighbor_lists_cover_swap_inherited_rotations`, including a differential check that locking the 45°-seeded partner restores the tighter exact exclusion. 10/10 quench tests pass.

drandyhaas:
Deep review done (with Claude's help) — all 11 commits verified sound, all 44 of the branch's tests pass, and it merges clean onto today's main (169f6b8) both textually and semantically (#513 item 5 touches neither the JSON_SUMMARY keys, the retry-set derivation, nor the abort message `merge_route_summaries` anchors on; the sub-run still prints its summary before the board write, so the raise-after-print fallback stays valid).

The two claims that most needed adversarial checking both hold:

- **Pruning exactness, end to end.** Every reachable position stays within `max_displacement + grid_step` of the seed (nudge candidates are seed-radius-checked then grid-snapped; swap destinations pass the seed-cap check), every reachable rotation is inside the recorded group lattice — and the 2ea4abc follow-up is genuinely load-bearing there, the exactness proof has a hole without it. The margin formula kills both interaction kernels exactly (halo returns exactly 0.0 beyond `halo_a + halo_b`, overlap needs separation ≤ clearance), so excluded pairs contribute an exact `+= 0.0`. The bit-identical parity test pins it on a real board.
- **Swaps skipping `candidate_valid`** is rigorous given the new bounds-equality guard: every rotation entry on both parts is the same pure function of the raw courtyard tuple, so equal bounds ⇒ the swap permutes the occupied-rect set without changing it.

Non-blocking observations:

1. `merge_route_summaries` deficit-unit shift: a pass-1 partially-routed multipoint net that produces no result at all in the sub-run lands in the sub-run's `failed_single` and weighs 1 instead of its pad deficit. Fair across candidates (every round is scored by the same rule), but the scalar isn't unit-stable — worth a comment.
2. The "all-ones weights leave the cost *exactly* unchanged" docstring overclaims by the 0.5 odd-total ccw-asymmetry corner the code itself documents; the real-board equality test is the honest version of the claim.
3. The differing-courtyard same-footprint-name swap refusal has no test (the other c7f95ba guards do).
4. The route.py line citations in place_route_loop.py comments drifted ~26–57 lines against current main — cosmetic.
5. Pre-existing, not yours: the loop's final `shutil.copy` of the board drops the sibling `.kicad_pro` (the #441 rule). Largely inert for placed-not-routed boards, but noting it.

**The #432 interaction: confirmed exactly as your review notes predicted.** On the combined tree (merges clean in either order, both PRs' tests pass together), `dict(summaries[-1])` makes `pad_pairs_*` reconcile-subset-scoped whenever a reconciliation ran (a 50/40 whole-board reading becomes the sub-run's 3/2, or 0/0 when the sub-run omits the keys), and a sub-run without a `blockers` key drops pass 1's attribution so `run_route` regresses to the whole-log regex. Harmless today since nothing consumes those metrics, but I've implemented your proposed reconstruction on a local integration branch: `pad_pairs_total` from the first summary, connected = that total minus the deficit still open in the last summary's `pad_pairs_open` (every pass-1 open net is in the retry set by construction, so absence means recovered), `blockers` preserved filtered to still-failed nets when the sub-run omits the key, and `run_route` treating an explicit empty blockers list as "nothing to attribute" rather than falling back to the regex. Your regex-union test still passes untouched on legacy logs. It will land together with the two merges.

drandyhaas:
Merged into the `placement` integration branch (along with #432 and the two follow-up commits from the review, including your proposed pad_pairs reconstruction in merge_route_summaries). Closing to keep the open-PR list clean — this is a merge, not a rejection; the branch will land on main after it soaks. Thanks for the thorough engineering and honest disclosures throughout.

Issue #498, opened 2026-07-26
Per-layer clearance override (inner vs outer), composed with net-clearance


GreenFoot:
**Summary** — The router resolves obstacle clearance globally (config.clearance) plus per-net cross-class spacing (net_clearances / obstacle_clearance(net_id), auto-read from .kicad_pro netclasses). There's no way to vary clearance **by copper layer** — a common DRC requirement: outer layers at a wider clearance (e.g. 0.30 mm) than inner (e.g. 0.15 mm). Netclasses can't express this (a net spans layers but carries one clearance); in KiCad it's a .kicad_dru rule with an A.onLayer('In1.Cu') condition, which the router doesn't read for obstacle sizing.

**Proposed feature** — an optional per-layer clearance override, keyed by layer name, composed by max() with the existing clearance at every track/via/pad site: effective(net, layer) = max(obstacle_clearance(net), layer_clearance[layer]). Most-restrictive-rule-wins (matches KiCad), so it composes with net_clearances rather than replacing it, and is a **no-op when unset** (empty map → 0.0 → identical to today).

**Natural source** — the board's own .kicad_dru per-layer rules, analogous to how route.py already auto-reads netclass clearances from .kicad_pro; with an explicit --layer-clearances F.Cu:0.30,In1.Cu:0.15 for scripting.

**Scope** — Python-only, no rust_router change.

**Offer** — I have a working, tested engine implementation ready; before a PR I'd like to confirm (1) is this wanted upstream, and (2) preferred surface (auto-read from .kicad_dru vs CLI flag vs both). I'd wire it end-to-end per CLAUDE.md and include before/after check_drc.py grading on a real board.

(Env: KRT v0.19.0, Python 3.11)

drandyhaas:
Implemented upstream in `92acb77` + `807ae83` — thank you for the well-shaped proposal, @GreenFoot; it prompted the design discussion that landed this. Answering your two questions with what shipped, including two deliberate divergences from the sketch:

[...]

PR #532, opened 2026-07-30
obstacle map: routed track under-delivers clearance near segments and via holes

bergmorten:
Two related under-deliveries in `build_routing_obstacle_map` that produce boards which route "clean" but fail KiCad DRC:

1. **Segment keep-outs miss the router's own discretisation.** The capsule is exact on the existing-segment side, but the new track routes to grid-cell centres, so its ±half-cell quantisation lets copper land up to `grid_step/2` inside the clearance ring — edge-to-edge DRC then fails by up to grid/2. The via and pad stampings already carry this cushion; segments now do too.

2. **Via keep-outs only enforce the copper envelope.** When the fab's copper-to-hole clearance exceeds the class clearance (e.g. JLC 0.2 mm hole vs 0.1 mm copper), the hole envelope (`drill/2 + hole_to_hole_clearance`) is larger than the copper envelope — a track kept only off the copper still grazes the drill and fails KiCad's `hole_clearance` check. The keep-out now takes the max of both envelopes.

Rebased over #498's cross-class `layer_clearance` (both fixes compose with it).

Measured on a real 6-layer 56 mm board (82 nets): KiCad clearance violations **273 → 2** with no routability loss.

### DIFF
--- a/plane_obstacle_builder.py
+++ b/plane_obstacle_builder.py
@@ -1064,9 +1064,15 @@ def build_routing_obstacle_map(
         # The exact capsule keep-out measures from the real segment, so the blocked
         # halo matches the true clearance envelope without the grid-rounding that used
         # to leave connection traces within clearance of signal copper (#146/#173).
+        # kad#69: the capsule is exact on the EXISTING-segment side, but the NEW
+        # track routes to grid-cell centres, so its own ±half-cell discretisation
+        # still lets it land ~grid/2 inside the clearance ring (edge-to-edge DRC
+        # then fails by up to grid/2). Add the same grid_step/2 cushion the via and
+        # pad stampings above already carry, so the delivered copper meets spec.
         seg_expansion_mm = (route_track_w / 2 + seg.width / 2
                             + config.layer_clearance(  # #498 (#434 cross-class)
-                                route_layer, config.obstacle_clearance(seg.net_id)))
+                                route_layer, config.obstacle_clearance(seg.net_id))
+                            + config.grid_step / 2)  # discretisation cushion (see docstring)
         _add_segment_routing_obstacle(obstacles, seg, coord, layer_idx, seg_expansion_mm)
         seg_count += 1
     if verbose:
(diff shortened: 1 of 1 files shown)


drandyhaas:
Thanks for the detailed PR and the measurement — this is a well-written report, and the segment-cushion observation in particular points at a real inconsistency (the pad stamp's comment even claims the segment stamp already carries the cushion, which it doesn't). After review we're going to close this rather than merge, for the reasons below — with a path forward for each half.

**Fix 2 (via hole envelope): wrong constraint, and it conflicts with the chain's DRC-floor policy.**

The KiCad rule involved is `hole_clearance` (copper-to-hole, board setup `min_hole_clearance`). The PR substitutes `config.hole_to_hole_clearance`, which is the *drill-to-drill* spacing (default 0.20 mm, the JLC hole-to-hole floor) — a different constraint with a different board value, so the keep-out over- or under-delivers depending on the board and never routes to the actual rule.

More fundamentally, the toolchain deliberately treats stock `min_hole_clearance` as aspirational, the same way it treats stock netclass clearances: the DRC writeback (`fix_kicad_drc_settings.py`, `build_targets`) clamps `min_hole_clearance` down to the routed copper clearance in the output `.kicad_pro`, so on the tool's own chain the violations this fix targets never fire. Could you say how the 273-violation baseline was graded — against the chain-written `.kicad_pro`, or against a stock/original project? If it was the latter, most of those 273 are the known "grading stricter than the route used" class rather than a routing under-delivery.

If your fab genuinely requires a copper-to-hole clearance larger than annulus + copper clearance (small-annulus vias can hit this), that's a real feature request — but it needs the board's actual `min_hole_clearance` read and threaded through *both* obstacle builders (`obstacle_map.py` stamps no hole envelope either, so signal tracks — the bulk of copper — would still violate), through `check_drc`, and through the writeback (which must then stop clamping below the enforced value). Happy to take that as an issue if it's a real fab constraint for you.

**Fix 1 (segment half-grid cushion): plausible bug class, but the magnitude claim doesn't hold and the routability cost is unproven.**

The router places track centerlines *on* free cell centers, which by construction are at least the full margin from the obstacle capsule — so copper can't land "up to grid/2" inside the ring. The real exposure is the chord dip between adjacent free centers and corner-cutting during smoothing, which is µm-to-tens-of-µm (the class the pad cushion fixed at 1.5 µm on a corpus board). Meanwhile the cushion costs `grid/2` (typically 0.05 mm — half a typical clearance) of keep-out on every foreign segment for plane taps in dense regions, and the main signal router carries no such segment cushion yet grades clean against kicad-cli across our stress corpus. We'd want a corpus A/B (kicad-cli violations down-or-flat, tap/connectivity counts flat) before paying that, and we'll also reconcile the misleading comment on the pad stamp either way.

If you can share the board (or a stripped repro) behind the 273→2 measurement, that would let us confirm which rule the violations actually fired under and re-evaluate — genuinely appreciated either way.

drandyhaas:
Please try things out on HEAD - a lot has been fixed recently.

bergmorten:
Answering your direct question, and conceding the substantive point.

**How the 273 baseline was graded:** against a **stock/tool-external** `.kicad_pro` — the one my generator (kad) emits from the design source, not the chain-written one. So you are right that a large share of those 273 fall into your "grading stricter than the route used" class rather than being routing under-delivery. I should have stated the grading basis in the PR; presenting the number without it made the claim look stronger than the evidence supports.

**Fix 2 was simply wrong** — I substituted `config.hole_to_hole_clearance` (drill-to-drill) for `min_hole_clearance` (copper-to-hole). Different constraint, different board value; the keep-out could only over- or under-deliver. Thank you for catching it rather than merging it.

On whether the underlying fab constraint is real for me: **yes, but not in the shape this PR assumed.** My boards use via-in-pad (JLCPCB FPOV) at 0.25/0.15 mm on 6 layers, i.e. a 0.05 mm annulus, where copper-to-hole genuinely becomes the binding rule rather than annulus + copper clearance. My generator now emits `min_hole_clearance`/`min_hole_to_hole` explicitly (0.15 mm at the JLC advanced tier) instead of leaving KiCad's 0.25 mm default in place. So the requirement is: read the board's actual `min_hole_clearance` and thread it through **both** obstacle builders, `check_drc`, and the writeback — exactly the scope you outlined, including `obstacle_map.py` stamping no hole envelope either. I'd rather that be your issue with your policy decision on the writeback clamp than a patch from me that fights the chain's DRC-floor policy; happy to supply the failing board and per-violation geometry if useful.

**Fix 1:** your centerline argument is correct and I have no corpus A/B to offer, so I'm not going to press it. If I can produce a reproducible case where the chord dip or smoothing corner-cut exceeds the pad-stamp scale, I'll open it as an issue with the measurement rather than a patch.

Also: apologies for the `.history` gitlink in all three PRs — VSCode local-history, now gitignored in the fork and absent from the resynced branch.

Closing loop on my side: I've rebuilt my fork on current `main` and dropped my superseded commits in favour of 81d1dda2 / bc8d77d0 / 93f3f2f9. Good catch on #545 — eleven more instances of the terminal-selection defect class is a far better outcome than my two-line patch.

Issue #515, opened 2026-07-29
Proposal: --force-reroute flag to rip + freshly re-route already-connected nets (#513 item 4)


drandyhaas:
Split out of #513 item 4 per review policy: new CLI parameters / GUI controls get filed as an issue first so the interface question can be decided separately.

## Problem (from #513)

`filter_already_routed()` removes a fully-connected net from the candidate set before `--rip-existing-nets` logic is ever consulted. Naming the net in *both* `--nets` and `--rip-existing-nets` is a silent no-op: `All nets are already fully connected - nothing to route!`, and an unchanged copy is written. Hit on 3 boards in the sets 21–25 wave (nascom_eprom_80bus re-widen, sipm_bias_lt8362 graze, ice4pi edge floor).

## Proposed interface

`route.py --force-reroute` (store_true): nets matched by `--nets` are ripped and freshly re-routed even when already fully connected. GUI: a \"Force re-route connected nets\" checkbox (control name `force_reroute`) on the Basic tab next to the rip/keep controls, plus `FLAG_PARAMS` mapping so recorded manifests convert.

Open interface questions:
- flag name: `--force-reroute` vs `--rewiden` vs overloading `--nets` ∩ `--rip-existing-nets` (the overload changes replay semantics of existing manifests, so the prototype used an explicit flag);
- whether `route_diff.py` should get the same flag (the sipm_bias case was a diff pair);
- interaction with `--keep-input-copper` (prototype: force wins for the forced nets).

## Prototype (validated, not committed)

A working implementation was built and validated in-session, then reverted pending this decision:

- Rip integrates with existing machinery: the per-net original snapshot predates the rip, so the #220 stale-input strip removes ripped copper from the output text automatically; the base obstacle map sees the vacated corridors.
- **Restore-on-failure**: a forced net whose re-route does not reconnect it gets its original copper back (rerouted partial copper removed from board + write-list, originals reinstated by identity) — never ships a previously-connected net open.

Validation results on splitflap_driver:
- without flag: silent no-op reproduced;
- with flag at `--track-width 0.5`: net fully re-routed at 0.5 mm (old 0.25 copper stripped), board DRC-clean and fully connected;
- forced failure (`--track-width 6.0 --max-iterations 200`, `KICAD_NET_RESCUE=0`): restore fires, output identical to input for that net, fully connected;
- `manifest_to_plan` carries `--force-reroute` → `force_reroute: true`;
[...]

PR #533, opened 2026-07-30
route: net in both --nets and --rip-existing-nets is re-routed (automates the #515 recipe)

bergmorten:
#515 declined a general force-reroute flag and added a warning documenting the manual recipe: *delete the net's copper first, then route it normally*. This PR automates exactly that recipe, for exactly the flag combination the warning fires on — naming a net in **both** `--nets` and `--rip-existing-nets` is an explicit request to replace its existing route.

- `filter_already_routed` gains `force_route_patterns`: rip-matched *selected* nets skip the already-connected grade and re-enter the route set.
- Their copper is pre-stripped from `pcb_data` before the originals snapshot, so the router replans from scratch; the stripped originals are merged into the writer's removal lists (the writer copies the input file verbatim, so pcb_data removal alone would still ship the old copper).
- Nets matched by `--rip-existing-nets` but **not** selected by `--nets` keep #515's blocking-rip-only semantics unchanged — no behavior change outside the previously-warned no-op case.

Motivation: a connected-but-DRC-violating route (ours squeaked past a 0.35 mm HV clearance corridor at 0.25 mm) previously required hand-editing the board file to get the router to touch it. This also unlocks a useful post-route quality pass: per-net isolated rip+reroute, keeping the result only when strictly shorter.

Test: `tests/test_rip_forces_reroute.py`; functionally verified on a 6-layer 82-net board (rip → replan → clean DRC).

### DIFF
--- a/route.py
+++ b/route.py
@@ -851,25 +851,36 @@ def batch_route(input_file: str, output_file: str, net_names: List[str],
         _write_passthrough_output(input_file, output_file)
         return 0, 0, 0.0
 
-    net_ids, _already_routed = filter_already_routed(pcb_data, net_ids, config)
-    # #515: --rip-existing-nets only rips copper that BLOCKS a net being
-    # routed; a net dropped here as already-connected never routes, so naming
-    # it in both --nets and --rip-existing-nets is a no-op. Warn instead of
-    # staying silent (the flag for a true rip+re-route was declined in #515;
-    # recipe: delete the net's copper first, then route it normally).
-    if rip_existing_nets and net_names:
-        from net_queries import matches_net_filter as _mnf_ripwarn
-        _rip_noop = [n for n, _reason in _already_routed
-                     if not _reason.startswith('Only')
-                     and _mnf_ripwarn(n, rip_existing_nets)]
-        if _rip_noop:
-            print(f"WARNING: {len(_rip_noop)} net(s) named in both --nets and "
-                  f"--rip-existing-nets are already fully connected and will "
-                  f"NOT be re-routed ({', '.join(_rip_noop[:6])}"
-                  f"{', ...' if len(_rip_noop) > 6 else ''}). "
-                  f"--rip-existing-nets only rips nets that block another "
-                  f"route; to force a fresh re-route, delete the net's copper "
-                  f"first and route it again (#515).")
+    # #515 follow-up: a net named in BOTH --nets and --rip-existing-nets is an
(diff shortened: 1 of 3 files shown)


drandyhaas:
Thanks — the diagnosis here is exactly right: `--rip-existing-nets` only rips *blocking* nets, so a connected-but-unwanted route named in both flags was a warned no-op with only a manual recipe, and that's a real gap (#515's decline notwithstanding).

Rather than merging this implementation, we've landed the feature as an explicit **`--force-reroute` flag** in 81d1dda2 (with a matching "Force re-route selected nets" checkbox in the GUI's Basic tab), which automates the same recipe with rails this PR's approach was missing:

- **Protected and locked copper are honored.** The pre-strip here bypassed the protected-nets machinery entirely: length-matched groups and routed diff pairs (recorded in the sibling `.kicad_pro`) would be ripped by any glob match, and KiCad-**locked** segments/vias — never rippable by design, with no override — would be stripped too. The landed version routes the strip through the same `protected_nets` gating as rip-up: exact name = deliberate override, glob = no override, locked = never. (Implementing that surfaced a latent hole in the *existing* rip-up override — the CLI expands `--nets` globs to exact names before the engine sees them, so a glob-selected protected net counted as "named exactly" — fixed in the same commit.)
- **Existing flag combinations keep their semantics.** Keying the force behavior off `--nets` ∩ `--rip-existing-nets` patterns silently rewrites recorded chains: `--nets ... --rip-existing-nets '*'` is the standard retry idiom, and under this PR every already-connected *and every partially-routed* selected net would be stripped to a from-scratch replan — a much wider change than the warned no-op case. An opt-in flag leaves all of that untouched.
- **A failed replan can't destroy a working route.** Stripping before the originals snapshot means a net whose re-route fails ships disconnected, minus the copper it had. The landed version restores the original copper (identity-preserved through the writer's custody machinery) whenever the replan produces nothing.
- Requires an explicit `--nets`/`--component` scope, skips plane (zone-owning) nets, and preserves each ripped net's original dominant track width.

Also note the PR accidentally commits a `.history` gitlink (VSCode local-history folder as a phantom submodule).

Your motivating case works directly now:

```bash
python3 route.py board.kicad_pcb out.kicad_pcb --nets "/HV_SENSE" --force-reroute
```

and the per-net "rip + reroute, keep if strictly shorter" quality pass you sketched is a nice idea — worth filing as its own issue if you'd like to pursue it. Thanks again for the report and the push on this; closing in favor of 81d1dda2.

PR #534, opened 2026-07-30
parser + endpoints: uuid-less copper parses; copper-group vias are route terminals

bergmorten:
Two connectivity-model fixes, found together on a real 6-layer board:

**1. `kicad_parser`: uuid-less segments/vias silently vanish.** The via/segment regexes require a `(uuid ...)` token, but KiCad itself accepts uuid-less copper, and tool-injected copper (stitch vias, escape stubs, scripts) often omits it. Such items are silently absent from the model — measured 52 of 203 vias invisible on one board, with the router then planning a track 0.014 mm from an invisible via barrel (a manufactured short that no stage could see). uuid is now an optional group in both patterns (rebased over the new `(locked yes)` handling — both compose).

**2. `get_net_endpoints` Case 1: copper-group vias are not terminals.** With two+ copper groups, only segment endpoints are emitted, each bound to its segment's layer. A group whose segments sit entirely outside `--layers` — canonical shape: an F.Cu pad stub ending in an escape via, routed `--layers In2.Cu In3.Cu` to duck under an HV corridor — contributes no endpoint at all and the net is unroutable, even though its via is reachable on every inner layer. Case 2 (one group + unconnected pads) has emitted vias as all-layer endpoints for years; Case 1 now does the same, with vias associated to their group by proximity to its segment endpoints.

Tests: `tests/test_parse_uuidless_copper.py`, `tests/test_group_vias_as_terminals.py`.

### DIFF
--- a/connectivity.py
+++ b/connectivity.py
@@ -1028,6 +1028,32 @@ def _get_net_endpoints_ordered(pcb_data: PCBData, net_id: int, config: GridRoute
                         if (gx1, gy1) != (gx2, gy2):
                             targets.append((gx2, gy2, layer_idx, seg.end_x, seg.end_y))
 
+                # #72: each group's VIAS are terminals on every routing layer
+                # (a via spans the whole stack). Without this, a group whose
+                # segments all sit on layers OUTSIDE config.layers -- e.g. an
+                # F.Cu pad stub + escape via, routed with --layers In2/In3 --
+                # contributes no endpoint at all and the net is unroutable
+                # even though its via is reachable on every inner layer.
+                # Case 2 below has done this for years; Case 1 never did.
+                def _group_vias(g):
+                    pts = set()
+                    for _sg in g:
+                        pts.add((round(_sg.start_x, POSITION_DECIMALS),
+                                 round(_sg.start_y, POSITION_DECIMALS)))
+                        pts.add((round(_sg.end_x, POSITION_DECIMALS),
+                                 round(_sg.end_y, POSITION_DECIMALS)))
+                    return [v for v in net_vias
+                            if any(abs(v.x - px) < 0.05 and abs(v.y - py) < 0.05
+                                   for px, py in pts)]
+                for _vlist, _side in ((_group_vias(source_segs), sources),
+                                      (_group_vias(target_segs), targets)):
+                    for via in _vlist:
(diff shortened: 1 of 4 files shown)


drandyhaas:
Both findings here are real, and this was a genuinely valuable report — thank you. We've landed both fixes in-house rather than merging, as bc8d77d0 (parser) and 93f3f2f9 (connectivity), which extend your patches in ways that needed the repo's history:

**uuid-less parsing (bc8d77d0):** your diagnosis was exactly right and reproduces on a two-line board. But the PR patched only the numeric-dialect patterns — the KiCad-10 `(net "name")` twins (`via_pattern_v10` / `segment_pattern_v10`) and both copper-arc patterns required uuid too, so a v10-style or arc-carrying board kept dropping copper. All six patterns are optional-uuid now. Bonus you'll appreciate: KiCad 6/7 wrote `(tstamp ...)` instead of `(uuid ...)`, so that entire era's boards previously parsed to **zero** copper — your fix generalized repairs that too.

**Case-1 via terminals (93f3f2f9):** confirmed as an oversight dating to the original import (Case 2 had the via block from day one; no comment ever defended the asymmetry). One correction to the report: on current HEAD the canonical scenario isn't strictly *unroutable* — full-stack normalization keeps the stub endpoints visible on the forbidden layer and the stub-swap rescue ladder saves the net — but it saves it *badly* (your repro shape measured: 208k iterations, 18 segments, 2 new via-in-pads, both existing escape vias unused, vs. 163 iterations and a direct 3-segment route with the fix). The landed version also position-sorts the via terminals (board order differs between the CLI text parser and the GUI's pcbnew build — a known GUI/CLI divergence class here) and matches group membership exactly to `find_connected_groups`' own via-join rule. Validated over a 15-board replay corpus: zero regressions, two boards strictly improved.

Your report also prompted a full audit of every other terminal-selection path for the same defect class — it found eleven more instances (vias/stubs invisible to multipoint tap routing, diff-pair endpoint selection, rescue sizing, and more), now being fixed under #545, with zone-anchoring tracked as #544. So this PR's impact goes well beyond its diff.

(Housekeeping for future PRs: this one, like #532/#533, includes an accidental `.history` gitlink — worth adding to your fork's global gitignore.)

Closing in favor of bc8d77d0 + 93f3f2f9. Thanks again — this was the best of the three reports.

Issue #189, opened 2026-06-24
route.py reports a multipoint single-ended net as routed when an MST edge fails, silently leaving pads disconnected


drandyhaas:
**Symptom:** route.py's multipoint (MST) single-ended router can mark a net as **routed** (`routed_single` in the JSON summary) even when one of its pad-to-pad MST edges **fails to route**, silently leaving the pads on that edge disconnected. The net passes the run's own accounting but `check_connected` reports it split into multiple components.

This is the single-ended-routing analogue of the plane-tap silent disconnect fixed in `ae2c069` (fill-aware plane sweep) and the diff-pair degrade-to-single-ended cases — same theme: **a router reporting success while leaving copper disconnected.**

**Concrete case — `glasgow_revC` (set 1), net `/IO_Banks/IO_Buffer_B/Z2` (id 95, 6 pads: J6.3, RN3.6, RN5.2, J3.7, D20.2, D20.13):**

`route_final.log`:
```
[7/50] Routing /IO_Banks/IO_Buffer_B/Z2 (id=95)
  Phase 1 edge (pads 3,4) failed after 1061 iterations - trying next MST edge
  FAILED: Could not find route (0.12s)
  ROUTE FAILED - no rippable blockers found
```
…yet the same run's `JSON_SUMMARY` lists `/IO_Banks/IO_Buffer_B/Z2` in **`routed_single`**.

`check_connected` on the routed board: net 95 has **3 disconnected components**; pads **RN5.2 (113.15, 106.50)** and **RN3.6 (105.45, 105.50)** are floating (nearest net-95 copper to RN5.2 is 2.24 mm away). So the MST edges reaching those two pads failed, but the net was still counted routed.

**Two problems (one per layer of the fix):**

1. **Reporting / accounting.** A multipoint net whose MST has a failed edge (or any pad left in a separate component) must be reported **failed**, not `routed_single`, so the disconnect is surfaced (and a retry / follow-up can act on it) instead of being lost. The authoritative check is the same union-find `check_connected.check_net_connectivity` uses.

[...]

PR #535, opened 2026-07-30
escape_stub.py: congested-pad micro-escape tool

bergmorten:
New standalone tool for the class of pad that `route.py`'s rescue gives up on: fine-pitch QFN pins and connector pads walled in by neighbor pads/tracks/via barrels, and ball-grid pads with no inter-pad trace corridor at all. A legal escape usually exists — this finds it and injects the stub + via; `route.py` then finishes the net normally (with #534, the escape via is a terminal on every routing layer).

- **via-in-pad attempted first** — on 0.5 mm ball grids (0.3 mm pads, 0.2 mm gaps) no trace fits between pads at any legal clearance; the pad itself is the only exit.
- Otherwise **BFS on the pad's layer** at the requested width/clearance to the first spot where a through-via is legal on *every* copper layer (or to existing same-net copper), emitting a merged rectilinear stub.
- Obstacle model: exact pad geometry via `check_drc.point_to_pad_distance` (rotation/shape aware), per-instance via sizes, per-net netclass clearances with cross-class max(A,B), and the drill hole-to-hole floor.
- `--replace-radius R` strips the pad's net's copper within R mm first (redo a bad stub); pads are never touched.

Usage: `python3 escape_stub.py board.kicad_pcb --pad U33.13 --output out.kicad_pcb`

Proven on a 6-layer board: two 0.4 mm-pitch parts and one 0.5 mm ball-grid part (with another IC mounted directly under it on the back side) that route.py+rescue could not connect — all escaped and routed to a clean KiCad DRC.

### DIFF
--- a/escape_stub.py
+++ b/escape_stub.py
@@ -1,0 +1,357 @@
+#!/usr/bin/env python3
+"""Congested-pad micro-escape (#73).
+
+route.py's rescue gives up on pads whose only exit is a multi-segment escape
+at board-minimum width/clearance -- fine-pitch QFN pins and connector pads
+walled in by neighbor pads, tracks, and via barrels. A legal escape usually
+EXISTS: a short BFS path on the pad's layer to the first spot where a
+through-via is legal on every copper layer (or to existing same-net copper).
+This tool finds it and injects the stub + via; route.py then finishes the
+net normally (the via is a terminal on every routing layer, #72).
+
+Usage:
+    python3 escape_stub.py board.kicad_pcb --pad U33.13 --output out.kicad_pcb
+    python3 escape_stub.py board.kicad_pcb --pad J50.30 --replace-radius 6 \
+        --trace-width 0.1 --via-size 0.25 --via-drill 0.15
+
+--replace-radius R strips the pad's net's existing segments/vias within R mm
+of the pad first (redoing a bad hand stub); the strip never touches pads.
+
+Obstacle model: same-layer foreign segments and pads (exact pad geometry via
+check_drc.point_to_pad_distance -- rotation/shape aware), all foreign via
+barrels at their PER-INSTANCE size, all drill holes (hole-to-hole floor), and
+per-net clearances from the board's netclasses (cross-class max(A,B)).
+"""
+from __future__ import annotations
+
+import argparse
+import math
+import os
+import sys
+from collections import deque
+from typing import Dict, List, Optional, Tuple
+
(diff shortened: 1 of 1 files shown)


drandyhaas:
Thank you @bergmorten — this was a genuinely good find, and the capability is now **in the engine** as of 01bdba8 rather than a standalone script, so it fires automatically (CLI *and* GUI) whenever a boxed pad needs it.

**What we took, with credit:** your core insight — *route.py's rescue gives up on pads whose only exit is a short min-width escape stub to the nearest legal through-via site, and such an escape usually exists* — is exactly right, and it's now an escalation rung inside the #189 rescue. When the in-pad via fails, the same fab-ladder loop retries with a nonzero search radius (`KICAD_ESCAPE_STUB_RADIUS`, default 1.0 mm, 0 disables): the plane-tap machinery finds the first spot where a through-via is legal on every layer and routes the pad→via stub at fine grid / capped clearance. Your ordering insight (via-in-pad first, then off-pad) is preserved.

**Why absorbed rather than merged as-is:** the engine already contained a strictly stronger implementation of the search+stub mechanic — the plane-tap machinery (`tap_pad_with_escalation`) — with an obstacle model the standalone script couldn't match: zone copper, `.kicad_dru` per-layer rules, board-edge clearance, NPTH overrides, per-instance via sizes, fab-ladder size escalation, and (new in v0.20.1) the #581 same-net pad via keep-out. Wiring the rung there was ~100 lines and inherits all of it; with `--same-net-pad-clearance > 0` (where via-in-pad is forbidden outright) your off-pad escape is now the *only* compliant rescue, so the two features compose.

**Measured on your failure class** (dense fine-pitch, rescue-heavy): neo6502 with via-in-pad forbidden went from 43 failed nets to 32 — the rung fired 37 times and recovered 11 nets at ~1.4 s per firing (+12% wall on that worst case, ~0% on boards where it never fires). Boards that previously needed the standalone tool should now just route; if you still have the 6-layer board from the PR description and it doesn't, that's a bug report we very much want.

The `--replace-radius` redo workflow is covered by `--force-reroute` / the rip machinery, and `--pad U33.13`-style manual targeting isn't needed once the rescue fires automatically — but if you hit a pad the rung still misses, please open an issue with the board.

Closing since the functionality landed via 01bdba8. Thanks again for the contribution and the proof-of-existence on real boards — it's what moved this from 'known gap' to 'shipped'.

Issue #485, opened 2026-07-25
Via stitching: area lattice + board-edge fence for plane nets (as part of / after the route_planes step)


drandyhaas:
## Summary

We have no **area via stitching**: a deliberate, periodic lattice of vias bonding
same-net plane copper across layers, plus a board-edge via fence. Everything we
place today is *demand-driven* (tie this pad in, fix this disconnection) or
*pairwise* (a companion via next to one signal via). That covers connectivity but
not the signal-integrity / EMI / thermal reasons real boards are stitched.

The natural home is the **route_planes step** (or a new opt-in pass invoked from
it), because the plane engine already owns every constraint a stitching via must
respect: the pour geometry, keep-outs (incl. footprint-nested, #478),
`copperpour` keepouts, fab floors (hole-to-hole, edge clearance), the
`ZoneFillModel` min-thickness pinch model, and the "don't box in a protected-net
pad" rule (#180).

## What exists today (so we don't rebuild it)

| Mechanism | Where | What it does | Trigger |
|---|---|---|---|
| Plane **pad taps** (confusingly called "stitching vias" throughout `route_planes.py`) | `plane_pad_tap.py`, `route_planes.py` | One via (+ optional short trace) tying a target pad into the pour | Per *pad*, always |
| **Region/island bonding** | `plane_region_connector.py` | Unions same-net islands through same-net vias/plated barrels and joins what's still split | Per *detected disconnection* |
| **`--add-gnd-vias`** | `add_gnd_vias.py`, post-pass in `route_planes.main()` (~L3800), GUI planes tab, `claude_plan` param | Places one GND via within `--gnd-via-distance` (2.0mm) of each existing **signal via** that hasn't got one | Opt-in, per *signal via* |
| **Diff-pair GND vias** | `route_diff.py` / `length_matching.py` (`gnd_via_enabled`, default True) | GND vias flanking pair layer transitions | Per *pair transition* |

So: multi-layer same-net pours *do* get bonded, but only wherever a pad tap or a
repair join happens to land. There is no pitch target, no coverage metric, and no
via at all in a large pour region that contains no pads.

## What's missing

1. **Area lattice stitching.** Fill the interior of each same-net multi-layer pour
   with vias on a pitch, so the layers are bonded as a plane pair rather than at
   scattered points. Pitch should be specifiable directly (mm) and ideally
   derivable from a frequency target (the usual λ/20-at-f_max style rule) using
[...]

Issue #424, opened 2026-07-17
Connectivity roadmap: remaining levers after the #217/#412 wave (detailed follow-up plan)


drandyhaas:
Roadmap issue: the remaining connectivity levers after the 2026-07-16/17 wave (#217/#412), with enough design detail per lever to execute each one well later. Baseline after the wave: incomplete nets 284 → ~202 on the comparable corpus (−27%), kicad-matched real DRC 118 → 37, coupling ~66%. The residual tail is mostly architectural (layer budget, escape ceilings), which is what these levers target.

## Lessons that bind all of this (from the wave)

- **Prove a manual win on 2–4 real boards before building any automatic lever.** The failed-first restart was fully engineered (~660 LOC, keep-best, flag threading), was never-worse by construction — and never won on a real board (castor, butterstick, ddr5, glasgow; twice actively vetoed). It was reverted; ordering survives only as a manual tune-mode knob. Ordering-diversity ideas (including the "shake" rung below) must clear this bar first.
- **Weak-early copper must not outrank strong-late routing.** The diff SE-member fallback recovered ulx3s (+all 14 members) but starved zynq (−12 nets) because fallback copper became protected input to the signal stage (#421). Any lever that commits copper early needs a custody/rippability story for later stages.
- **Honesty first.** The T5 custody work showed silent failures (ghost-copper grading, erased failure reports) mask the real tail. Every lever must ship with per-net attribution in JSON_SUMMARY so the next triage can see what it did.
- **Grading regime changes look like regressions.** The pr392 declamp exposed ~1.9k pre-existing class violations. When a lever changes what is measured, split "new copper badness" from "newly visible" explicitly (kicad_matched is the cross-regime metric).

### A1. #421 — SE-fallback casualty rippability (fix option 1)
Emit the fallback-routed net names (already in `JSON_SUMMARY.single_ended_fallback`) AND mark them on the board (board-level property or group), so the downstream signal stage treats same-chain fallback copper as freely rippable instead of protected input copper. Validation: zynq must recover most of its −12 while ulx3s keeps its +14; picodvi/hid_remapper wobbles gone. This is the highest-value single item (fixes the wave's one real per-board regression class).

### A2. #423 — coarse-grid-far-from-obstacles (parked, design agreed)
[...]

PR #536, opened 2026-07-30
route-quality toolset: smooth_route, shorten_pass, stitch_planes

bergmorten:
Grid A* output is DRC-clean but maze-router-raw: staircase micro-jogs at every avoidance, hairpin detours that outlived the congestion that caused them, dead-end stubs, and pour pockets that island removal deletes into bare voids. Three post-route tools that fix copper *quality* while provably preserving DRC:

**`smooth_route.py`** — clearance-checked any-angle shortcutting (greedy farthest-visible-vertex) collapses staircases and meanders. Chains split at junctions, same-net via taps (mid-span too), pad touches, cross-width soft joints, and width changes (power neckdowns keep their width). Sequential obstacle model (already-smoothed nets are obstacles at their new geometry); exact pad geometry; per-net class clearances with cross-class max. `--trim-vias` removes dangling vias with a copper-overlap touch test, stack-aware union for co-located vias, barrel-junction bridging (a via joining two same-layer tracks inside its barrel is removed *and* the contacts bridged), and pad-anchored feeder cleanup.

**`shorten_pass.py`** — per-net isolated rip+reroute (route.py with the net in both `--nets` and `--rip-existing-nets`, see #533), keeping the result only if strictly shorter. Kills hairpins; worst case a net stays as it was.

**`stitch_planes.py`** — `--stitch-pads` places via-in-pad plane stitches for plane-net SMD pads; `--tie-pockets` fills zones via pcbnew with island removal disabled and drops a clearance-checked via inside every would-be-removed pour pocket, so the copper is kept (return paths, shielding, copper balance) instead of shipping voids. Complementary to #485's area lattice: that fences edges/areas inside route_planes; this targets pad bonding + island rescue on finished boards.

Measured on a 6-layer 82-net board: 200 chains smoothed (−81 mm copper), 78 junk segments + dangling vias removed, 14 nets shortened (−62 mm), 19 pour pockets kept — KiCad DRC at 0 unconnected / 0 violations through every stage.

**Stacked on #534** (the tools' writers emit uuid-less copper; without the parser fix, downstream stages silently drop it). Test: `tests/test_smooth_route.py`.

### DIFF
--- a/connectivity.py
+++ b/connectivity.py
@@ -1028,6 +1028,32 @@ def _get_net_endpoints_ordered(pcb_data: PCBData, net_id: int, config: GridRoute
                         if (gx1, gy1) != (gx2, gy2):
                             targets.append((gx2, gy2, layer_idx, seg.end_x, seg.end_y))
 
+                # #72: each group's VIAS are terminals on every routing layer
+                # (a via spans the whole stack). Without this, a group whose
+                # segments all sit on layers OUTSIDE config.layers -- e.g. an
+                # F.Cu pad stub + escape via, routed with --layers In2/In3 --
+                # contributes no endpoint at all and the net is unroutable
+                # even though its via is reachable on every inner layer.
+                # Case 2 below has done this for years; Case 1 never did.
+                def _group_vias(g):
+                    pts = set()
+                    for _sg in g:
+                        pts.add((round(_sg.start_x, POSITION_DECIMALS),
+                                 round(_sg.start_y, POSITION_DECIMALS)))
+                        pts.add((round(_sg.end_x, POSITION_DECIMALS),
+                                 round(_sg.end_y, POSITION_DECIMALS)))
+                    return [v for v in net_vias
+                            if any(abs(v.x - px) < 0.05 and abs(v.y - py) < 0.05
+                                   for px, py in pts)]
+                for _vlist, _side in ((_group_vias(source_segs), sources),
+                                      (_group_vias(target_segs), targets)):
+                    for via in _vlist:
+                        gx, gy = coord.to_grid(via.x, via.y)
+                        for layer in config.layers:
+                            layer_idx = layer_map.get(layer)
+                            if layer_idx is not None:
+                                _side.append((gx, gy, layer_idx, via.x, via.y))
+
             if sources and targets:
                 return sources, targets, None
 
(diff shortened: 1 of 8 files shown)


drandyhaas:
Thanks — the core idea here was measured and adopted, though heavily reshaped, and the other two tools were declined. Closing with the rationale for each part.

**`smooth_route` → shipped as an in-engine cleanup pass** (`smooth_octolinear_chains`, commit lands with the next push). Differences from this PR's shape, and why:

- **Octolinear output only** — staircases collapse to a single 45° diagonal + axis run via the existing `_octolinear_bends`, never any-angle chords. Keeps the copper compatible with everything downstream that assumes grid-friendly geometry, and the diagonal is where nearly all the length win is anyway (any-angle buys at most ~8% more).
- **Inside the shared cleanup pipeline** (pass 9b, after width-neck, before `close_soft_joints`) rather than a standalone script — both the CLI and GUI fronts get it for free, the write-list/board custody contract is enforced by the pipeline (`KICAD_BOARD_LEDGER` audits green), and the #534 uuid-less-copper dependency disappears because it authors through the engine's own writer.
- **Opt-in via `KICAD_SMOOTH_ROUTE=1`, intended for the final chain step only.** The shortcut is hard-clearance-checked but soft-cost-blind, so copper it moves can squat in corridors a later routing step's soft model deliberately kept open.
- **Extra guards the PR version didn't have:** protected/length-matched/diff-pair/impedance nets are skipped (meanders and pair coupling are the spec); `.kicad_dru` per-layer rules are honored; and foreign-net **pour outlines are hard connector keep-outs** — testing on a corpus board (anyshake) showed a DRC-legal shortcut through a surface pour can sever the pour's only corridor to a pad, invisible until refill. The per-net connectivity guard also runs without pour credit for the same reason.

Measured on 8 finished stress-corpus boards: **397 mm of 46.9 m routed copper (0.85%) collapses**, zero connectivity regressions, KiCad-clean before and after — same ballpark as the numbers reported here once pour-covered layers are excluded for safety.

**`shorten_pass` — declined**: `--force-reroute` (shipped earlier from #533's motivation) already automates per-net rip+reroute; a keep-if-shorter loop on top optimizes the wrong objective (longer routes often pay deliberate soft costs — bus decoupling, pad windows).

**`stitch_planes` — declined**: `--stitch-pads` duplicates the plane-tap machinery, and `--tie-pockets` overlaps the #485 stitch lattice and #424 plane drops/fragility work; pour-pocket rescue belongs inside the plane step if it proves worthwhile.

https://claude.ai/code/session_0176bS9zFYRSx67KWmRpVPuD

Issue #521, opened 2026-07-30
Length matching is not an invariant across chain steps: retry steps rip matched groups and reroute without matching (allwinner_h3_ddr3 left 40/41 group nets unmatched, SA11 disconnected)


drandyhaas:
Found while explaining "I don't see any meanders in runs_set24/allwinner_h3_ddr3" (#501/#520 wave work). The board's step4 ran `--length-match-group auto` and placed meanders — then the chain destroyed them:

**Step 4 (route.py, `--length-match-group auto`)** matched the DDR groups and placed bump trains (log shows 14/34/10/16-bump runs).

**Step 4b (retry)**: `--nets <7 failed DDR nets> --rip-existing-nets '/DDR3 16x1/*'` — rips the ENTIRE DDR match group as potential blockers, then routes only the 7 failed nets with **no `--length-match-group`**. Steps 4c/4d/4e route more failed nets, also without matching. Nothing after ever re-matches.

Measured on the recorded run (`step4_signal.kicad_pcb` vs `step5f_planeverify3.kicad_pcb`), 40 of the 41 group nets changed length, swinging both directions by tens of mm:

```
/DDR3 16x1/SA11    76.74 ->  0.00   (-76.74)   <- DISCONNECTED on the final (0 segments, 2 pads)
/DDR3 16x1/SA0    105.21 -> 37.73   (-67.49)
/DDR3 16x1/SA3     61.59 ->  0.07   (-61.51)
...
/DDR3 16x1/SBA0     4.50 -> 57.80   (+53.30)   <- failed at step4, routed later unmatched
```

The final board's DDR "match groups" are pure fiction — spreads of tens of mm — while every individual step reported success. SA11 is the net-coverage invariant violated concretely: ripped by the group glob, not in the retry's `--nets`, rip-rescue didn't reclaim it, no later step did either (this run predates several rip-rescue fixes; the ab_meander_0729 replay at HEAD will show whether current code still strands it).

Two distinct problems:

1. **Matching is per-step, chains are multi-step.** Any retry/rip step over matched nets silently voids the earlier step's matching. Candidate directions (design decision):
   - retry guidance (planner prompts / RUNBOOK / plan-pcb-routing skill): a retry touching nets of a matched group must carry the same `--length-match-group`;
   - engine warning: `route.py` warns when `--rip-existing-nets` rips copper carrying meander trains (or nets that formed a match group by the auto-grouper) while the step has no `--length-match-group`;
[...]

Issue #459, opened 2026-07-21
placement: block-scale relocation (the gap between 3mm nudges and manual re-floorplanning)


edgehero:
Follow-on from #407 and the placeholder in #110, filing the mover itself since nothing tracks it.

The quench fixes what it can reach: passives within a few mm. The failures that actually sink a board are block-scale. On my 10-layer board the magnetics block sat 80mm from its endpoints and about 22 nets were unroutable because of the floorplan, not the routing. Nothing in the toolchain lives between "nudge a passive 3mm" and "a human re-floorplans for days".

What I think this wants to be, based on the dose-response I saw (at a 10mm cap the quench destroyed a working corridor rather than fixing the real problem):

1. Diagnosis picks the movers, replacing pin-count filters. Use the routing evidence (the #409 blocker JSON once #432 lands: per-net blocked counts, cells, coordinates) plus geometry (connectivity-centroid displacement, share of frontier-blocked cells owned) to flag the block that must move. Never touch the rest. --max-target-pins is a proxy for this and picks wrong exactly when it matters, because the part that needs to move is never a passive.

2. Relocation is a constraint problem, not an optimization problem. Move only diagnosed blocks; keep the relative order of everything else as a hard constraint; minimize displacement. A bounded "relocate this one block among frozen neighbors" solve stays tractable at PCB sizes; global solvers do not. The ASIC incremental-macro work (diagnose which macros are badly placed, then solve over ordering constraint graphs) is the working precedent, and the constraint graph doubles as the explanation of why a block sits where it does, which is exactly what an AI director needs to justify a move.

3. Groups need to exist first. Parse KiCad `(group ...)` and footprint sheet paths; fall back to net-name-prefix clustering plus bus detection. docs/placement-optimization.md:200-213 already names rigid group moves, decap tethers and side-flip as intended scope, so this is filing the plan of record, not inventing one.

4. The router stays the judge. Every candidate goes through route-and-measure accept/revert, ratchet only, byte-identical revert.

#411's perturbed corpus is the harness that grades all of this (displace known-good blocks 5/10/20/40/80mm, measure recovery). Sequencing I'd propose: groups plus rigid translate moves in the quench first (the mild case), the bounded relocation solve second (the 80mm case).

edgehero:
**Assessment vs. the OmniEDA benchmarks (context in #411) — checked; no meaningful influence.**

[...]

Issue #439, opened 2026-07-17
route.py/route_diff route BELOW the board's own Default netclass clearance when --clearance is smaller (daisho 499, zynq 355 DRC)


drandyhaas:
## Bug

`route.py` / `route_diff.py` auto-read non-Default netclass clearances (PR392) but **`net_clearance_map_by_id` only returns nets whose class clearance EXCEEDS the Default class clearance**. So when the plan's `--clearance` is SMALLER than the board's own **Default** (and everything-equal) netclass clearance, the map is empty, `config.clearance` (the small routed value) is used for every pair, and the router lays copper below the clearance the board itself enforces. KiCad then flags essentially every trace.

This is the **single largest DRC class in the whole HEAD wave** — bigger than every sets-1+2 item combined.

## Repro (object-level: routed final vs unrouted baseline)

| board | routed `--clearance` | board netclasses | router-introduced `clearance` DRC | required | actual |
|---|---|---|---|---|---|
| daisho (set3) | 0.1 (16 of 17 steps) | Default/45Ω/50Ω all **0.127**, 674 patterns | **499** | 0.127 | 0.100–0.126 |
| zynq_ad9364 (set4) | 0.05 / 0.09 / 0.1 | Default/ADDRES/DQ*/All Nets all **0.2** | **355** | 0.200 | < 0.2 |
| kuchen (set4) | (low) | (0.x) | 87 | — | — |

daisho's 499 are adjacent-bus signal↔signal grazes (DQ8↔DQ9, CLK_N1↔CLK_P1, D37↔D39); zynq's 355 are DDR3 signal↔power + signal↔signal. Both unchanged vs the 0716a baseline (daisho was 499 there too) — long-standing, not a regression.

## Root cause

`net_clearance_map_by_id` (list_nets) returns `{}` for a net whose resolved class clearance is `<=` the Default class clearance — correct for cross-class *foreign obstacle* pricing (a net at Default needs no extra), but it means the routed net's OWN class clearance never floors `config.clearance`. Nothing enforces `config.clearance >= board Default netclass clearance`. A plan (or the #153 tune loop) that picks `--clearance` below the board's Default silently produces a fully DRC-violating board.

## Fix direction

[...]

PR #542, opened 2026-07-30
#459 follow-on: route, preview and undo one placement block (+ a GUI plan-reset fix)

### LINKED ISSUE #431: placement: headless PNG stills of placement status per place_route_loop round
The placement direction (#110, #407) has a review problem the routing side doesn't have. A routed board is self-evidently inspectable: open it in KiCad, the tracks are there. A *placement repair* is a set of position deltas whose merit is invisible without the context that motivated them: which airwires were failing, which nets walled them off, what moved and how far. `docs/placement-optimization.md` names the acceptance criterion itself, *"output that is visibly 'your placement, nudged' can be reviewed in minutes"*, but today no artifact shows that. Reviewing a `place_route_loop.py` round means diffing two `.kicad_pcb` files by hand.

## What already exists

- `pygame_visualizer/` renders the router live, but is interactive-only. No image export path.
- The one headless-rendering precedent in-repo: `animate_fanout_clearance.py` (pygame surface + PIL GIF encode, lines 102-103, both deps it already requires) driven by the engine's `on_move` hook. It proves the pattern and already draws exactly the right vocabulary: part rects, keepouts, ghost seeds.
- `place_route_loop.py` already writes per-round artifacts to the work dir: `loop_roundN.kicad_pcb` plus route logs (:153, :206). Everything a renderer needs is on disk after every round.
- The airwire data is one helper call away: the quench's own MST airwires (`compute_mst_edges`, `placement/quench.py:57-66` / `connectivity.py`), courtyards via `placement/parser.py`.

## Ask

A `render_placement.py` that takes a `.kicad_pcb` (plus optionally a "before" file and a route log / JSON summary) and writes PNGs:

- board outline + courtyard rects, locked parts dimmed
- airwires, with **failed nets** in one highlight colour and **blocker nets** in another (attribution comes from the log scrape today; becomes cleaner if/when #409 lands, but the renderer is independent of it)
- for every moved part: ghost rect at the seed position + a displacement arrow. The placement diff, literally drawn.
- **delta-first by default**: draw the moved parts and their involved nets prominently, everything else as faint context. On a big board the interesting subgraph is a few dozen parts, not all of them.
- **focused crops, not just the global frame**: a `--focus` mode that emits one cropped panel per failed-net cluster (the failure coordinates are already in the JSON summary / diagnostics), alongside the full-board overview. This is the difference between an artifact that works on a 25-part demo and one that works on the boards where placement repair actually matters.
- **per-side panels** (F.Cu / B.Cu) rather than one flattened projection, since placement lives on two sides
- **a metrics caption strip per frame**: round number, failures, router iterations, vias, accepted/rejected, so the picture carries its own verdict
- `--per-round`: the above per `loop_roundN.kicad_pcb` in a work dir, so a whole repair run reads as a filmstrip

## Known limits: what a PNG can and can't do

Stating these up front so the artifact doesn't oversell:

1. **Dense boards don't fit one readable frame.** On a 10-layer, ~290-part board (the #407 case) a full-board still with every airwire is exactly the ratsnest hairball that's already useless inside KiCad. A single global PNG degrades with board complexity, which is why delta-first rendering and failure-centred crops are in the ask as first-class requirements, not polish. Without them the tool demos well on `interf_u` and fails on the boards that motivated it.
2. **A flat image projects away the third dimension.** Airwires have no layer; blocker geometry (once #409 exposes cells) is per-layer; escape/fanout congestion under a BGA is a depth phenomenon. Per-side panels recover some of it, but a PNG fundamentally cannot show *why* an inner-layer corridor is full. For that, the structured diagnostics are the real interface. The render is a summary of them, not a replacement.
3. **Geometry is not causality.** Arrows show *what* moved; they don't show *why* it helped. A frame with big displacements and a frame with tiny ones can be equally good or bad. Only the caption strip (failures/iterations before vs after) carries the verdict; a render without the metrics invites exactly the wrong review heuristic ("lots moved, looks broken" / "barely moved, looks safe").
4. **These renders are for human eyes, not machine ones, for now.** The obvious temptation is to point a vision model at the PNGs and ask "what should move?". Current published evaluations say that's premature: frontier multimodal models still misread board layouts badly (layer-assignment bias, no reliable grasp of functional grouping or hierarchy; see OmniLayout, arXiv:2607.03261, and PCB-Bench, ICLR 2026). An agent in the loop should consume the #409 JSON, where every fact is exact; the render is the human-facing summary of the same data. If someone wants to test VLM-on-render as an experiment, the #411 perturbed corpus is the right harness for it, but nothing in this issue depends on that working.
5. **Not a KiCad replacement.** Final judgment on a placement still means opening the board. The render's job is triage: decide in seconds whether a proposed repair is worth opening at all.

## Why

1. **Review in seconds, at the scale where it counts.** "35 small parts moved near the failed/blocker nets" becomes a focused crop with arrows and a caption, instead of two boards open side by side. This is the trust story the doc's own research settled on: perturbative output is acceptable *because* it's reviewable, but only if there's something to review.
2. **Issues, PRs, and the corpus.** The #411 perturb-and-recover experiments produce dose-response results (5/10/20/40/80 mm) that want before/perturbed/recovered stills; routing already has `routed_all.png`-style artifacts for this, placement has nothing.
3. **The Claude tab / plan flow** (docs/claude-skills.md) gets a board-state artifact it can surface after a placement step, the same way routing steps have inspectable output, with the machine-consumption caveat in limit 4 above.

## Implementation sketch

Pure-PIL drawing (rects, lines, polylines, text) keeps it headless with zero new deps; no pygame import needed outside the existing animator. Reuse `animate_fanout_clearance.py`'s world-to-pixel mapping; parse via the existing `kicad_parser` + `placement/parser` courtyard extraction. Crops and per-side panels are the same draw calls with a different viewport/filter, so the scale-awareness costs little extra code. Roughly the animator's footprint (~250 LoC). Happy to PR.

### LINKED ISSUE #548: placement: align parts to their peers, and rotate them toward the nets they serve
The quench moves parts to reduce airwire length, crossings, halo and edge cost. Nothing in it rewards a placement *looking* like a human made it, and two specific things a human does constantly are absent:

## 1. No alignment notion exists

Verified, not assumed: `snap_to_grid` (`placement/utility.py`, used at `quench.py:1003`) snaps to the **routing grid**, which is unrelated. There is no row, column, shared-axis, or even-spacing concept anywhere in `placement/`. Grep for `align`, `row`, `column`, `tidy` returns nothing but unrelated prose.

So a decap row that a human placed on one y-coordinate comes out of a quench pass with each cap on its own slightly different y — every one of them an improvement by the cost function, and collectively a mess. The cost function is indifferent between "four caps on a line" and "four caps scattered within 0.3 mm of that line", because the airwire length is nearly identical. A human is not indifferent, and neither is the person reviewing the board.

**What it would take:** a snap-to-peer term, or a post-pass that collapses near-collinear same-footprint parts onto a shared axis when doing so costs less than some epsilon. It has to be cost-aware rather than cosmetic-only, or it will happily undo a move the router needed.

## 2. Rotation is not orientation-aware

`quench.py` rotates on each part's own angular lattice (`orig_rot` at `:236`, `group_rots` at `:920`) so a part placed at 45° rotates to 135/225/315 rather than snapping to the axes — that part is right. But the cost is `length + crossings + halo + edge` (`total_cost`, `:765`), and none of those terms know *which end* of a part its nets leave from.

A two-pad part whose net runs to the left is scored identically whether its net-1 pad faces left or right, as long as the centroid is where it is — the pad-level airwire difference across a 1 mm part is noise next to a 20 mm net. So the optimizer will not turn a cap around to shorten the connection it exists to make, and on a review render you see rows of parts pointing whichever way they happened to start.

**What it would take:** score airwires from the actual pad the net lands on (the geometry is already there — `_Part.pad_globals()` returns per-pad positions with net ids), and let a rotation candidate be evaluated on that rather than on centroid distance. That is a change to the cost kernel, so it needs the bit-identity suites re-baselined deliberately.

## Why this is its own issue

Both are changes to the quench's **cost model**, which is the piece every existing placement test pins. #431 is a rendering issue and its PR is already large; bolting a new cost term onto it would mean re-baselining the placement bit-identity suites inside a PR whose subject is PNGs.

It is also worth noting the two interact: aligning parts and orienting them are both "make the placement legible", and a shared rendering of *why* a part is where it is (which pad, which net, which direction) would serve both.

## Evidence this matters

The renders added in #431 make it visible for the first time — which is how it was noticed. A reviewer looking at a per-side panel of a quenched board sees parts at near-identical-but-not-identical offsets and slightly arbitrary orientations, and the caption metrics say the placement got *better*. Both are true, and that gap is the point of this issue.

### LINKED ISSUE #549: plan-pcb-routing: the skill produces the placement artifacts and reads none of them
The placement tooling now emits a lot of machine-readable evidence — `place_optimize`'s `JSON_SUMMARY`, `render_placement --json`, the `loop_round{N}.json` sidecars, the lock advisor's per-finding evidence records, `placement_state`'s signals, #409's blocker JSON, #504's ratsnest metrics, #538's blocks. `plan-pcb-routing` uses almost none of it.

The clearest symptom: **the skill never looks at a picture.** 1930 lines, 14 occurrences of "render", and every one of them is either *produce this* or *show it to the user*. There is no instruction anywhere for the model to `Read` a PNG it just made. Renders were added so a placement could be reviewed; the reviewer is a human who may not be in the loop.

The result is placements graded by vibes, and it shows on the boards.

## What is emitted and unused

| emitted today | used by the skill |
|---|---|
| `render_placement --json` — `panels[].path/.view/.side`, `moved`, `failed_nets`, `blocker_nets`, `metrics.*`, `unplaced`, `no_outline` | **the flag is not mentioned at all** |
| `--zoom-group`, `--focus`, `--focus-gap`, `--max-focus`, `--no-pads`, `--no-ghosts`, `--no-arrows`, `--no-delta-first`, `--ratsnest-all`, `--clearance` | none mentioned |
| `loop_round{N}.json` — `parent`, `accepted`, `screened`, `targets`, `groups`, `moved[]`, `metrics.*` | **no sidecar is ever read** |
| `place_optimize` — `oob_count`, `oob_amount`, `oob_area`, `parts_moved`, `blocks`, `block_parts`, `cost_*`, `airwire_length_*` | only `crossings_*`, `hpwl_*`, `overlap_area` |
| lock advisor — `findings[].evidence.*`, `advisories` (`high_pin_count`), `lock_patterns`, `lock_argv`, `locked_refs` | only `unlocked_high` |
| `placement_state` — `duplicate_fraction`, `spread_ratio`, `outside_fraction`, `stacked_refs`, `partially_unplaced` | only the exit code — **and these have no emitter at all**, so they are unreachable from a CLI today |
| `placement.groups.describe()` / `--list-groups` | mentioned for `route.py`, never as a placement decision input |

Three of these are not merely unused but actively misleadable:

- **`parent`** in the round sidecar is the last *accepted* board, not round N−1. Rendering a delta against N−1 draws a change that never happened.
- **`better()`** rejects a round whose `iterations` did not improve by 5% at equal failures. That rejection is the tool working, not evidence the placement is wrong — but nothing says so.
- **`no_outline: true`** means the out-of-board metrics are *unavailable*, not zero. Reported as "0 parts off board" it is a false clean.

## Three things that are structurally missing

**1. No floorplan intent.** Nothing is ever declared about where things are supposed to go, so nothing can check whether they went there. Every judgement collapses to "did `crossings` go down", which is indifferent between a sensible layout and a scattered one with the same wirelength.

**2. No decision procedure for blocks.** `--group-by` defaults to `none` on the placement CLIs and the skill's only guidance is a warning list. Which source to derive, whether to derive at all, and where — all undecided. The measured facts exist (`kicad` on 0 of 27 boards; `sheet` on 12 of 22 but moved nothing on any board tried; `decap` the only source that measurably moves anything, and 0% internal by construction) and none of them reach a decision.

**3. No independent verification.** A tool's own report satisfies its own gate. `place_optimize` says the placement improved and nothing re-measures that on the written file.

## What would close this

- An **artifact → datapoint → decision** map covering every key above: produce X, read key Y, do Z. Never a picture on its own.
- An **image doctrine**: when the model reads a render itself versus shows one, and what a render may never be used to conclude (renders are for intent; clearance, connectivity and DRC are for numbers).
- A **block decision procedure**, grounded in the measured corpus numbers rather than a caution.
- **Independent verification** — a tool's report re-measured on a different channel before it counts.
- A **feedback edge**: routing evidence classified as floorplan-shaped, placement-detail-shaped or parameter-shaped, naming the step to return to. This is discussion #407's *"knowing when to stop routing and go move something"*, and it is the piece that makes placement and routing one loop rather than two stages.

Routing-only must stay the default path throughout — most boards want a route, not a re-place.

Related: #431 built the renders, #548 covers the cost-model gaps they exposed, #459 the block-scale mover.

edgehero:
Closes #431
Closes #548
Closes #549

`#459` stays **open** deliberately — see [the split comment](https://github.com/drandyhaas/KiCadRoutingTools/issues/459#issuecomment-5140975464). Parts 3 and 4 are done, part 1 is now feasible, and part 2 (the bounded 80 mm relocation solve) is blocked on three invariants and on a harness that does not exist.

---

# The short version

The tooling emits everything needed to place a board well. Almost none of it was being used.

Three things follow, and this PR does all three:

1. **Block-scoped routing** (`#459` follow-on) and **headless placement renders** (`#431`) — the original scope.
2. **Floorplan intent, graded** (`#549`) — declare where parts belong, and a new `check_floorplan.py` measures the board against it and exits non-zero with the number that broke.
3. **Alignment and orientation in the cost kernel** (`#548`) — the two things a human does constantly that the quench did not.

Plus the skill rewrite that makes any of it matter: it now **reads the renders it produces**, pairs every one with the JSON key that confirms or contradicts it, and classifies a routing failure as floorplan-, placement- or parameter-shaped so the two halves form one loop.

**Routing-only remains the default and is now pinned by a test.**

---

# The gap this closes

`plan-pcb-routing` is 1930 lines and mentions "render" fourteen times. It never once told Claude to **look** at one. Renders were added so a placement could be reviewed; the reviewer was assumed to be a human who may not be in the loop.

Same for the numbers. `render_placement --json` existed and the skill did not know it. Neither did it know the `loop_round{N}.json` sidecars, `--zoom-group`, `--focus`, `oob_count`, `stacked_refs`, or `describe(blocks)`.

Three of those are not merely unused but actively misleading by their absence:

- **`parent`** in a round sidecar is the last *accepted* board, not round N−1. Rendering a delta against N−1 draws a change that never happened.
- **`better()`** rejects a round whose `iterations` did not improve by 5% at equal failures. That rejection is the tool working, not evidence the placement is wrong.
- **`no_outline: true`** means the out-of-board metrics are *unavailable*, not zero. Reported as "0 parts off board" it is a false clean.

`references/evidence-map.md` now pairs every emitted key with the decision it drives.

## The image doctrine

> Renders are for **intent**; numbers are for **legality**. No lens adjudicates clearance from pixels.

Four named situations where Claude `Read`s the PNG itself and says what it saw — before writing an intent, after an accepted placement change, when routing failed and the question is *do the failures share one pocket*, and when a block decision is live. Everything else is shown to the human. Budget: 3 images per turn, chosen by the question rather than by availability.

Two things a picture cannot show at all, so they are called out: `board_edge_contours` are **never drawn**, and a board whose outline failed to chain renders as a clean **rectangle**.

## The board outline is not ours to change

Researched, and worth stating plainly: **no such rule existed anywhere in this repo.** It was true only *by construction* — no writer emits an `Edge.Cuts` primitive — which means nothing stopped a future change, or a confident model, from resizing a board to make parts fit.

Now stated, and gated: never resize; if a board is genuinely too small, say so with the measured number and stop; never run the three corpus tools that *do* rewrite `Edge.Cuts` (`fix_outline_gaps.py`, `strip_routing.py`, `prep_set2.py`); and never gate on `oob_area`, because a part sitting **entirely inside a cutout** scores `0.0` there.

---

# `check_floorplan.py` — intent, graded (#549)

Placement here is judged by `crossings` and `hpwl`, and both are indifferent between a sensible layout and a scattered one with the same wirelength.

```bash
python3 check_floorplan.py board.kicad_pcb --emit-intent floorplan.json   # read it OFF the board
python3 check_floorplan.py board.kicad_pcb --intent floorplan.json --health
```

Nine rules, each measuring with the geometry **the optimizer itself gates on** — `QuenchState.legality_metrics`, `BoardOutlineGate`, `GradedPart`, `groups.decap_tethers`. A grader with its own idea of "legal" grades the reimplementation, so where re-deriving was plausible a test asserts the two agree exactly: all 34 of ulx3s's cap distances identical to the grouper's, all five legality numbers identical to the quench's.

**Exit codes are a contract**: `0` clean · `2` argparse or malformed intent · `3` board state · **`4` violations found**. `4` rather than `1` because `3` is taken and `1` is ambiguous with a crash — a caller has to be able to tell *"the floorplan is wrong"* from *"the grader is broken"*, and that distinction is the product.

### The round trip earned its keep by failing three ways

Emit an intent from a board, grade it back, expect zero errors. It did not:

- `legality_budget` written with `round()` landed up to **5e-5 below** the measured value — 50× `legality.EPS` — so watchy's emitted budget of 9.0972 was instantly violated by its own 9.09724. Rounds up now.
- Zone padding pushed zones outside the envelope. Clamped.
- **A zone per schematic sheet was simply wrong.** A sheet is a *functional* grouping, not a spatial one: its members scatter, so all ten of ulx3s's sheet bounding boxes mutually overlap, by up to 4508 mm² — an intent no placement could satisfy, and the emitter's fault rather than the board's. Zones are now emitted only where disjoint (ulx3s: 4 of 10), and the other six say why they have none.

That last one is the same spatial incoherence that makes sheet blocks useless for *movement*, arrived at independently.

### A board whose outline did not parse is refused, not graded

A broken outline degrades **silently**: unclosable segments dropped, no rings, `BoardOutlineGate.active` False, every containment test quietly falling back to the bounding box. A grader that inherits that reports a clean board because it stopped checking. `outline_state` reproduces the parser's own simple-rectangle short-circuit to tell the three "no rings" cases apart — a plain rectangular board **is** its bounding box exactly, and refusing it would refuse most of the corpus.

Found while building it and filed as **#550**: `extract_board_bounds` reads neither `gr_circle` nor `gr_curve`, so a round board reports `board_bounds: None` while its 64-point ring parses fine — and `add_board_edge_obstacles` returns immediately on a falsy bounds, so **a round board currently gets no board-edge keepout at all.**

### `--health`: what tells you the floorplan is wrong

Discussion #407's question. Block displacement, bus-corridor crossings, convergence — each computable or explicitly skipped with the reason. `blocked_cell_share` needs a routing attempt and says so.

**The measurement that made these real**, found by looking at the output rather than trusting it: GND owns 96 of ulx3s's parts and `+3V3` owns 45, out of 329 nets whose *median* is 2. Unfiltered, 8 of 10 blocks reported a foreign-pad count within 10% of the same median — every block was seeing the board's power nets, and displacement was measuring *distance from the middle of the board*. The same rails crossed **every** corridor. Filtered: median 332 → 40, the ranking changes, and the SDRAM corridor's worst offenders stop being `GND, +5V, +3V3` and become real signal nets.

---

# #548: alignment and orientation

**One correction to the issue's premise, because the fix depends on which way round it is.** #548 proposes to *"score airwires from the actual pad the net lands on"*. The cost path **already does that** — `_net_points` emits one MST node per connected pad, full rotation applied, and there is no centroid in the objective.

The gap is **numeric**. Measured: the four rotations of a part whose one connected pad faces away from its anchor differ by **0.500 mm across a 19.8 mm net**. Present, and drowned. What was missing is a term that is directional at *part* scale with its own weight.

| | |
|---|---|
| **alignment** | pairwise between peers, on the nearer shared axis: `w · min(d, radius)²`. **Continuous** at the radius (a cliff there pays a part to *flee* the row), **saturating** beyond it, **zero** on-axis |
| **orientation** | `Σ(\|r\| − r·û)` per pad, bounded at part scale on purpose — it breaks a rotation tie, it does not outrank a real length win |

Four caps seeded at y ∈ {99.8, 100.0, 100.2, 100.3} come out at **four distinct y** today and on **one shared y** with the term on.

### Both default to OFF, and that is the part worth arguing

The issue expected the bit-identity suites to be re-baselined. Looking at what they assert, that would be the wrong move: `test_458_quench_net_weights.py` **deliberately zeroes every geometry knob it knows about** so the objective is isolated enough to assert `total == 30.0`. Those are not goldens that drift — a term with a nonzero default is invisible to them and silently corrupts the isolation. Re-baselining would mean degrading a correctness test.

At `0.0` both hooks return before touching any geometry, so a default run is **bit-identical** — verified on two real boards, same poses and same metrics, not merely close.

### Measured, not recommended from taste

`--align-weight 5 --orient-weight 1`: best or joint-best crossings on both swept boards, tidier on both, **no legality regression anywhere**. With the caveats recorded next to the table — `--align-weight 15` is the tidiest row and its HPWL is *worse* than at 5, and alignment does nothing on a board with few repeated footprints because there are no peers to align.

### No GUI or converter work is needed — verified, so a reviewer need not go looking

Both placement CLIs are in `manifest_to_plan.REFUSED_TOOLS`, neither is in `test_cli_postpass_coverage.CLI_MAINS`, and there is no placement tab. So there is no dialog control, no `FLAG_PARAMS` entry and no `reset_params_to_defaults` line to add.

---

# Verification

**`VERDICT=`, not `RESULT=`.** `ai_backend.extract_result_line` takes the **last** `RESULT=` line and `ai_gui` parses it as the plan JSON, so copying the verdict spelling from the inspiration skill would have made every verified run report *"plan is not valid JSON"*. A gate blocks the collision.

Four new skill gates, **each verified to FAIL when its invariant is broken** rather than merely to pass: a `place` plan action added, the outline heading removed, the imperative removed, the `Edge.Cuts`-rewriting tools unnamed, verdicts spelled `RESULT=`. The outline gate needed the check itself fixed first — written as `or` over two phrases that were both present, it passed with either one deleted.

Routing-only is now pinned: placement stays out of `ai_plan.KNOWN_ACTIONS` and out of the plan template. Load-bearing rather than tidy — `ai_plan` **drops** an unknown action with a one-line note and runs the remaining steps anyway, so a placement step in a plan would silently route an unplaced board.

`references/*.md` and `docs/floorplan-intent.md` join the flag scanner's `SOURCES`, and `check_floorplan.py` its `TOOLS`: **73 flag citations, all real**, up from 57. Without that, every command block moved into a reference silently stops being checked.

---

# Also filed while researching

| | |
|---|---|
| **#550** | `extract_board_bounds` ignores `gr_circle`/`gr_curve` — a round board gets **no board-edge keepout**, and CLI/GUI disagree on its extent |
| **#551** | The placement CLIs advertise three things they do not do: a `--per-round` flag that does not exist, a `--metrics none` that is a no-op, and a dead `return 0` that would swallow any future exit code |
| **#552** | GUI skill runs cannot spawn subagents — `Task` is missing from `CLAUDE_ALLOWED_TOOLS`. The skill degrades to inline lenses and **says so in the report**, because a run must never look like a fan-out happened when it did not |

---

# Original scope (#459 follow-on, #431)

Follows on from #538 (#459 step 1), which landed the *concept* of a placement block but gave it no consumer outside the placement optimizer. This makes a block actionable on `route.py`:

```
--group BLOCK    scope a routing run to one block's nets
--group-by SRC   how blocks are derived (kicad, sheet, netprefix, decap; auto = kicad,sheet)
--list-groups    show what exists, with both net counts, before trusting any of it
--preview        route it, report exactly what WOULD change, write no board
--preview-png    render the proposed copper over the existing board
--undo           strip the scoped nets' copper back to unrouted
```

`--group` is the N-part generalization of `--component` and composes with net patterns the same way: patterns given → intersect, none → the block's nets.

## Two scopes, because the block decides which is right

Measured, not assumed. A schematic sheet is **60–70% internal** (glasgow's two 68-part sheets: 52 internal vs 23 boundary), so "route this sheet" is a real self-contained job. A decap block is **0% internal by construction** — a decoupling cap bridges VCC to GND and both span the board — so `touching` is the only useful reading there.

**The default depends on the operation, and that distinction is load-bearing.** Routing defaults to `touching`; routing a block's interface is the point. `--undo` defaults to `internal`, because a block's touching set contains GND/VCC and undoing *those* strips their copper across the whole board:

| `--undo` on rp2350's 16-part USB sheet | segments removed | outside the block |
|---|---|---|
| `--group-scope touching` | 170 | 54 |
| `--group-scope internal` (default) | 75 | 0 |

The wide scope still works and warns. `--undo` also refuses to run **unscoped** — for a route, "no scope" sensibly means the whole board; for an undo it means erasing every track on it.

## What `--undo` will not do, and says so

| | why |
|---|---|
| KiCad-**locked** copper | no override, segment path *or* arc path. #521 gives locked no override anywhere else and an undo is not the place to invent one. Nets protected for a *reason* (length-matched, diff-pair) **are** removed — naming a net exactly is the deliberate targeting those reasons exist to allow |
| **zone pours** | a filled plane is copper, so a net with a surviving pour is not back to "no copper". Deleting one is a plane decision |
| pad/target **swaps** | no inverse exists, and they can touch nets outside the scope |
| copper **graphics** | user artwork; counted and named, not deleted |

It **does** remove `(arc ...)` tracks, which needs its own pass. The parser *linearizes* arcs into `pcb_data.segments` so an undo counts them, but the writer only matches `(segment ...)` blocks — without it an arc-routed (i.e. hand-routed) net kept its copper while the run reported success: `8/24 removed`, silent. Matched with the parser's own field pattern, so a polygon's `(pts (arc ...))` entry, which carries no width/layer/net, cannot match.

It carries the sibling `.kicad_pro`/`.kicad_prl`/`.kicad_dru` (#441) — an undo only removes copper, so the input's DRC floor is still correct and must travel.

## The GUI/CLI gap this found

All three flags fell through `manifest_to_plan` into the generic `params` bucket, where `ai_plan` drops any param with no matching dialog control — and `nets` then defaults to `['*']`. So a recorded

```
route.py in.kicad_pcb out.kicad_pcb --undo --group decap:U3
```

converted to a plan step that **routes the whole board**: the inverse operation, at ~100× the scope, with nothing printed.

- `--group`/`--group-scope`/`--group-by` are now **hoisted** to the step (like `--component`) and resolved by `ai_plan._group_net_names` — CLI vs GUI verified equal under KiCad's python (`internal 14=14`, `touching 20=20`), and an unresolvable block is *noted* rather than silently widened.
- `--undo`/`--preview`/`--list-groups` are **refused** with a loud per-command stderr warning. The plan format cannot express them, and a wrong conversion is worse than a missing one.
- `--preview` exits **0** even when nets fail: `redo_commands.sh` runs under `set -e` and previews get recorded, so a read-only step must not be the only `route.py` mode that can kill a replay (a normal run with failed nets exits 0 too).

## Defects found by adversarial review, all with reproductions

Beyond the converter gap: arcs silently surviving; the `touching`-default board-wide strip; `output_file[:-10]` producing an empty base for a non-`.kicad_pcb` output and dropping a stray `.kicad_pro` **dotfile in the CWD** while announcing it (now `os.path.splitext`, matching `copy_board` and `pro_path_for_board`); `--list-groups` replaying as a full-board route; zones contradicting the "no copper" claim; and a fuzzy `--group U3` resolving to `decap:U3` and deleting copper without naming what it matched (now echoed, and accepted only when exactly one block matches).

Two categories were checked and found **clean**: cross-net via removal survived an adversarial board carrying a *different* net's via at the identical coordinate plus one 1.5 µm away (inside the 2 µm tolerance) — both intruders survived, only the target died; and `--preview` genuinely writes nothing (input byte-identical by md5, no stray files, preview counts matching a real route exactly).

## Second commit: a one-line GUI bug that was silently un-resetting every plan step

While auditing this work against `test_gui_engine_parity` I chased down a copper divergence I had initially assumed was someone else's regression. It was not, and the real cause is worth its own commit:

`reset_params_to_defaults` passed `50.0` to `impedance_value`, which is a `wx.SpinCtrl` — **integer-only**, so wxPython 4.2 raises `TypeError`. `ai_plan`'s per-step reset **catches** that and logs `"per-step reset skipped"`, so the reset abandoned itself at that line and every control below kept the previous step's value.

The damage was not the impedance field. ~80 lines further down sits the #439 loop that un-checks every geometry-floor override box, and it never ran. With those boxes left checked, `_effective_geometry_floor` returns the typed control value instead of the board's Default net-class — so a plan step that **omits** `track_width`, meaning "use the board's value" exactly as an omitted CLI flag does, got `routing_defaults.TRACK_WIDTH` instead.

| | segments | vias | copper sets |
|---|---|---|---|
| before | CLI 1307 / GUI **1308** | 144 = 144 | **DIFFER** — 58 GND plane taps at identical coordinates, CLI 0.127 vs GUI 0.3 (2.4× the copper) |
| after | CLI 1307 / GUI **1307** | 144 = 144 | **IDENTICAL** |

The gate printed `VERDICT: PARITY` the whole time, because DRC and kicad-unconnected never moved — only the copper-set comparison saw it.

Three things worth saying plainly:

- **The design and the #439 loop were already correct.** Nothing needed adding. My first attempt added a *second* un-check loop and I deleted it once I found the real cause — duplicating a correct implementation is how the two drift apart later.
- **It is not a regression in the recent span.** `reset_params_to_defaults` and the hardcoded `50.0` both arrived in `e8c72fe`, an ancestor of the `ef4c19a` parity marker. My earlier note in `.gui-parity-checked` blaming `d9bccbd` ("power TRACES get ampacity") was **wrong**, and is corrected in place rather than left to send someone bisecting. It is wxPython-build dependent — older builds coerced the float silently, wxPython 4.2.2 / wx 3.3.2 raises — which is why the marker could honestly record IDENTICAL while the same code diverges here.
- **The new gate asserts observable state, not that the reset ran.** A caught-and-logged exception in a reset is otherwise indistinguishable from a reset that worked, which is exactly how this hid.

## Verification

- **`tests/test_459_group_routing.py`** — 24 cases. Uses **real `(locked yes)` copper** rather than a patched `protection_map` (patching only proves the string is honored), and asserts **vias** as well as segments in the no-collateral case — vias are removed by *position*, the one path where cross-net damage was plausible. Arcs get a synthetic fixture, since no corpus board has one, which is exactly why that bug survived.
- **Three new gates, each verified to FAIL when its fix is reverted** rather than merely to pass: `check_group_flags` in `test_manifest_plan_parity.py` (no wx), `tests/gui_parity/test_group_scope_parity.py`, and `tests/gui_parity/test_geometry_floor_leak.py` (both re-exec into KiCad's python).
- **Block routing touches only the block**: +140 segments on rp2350's USB sheet, zero changes to any other net; DRC clean at the routed 0.1 mm floor before and after.
- `test_gui_engine_parity`: **copper sets IDENTICAL**, 1307 = 1307, vias 144 = 144.
- `run_all.py --fast` **186 passed / 0 failed**; `pytest --collect-only` **327**.

`placement.groups._short` is promoted to public `short_name` — it is what the user *sees* in `--list-groups` and types back into `--group`, so both sides need the same rule.

## On the parity marker

`.gui-parity-checked` gains a flag audit of the `ef4c19a..HEAD` span: `--stitch-*` are in the converter tables; `--thermal-relief`/`--thermal-vias` reach real planes-tab checkboxes through the unknown-flag fallthrough (verified by conversion); the `place_optimize`/`place_route_loop` flags are N/A (no GUI tab, and neither script is in the gate's `CLI_MAINS`).

The recorded SHA is deliberately **still not bumped**. That audited new *flags* across the span — not every engine parameter, results-data key and post-pass, which is what a full audit means here. Bumping it would assert more than was checked.

Worth knowing if you run these gates on Windows: KiCad's bundled python had no `scipy` (it is in `requirements.txt`), so the GUI leg died in `run_all_checks` and the gate reported `GUI=0 segments` — which reads like "the GUI routed nothing" rather than "the GUI never started".


edgehero:
@drandyhaas do not merge yet, found a bug 😔😅

edgehero:
## What the artifacts actually look like

Everything below is produced by the tools in this PR, on tracked boards, with the command that made it. All six are committed under `docs/` (580 KB total, quantized — in line with the 666 KB fanout GIF the repo already carries) and are shown in `placement/README.md`.

### The movie

`make_movie.py WORKDIR --camera auto` over a real `place_route_loop` work dir. **13 components, 4 of them on the back, with routing** — 4 rounds, failures 4, iterations 84k → 66k.

Establishing overview → routing appears → front parts move → **the board turns over** → pan → back caps move → routed again → back to overview.

placement movie

The flip is a real 180° turn about the vertical axis, not a dissolve: the board narrows to an edge as it passes 90°, then opens out already reversed, and **every frame after it stays mirrored** — because you are looking at the other face. Reading a back-side placement off an un-mirrored X-ray means mentally reversing every x coordinate, which is exactly the error a render should remove.

One rule makes the sequencing legible: **a frame either moves the camera or changes the board, never both.** Every transition happens over a frozen board and is followed by a settle beat, so the moves only play once the camera has arrived. `--tween 0` (used here) cuts to each new placement instead of gliding.

### The stills

**The placement diff, drawn.** `--before` turns on dashed seed rects and displacement arrows. This is interf_u's tracked 22-part delta; `C4` travelled 95.65 mm.

```bash
python3 render_placement.py placed.kicad_pcb --before seed.kicad_pcb -o delta.png
```

placement delta

**Per-side panels.** Placement lives on two sides, so `--per-side` emits one panel per side rather than one flattened projection. The near side is bright, the far side's SMD pads are dimmed (still context — a back-side part is why a front trace has to detour), and **through-hole pads stay full brightness on both**, because they are physically on both and a hole cannot move on one side only.

```bash
python3 render_placement.py board.kicad_pcb --before seed.kicad_pcb --per-side -o out/
```

| front | back |
|---|---|
| front | back |

**Zoom to one placement block.** `--zoom-group` takes the same block names as `route.py --group`, resolved through the same code, short-uuid forms and ambiguity refusal included.

```bash
python3 render_placement.py board.kicad_pcb --list-groups --group-by sheet
python3 render_placement.py board.kicad_pcb --zoom-group 58d913ec --group-by sheet -o blk.png
```

group zoom

**An unplaced board.** The placement CLIs refuse this one (exit 3, before round 0 pays for a route). The renderer warns and draws it anyway — being able to *see* the state is the point — framed to the PARTS rather than the outline, since a pile at the origin would otherwise render as a dot in the corner of an empty board. 13 footprints at 1 distinct position, `overlap 792 mm²`.

unplaced

### Toggles

`--no-borders` · `--no-labels` · `--no-ratsnest` · `--no-arrows` · `--no-ghosts` · `--no-pads` · `--per-side` · `--focus` (one cropped panel per failed-net cluster) · `--zoom-group`.

`--ratsnest-all` is the deliberate hairball switch. By default only the moved and attributed nets are drawn, because a full ratsnest on a dense board reproduces exactly the unreadable mess KiCad already shows — which is limitation #1 in the issue text, and the reason delta-first is a default rather than polish.

### One thing the captions are doing

Every panel carries its metrics — crossings, HPWL, overlap, out-of-board, and for a loop round the failures and iterations. That is not decoration. Per the issue's own limitation #3, geometry is not causality: arrows show *what* moved, never why it helped, and a frame with big displacements and a frame with tiny ones can be equally good or bad. **The render is triage; the caption is the verdict.** A reader who judges a placement by how much moved has adopted precisely the wrong heuristic, so the docs say so next to the pictures.

edgehero:
## Every toggle, and a labels example

### Reference designators

Asking for a labels example turned up a real defect. Labels *were* being drawn — 653 pixels of them — but the font scaled with **image height alone** (`int(H / 90)`, clamped to 8 px), so a 400 px render got an 8 px label at any zoom. Technically present, completely unreadable.

They are now sized from the **part's own footprint on screen**, clamped to a legible range, with a dark stroke so a reference sitting on gold copper still reads. A part too small to letter is skipped and its arrow carries the meaning, rather than stamping a smudge nobody can resolve.

labels

`J1`, `J2`, `IC2`, `IC3`, `IC4`, `R1`, `R3`, `R4`, `C2`–`C5` — the blue ones are the four back-side caps.

### The toggles

Same 13-part delta, one flag at a time. Contact sheet rather than a dozen files: easier to compare against, and cheaper to carry in the repo.

toggles

| flag | what it drops |
|---|---|
| *(default)* | courtyards, labels, airwires, pads, seed ghosts, arrows |
| `--no-borders` | component courtyards |
| `--no-labels` | reference designators |
| `--no-ratsnest` | airwires |
| `--no-pads` | pad copper — outlines and topology only |
| `--no-ghosts` | the dashed seed rects, keeping the arrows |
| `--no-arrows` | the displacement arrows, keeping the seed rects |
| `--no-delta-first` | the dimming of parts that did not move |

**`--no-delta-first` renders identically in that sheet, and the caption says so with the reason**: 12 of the 13 parts moved, so there is no context left to dim. Shipping a panel that silently looks like a no-op would teach the reader the flag does nothing — so it is demonstrated on a board where the delta is a genuine subset instead, alongside the switch that matters most there:

ratsnest

ulx3s, 110 of 234 parts moved. By default only the moved and attributed nets are drawn; `--ratsnest-all` is the deliberate hairball switch. On a dense board it reproduces exactly the unreadable mess KiCad already shows — which is limitation #1 in the issue text, and the reason delta-first is a default rather than polish.

### The rest

`--per-side`, `--zoom-group`, `--focus` and the unplaced case are in [the earlier comment](https://github.com/drandyhaas/KiCadRoutingTools/pull/542#issuecomment-5139157582); all of it is also in `placement/README.md` so it stays with the code.

Everything above is quantized to a 128-colour palette — flat-colour renders, so it costs nothing visible. Whole docs set 860 KB.

`run_all.py --fast` 192 passed / 0 failed.

Issue #615, opened 2026-08-12
A fresh clone of tip is unrunnable: prebuilt grid_router says 0.20.1, Cargo.toml says 0.20.0 -- and the version string hides a stale ABI


edgehero:
`python build_router.py` on a clean checkout of `placement` tip (`0f45ba2b`) downloads `grid_router-windows-x86_64.pyd` **v0.20.1** from release v0.20.2 — and then every CLI refuses to start, because `py_router/startup_checks.py` requires `installed_version == cargo_version` and `rust_router/Cargo.toml` still says **0.20.0**. The tool that installs the binary and the check that rejects it disagree, and the printed remedy ("run `python build_router.py`") is the command that was just run.

Reproduced on Windows while dry-running patches in a fresh worktree of the tip; the same mismatch will hit any platform that takes the prebuilt path. Workaround we're using meanwhile: bump the local `Cargo.toml` to 0.20.1 (uncommitted).

**It gets worse: the version string cannot be trusted either.** The prebuilt `grid_router-windows-x86_64.pyd` from release v0.20.2 reports `__version__ 0.20.1` but carries an OLDER ABI: its `GridObstacleMap.add_stub_proximity_costs_batch` signature is `($self, stubs, radius, max_cost)` — missing the `block_vias` parameter that `rust_router/src` at tip defines and that `py_router/obstacle_costs.py:84` / `routing_context.py:108` pass. With that binary installed, every real route/route_diff/plane run dies with `TypeError: ... takes 3 positional arguments but 4 were given` (measured: 17 of 56 test failures in one sweep traced to exactly this). The strict version-equality gate passes it happily, because the stale binary and Cargo.toml agree on the string.

Fixes, maintainer's choice (the ABI half suggests the first is not enough alone):
- pin the release-asset download to the version `Cargo.toml` names AND rebuild/re-upload the release asset from current source, or
- let the gate accept a compatible newer patch release but ALSO probe one ABI-sensitive signature (e.g. `block_vias` in `add_stub_proximity_costs_batch.__text_signature__`) so a stale-but-same-version binary is caught at startup.

drandyhaas:
Thanks for the detailed report — the repro was spot-on, but the ABI diagnosis is actually backwards, and that changes the fix.

[...]

Issue #579, opened 2026-08-06
GUI/CLI copper sets diverged after the run-8 merge (CLI=1351 GUI=1335, was identical)


drandyhaas:
`tests/gui_parity/test_gui_engine_parity.py` reported **copper sets IDENTICAL** (segments 1335/1335, vias 140/140) before merging PR #576, and reports them DIFFERENT immediately after (`bec03cf`):

```
segments: CLI=1351 GUI=1335 common=1316 cli-only=35 gui-only=19
vias:     CLI=140  GUI=140  common=137  cli-only=3  gui-only=3
copper sets identical: False
```

Stable across three consecutive runs — not flaky. The verdict line is still `PARITY` (kicad unconnected 0/0, drc 0/0) and the gate exits 0, because it **reports** divergence rather than failing, and CLI-side post-passes are a known-divergent area. So this is not a stop-ship. But the sets were identical before this merge and that property is worth keeping.

The CLI gained 16 segments; the GUI is unchanged at exactly its previous 1335. That shape — one front moves, the other doesn't — is the Class-1/Class-2 divergence CLAUDE.md describes.

## Ruled out

- **`check_connected._pad_credit_disc`** (the PR's custom-pad credit fix). The gate board `kicad_files/splitflap_driver.kicad_pcb` has **zero** custom pads (297 pads, 0 custom), so the polygon branch never runs, and the non-custom fallback is behaviourally identical to the code it replaced:
  - old: `if pad.size_x and pad.size_y: reach_pad = min(size_x, size_y)/4 else: 0.05`
  - new: `return pad.global_x, pad.global_y, min(pad.size_x, pad.size_y)/4.0` … `return …, 0.05`
- **`fix_kicad_drc_settings`** (+93 lines, the fab-floor ratchet). Disclosure only — `_fab_floor_disclosure` prints; no written value changes. Confirmed by diffing out comments and print statements.

## Remaining suspects

The merge changed six engine files. After the two above, what is left:

| file | added | note |
|---|---|---|
| `py_router/repair_planes.py` | +51 | rip-casualty disclosure + `main()` returns 4 |
| `py_router/route_planes.py` | +32 | — |
| `py_router/check_drc.py` | +58 | grading only, does not route |
| `py_router/cli_banner.py` | +13 | printing only |

[...]

Issue #522, opened 2026-07-30
Organize root level python files into a few directories


drandyhaas:
Maybe py_router, py_tools, ?

drandyhaas:
Risk assessment (2026-08-04), measured at HEAD:

**Blast radius**: 122 root .py files; **1,083 flat cross-imports across 91 of them**; **454 recorded manifests bake absolute paths to root entry scripts**; 21 modules use `__file__`-relative paths (rust_router sys.path insert, kicad_files fixtures); the PCM-packaged GUI plugin imports engine modules from repo root.

**Danger classes, each with prior scar tissue in this repo:**
1. Manifest replay death: moving `route.py`/`route_planes.py`/fanout scripts breaks all 454 replay chains and the A/B infra (the repair_planes rename did this deliberately for ONE dead step).
2. Lazy-import style: hundreds of function-local imports mean a missed rewrite fails mid-run inside try/except, not at startup.
3. The #506/#507 class: an import-layout change once made the GUI silently abort the apply while reporting success — package-vs-flat resolution differs between dev python and KiCad's bundled python, and is only testable under the latter.
4. `__file__`-relative paths silently change meaning one directory down.
5. The taught surface (skills/RUNBOOK/docs/manifest_to_plan basename parsing) keys on root script names.

**Safe shape:**
- CLI entry points STAY at root permanently (route.py, route_diff.py, route_planes.py, repair_planes.py, bga_fanout.py, qfn_fanout.py, place_fanout_clearance.py, check_*.py, copy_board.py, build_router.py) — bga_fanout.py already models the thin-wrapper-over-package pattern.
- Internals move in small reviewed clusters (plane/*, writers/, engine/...), each followed by the full gate suite: manifest parity, postpass coverage, doc examples, wx engine-parity UNDER KICAD PYTHON (danger 3 is invisible without it), and a board e2e.
- No big bang. Roughly a session per 2-3 clusters done honestly.

Verdict: doable because the gate suite is strong, dangerous if done as a bulk `git mv`.

drandyhaas:
Landed on main (89914ce + marker 3a8579c, local — not pushed). Final layout:

- **`py_router/`** — the 104 engine modules + the `bga_fanout`/`qfn_fanout`/`placement` packages. Scripts stay directly runnable (`python3 py_router/route.py ...`); flat imports are untouched because the script's own directory is `sys.path[0]`.
[...]

PR #616, opened 2026-08-12
routing: boards graded at constants they contradict, and tools that reported success falsely

edgehero:
Two themes, one constraint honored throughout: nearly everything here is additive or a demonstrably-wrong rewrite of at most a dozen lines per site, each with a test that fails on the unpatched tree. Net footprint: ~212 existing lines modified (56 of those are the four duplicated floor-read blocks one shared resolver deletes) against ~2,400 added, most of it tests and one new module.

## Theme A — the board declares a floor, and the code used a constant

- **One resolver for CLI floors** (`list_nets.resolve_cli_floor` + the declared-0 guard in `board_floor_knobs`), wired through all four routing mains. KiCad writes 0 for "not configured"; the mains' private `is not None` reads took it as a real floor. Measured: a declared `min_copper_edge_clearance: 0.0` printed "using the board … 0.0mm" and dropped the plane inset 0.5 → 0.0, while the GUI plane tab already guarded `> 1e-9`. Pinned by `tests/test_board_floors.py`, which also runs each main live on a declared-0.0 board. (This file re-creates the surviving checks from the test removed with #579, minus the removed feature's blocks, under a new name.)
- **`min_hole_clearance` reaches the router and its checker** (`obstacle_map.resolve_hole_clearance`, `routing_config.hole_clearance`, `plane_obstacle_builder`, `check_drc --hole-clearance` with project auto-read, `FAB_FLOOR_KEYS`). Measured: a kicad-cli-confirmed 0.0237 mm routing-introduced violation, and neo6502 shipping 3 NPTH at 0.2126/0.2263 mm against an authored 0.25, graded clean because nothing consulted the key. Pinned by `tests/test_npth_seed_floor.py`.
- **BGA underpad escape reads the board's `min_hole_to_hole`** (fab-wrapped, tighten-only). Measured on ulx3s (declares 0.4): drill gap 0.3657 → 0.6000; control board byte-identical. 13 checks in `tests/test_bga_fanout_underpad.py`.
- **QFN underpad: candidates tested against this run's own output** (`run_output_conflict`, cross-side), and a reused via priced as a reuse (`adds_via=False`). Pinned regression on `kicad_files/routed_output.kicad_pcb` U2: 29/13/31 → 28/12/30. The original headline symptom (39/46 rejected escapes) did not reproduce and the body of the commit says so — the structural gap plus the regression pin are the evidence.
- **GUI/CLI floor divergence closed**: `_effective_geometry_floor` gains the declared-0 guard its sibling plane-edge helper already had (GUI drilled h2h 0.1 vs CLI 0.2, same board, same settings). Pinned under real KiCad python by `tests/gui_parity/test_geometry_floor_leak.py`; `.gui-parity-checked` gains a scoped addendum with the parity evidence (copper sets identical, 1307 segments / 144 vias, CLI = GUI).

## Theme B — false success

- `route.py`: a `--nets` list where every name missed wrote a passthrough copy and exited 0 (run 11 lost two laps to boards KiCad refuses to load while every in-repo checker read them fine); CRLF names stripped; summary gains `scope` + `ripped_open_uncounted` so the last `JSON_SUMMARY` of a log can't masquerade as the whole run.
- `route_diff.py`: zero matched pairs exited 0; a `partial` pair with both members incomplete counted `successful` (fixed engine-side — the GUI reads `successful` raw).
- `route_planes.py`: `"GND:B.Cu"` was accepted as a layer name and KiCad refused the written file; the pad tally reaches the JSON summary (recording `geometric_failed` without flipping status on it — under pours-first those are expected at the plane step).
- `check_connected.py`: a phantom zone layer joined the copper census (69/70 graded 70/70); a scope that matches nothing exits 2.
- `check_weird.py`: new `dangling-via` category — KiCad flagged 64 on a board where this tool flagged 0.
- `diff_pair_custody.py`: a pair the audit demoted to partial kept its credit.
- `kicad_oracle.py`: the timeout memo keyed on a realpath every chain step renames (240 s re-burned per step) — now keyed on the fill-cost signature; `KICAD_CLI` env override first; broader Windows discovery.
- `route_summary.py`: `oracle_check` merges worst-news-first — a real KiCad contradiction was erased by a later subset-pass `skipped`.
- `plane_fragility.py`: the one call site that dereferenced `refill_islands`' documented None return.
- `plane_region_connector.py`: the open-space fallback seeded strap ends on bare board 3–5 mm from the region (clearance saturates off the plane net's own copper; the strict first-maximum tie-break picks the far corner in raw grid order). Seeding on region material takes neo6502's GND from 15 of 41 region joins dead-by-construction to 0; two healthy boards byte-identical; `validity=None` keeps the old path verbatim. Pinned by `tests/test_region_join_nearest_pair.py`.

## Riders both themes stand on

- `py_router/krt_deadline.py` (new): cooperative `--deadline` in the four mains — arm/stamp/seal, exit 7, a guaranteed `JSON_SUMMARY` on the way out. The wiring is one closure into the `cancel_check`/`progress_callback` parameters the engine functions already have; `place_route_loop` treats exit 7 as a partial result with a summary, not a failed step. (An external kill leaves no summary and no exit code of the tool's own — that is the gap this closes from the inside.)
- `cli_banner.install_dash_hint`: `--nets -12V` is an argparse death, not a routing result; adopted only in the four mains (prefixed `JSON_SUMMARY:` lines, no bare-JSON stdout mode touched).
- `tests/run_all.py`/`run_utils.py`: timeouts reported as `TIME`, separate from `FAIL`, with per-child utf8.

## Testing

The board-floor family ran green inside a patched worktree of `0f45ba2b` before this branch existed (43/43 board-floor checks including live subprocess runs of all four mains, 13/13 BGA, 26/26 QFN, the GUI leak test under KiCad 10.0 python), and again on this branch. Full `tests/run_all.py` on the source branch: 341 passed, 6 failed — all six pre-existing and unrelated (the two #549 floorplan grades, multipoint_diff, tigard_usb 9/12, the opt-in placement probe, the skills-generic lint), 2 known-slow timeouts.

Two deliberate behavior deltas vs tip, so they don't surprise a reviewer's test run: `test_multipoint_diff_route` (7/7 → 1/7 scenarios couple-routed) and `test_tigard_usb_diff` (12/12 → 9/12) — the router now enforcing `min_hole_clearance` and the stricter pair-success accounting make `route_diff` defer pairs it previously couple-routed; the boards still finish connected and DRC-clean through the single-ended follow-up. These land at exactly the levels our source branch has carried since the fixes went in.

Note on running any of this from a fresh clone: the tip currently refuses to start after `build_router.py` (prebuilt 0.20.1 vs Cargo 0.20.0 strict equality) — filed as #615 with details; we test with a local uncommitted Cargo bump.

### DIFF
--- a/.gui-parity-checked
+++ b/.gui-parity-checked
@@ -627,3 +627,19 @@ HEAD 2026-07-30 parity audit 191abc4..d73b1d6 (3 commits). Outcome: no gaps.
     gate runs a fanout step; ea992d9), its stale _path bootstrap
     (3c4fc48), and the #562 rename breaking every recorded manifest's
     repair step (relocate_moved_scripts rename table, 89914ce).
+
+ADDENDUM 2026-08-12 (PR: board floors + false success) -- scope: the one
+GUI-relevant change in this PR is kicad_routing_plugin/swig_gui.py
+_effective_geometry_floor gaining the declared-0 guard (a board writing 0 for
+"not configured" was taken as a real hole-to-hole floor by the GUI while the
+CLI's resolver treated it as unset: measured, GUI drilled h2h 0.1 vs CLI 0.2,
+same board, same settings, under fab-overrides). The plane-edge helper beside
+it already guarded > 1e-9; this closes the divergence in the other helper.
+  * Pinned by tests/gui_parity/test_geometry_floor_leak.py (runs under real
+    KiCad python; "declared-0.0 hole_to_hole, fab floor 0.1 -> GUI 0.2
+    (CLI 0.2) ... 0 problem(s)").
+  * Parity evidence on the source branch (KiCad 10.0, pcbnew 10.0.0,
+    wx 4.2.2): test_gui_engine_parity VERDICT PARITY, copper sets IDENTICAL
+    (segments 1307=1307, vias 144=144, 0 one-sided); test_manifest_plan_parity
+    OK; test_cli_postpass_coverage 7 registered / 14 in use / 1 acknowledged
+    CLI-only (run_drc fanout graze audit, report-only) / 0 unreachable.
--- a/kicad_routing_plugin/swig_gui.py
+++ b/kicad_routing_plugin/swig_gui.py
@@ -754,6 +754,23 @@ class RoutingDialog(wx.Dialog):
             else:
                 netclass = _get_netclass_parameters('Default') or {}
                 board_val = netclass.get(name)
+            # A declared 0 is UNSET, not a floor of zero -- KiCad writes 0 into
+            # these fields for "not configured". Every other resolver in the
+            # tree applies that rule (list_nets.board_floor / board_floor_knobs,
+            # resolve_cli_floor, and _effective_plane_edge_clearance just above,
+            # which already guarded `> 1e-9`); this branch tested only
+            # `is not None` and so diverged from the CLI.
+            #
+            # Masked in the default tier, because _fab_floored then pins a 0.0
+            # up to the 0.2 fab hole-to-hole floor and the CLI lands on 0.2
+            # too. NOT masked once the fab floor moves: with a --fab-overrides
+            # declaring hole_to_hole 0.10 (fab_floor_ladder collapses to that
+            # one hard rung, and the GUI reaches it through the fab_tier /
+            # fab_overrides_path controls), a board declaring 0.0 resolved to
+            # GUI 0.1 against CLI 0.2 -- the GUI drilling twice as close as the
+            # CLI on the same board and the same settings.
+            if board_val is not None and board_val <= 0:
+                board_val = None
             val = board_val if board_val is not None else getattr(self, name).GetValue()
         return self._fab_floored(name, val)
 
--- a/krt_capabilities.py
+++ b/krt_capabilities.py
@@ -27,10 +27,33 @@ import sys
 
 ROOT = os.path.dirname(os.path.abspath(__file__))
 
+# The #522/placement-split layout spread the tools over py_router/, py_tools/
+# and py_placer/. A consumer keeps naming them bare ('route.py'), so every
+# lookup resolves through _tool_path, which accepts EITHER layout -- flat for
+# an older clone, packaged for this one.
+_TOOL_DIRS = ('', 'py_router', 'py_tools', 'py_placer')
+
+
+def _tool_path(root, mod):
+    """Absolute path to `mod`, wherever the #522/placement-split layout put it.
+
+    Falls back to the flat join so the error message a missing module produces
+    still names the place the caller expected it.
+    """
+    for d in _TOOL_DIRS:
+        p = os.path.join(root, d, mod)
+        if os.path.isfile(p):
+            return p
+    return os.path.join(root, mod)
+
+
 # The modules a consumer is likely to depend on by name. Presence is reported
 # for every one; absence is only an ERROR when --require asks for it.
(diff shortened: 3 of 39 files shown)


drandyhaas:
Thanks — there's a lot of genuinely good work in here. I've cherry-picked **17 of the 18 commits onto `placement`** (applied clean, author attribution preserved): the floor resolver + declared-0 guard, `min_hole_clearance` reaching the router and `check_drc`, the BGA and QFN underpad fixes, the GUI declared-0 h2h guard, the test-runner TIME/utf8 changes, the dash-net argparse hint, the `krt_deadline` machinery + the mains' false-success guards, the `check_connected` phantom-zone-layer + empty-scope fixes, the `check_weird` dangling-via category, the `kicad_oracle` memo/discovery fixes, the `route_summary` sticky-merge, the `plane_fragility` None-deref fix, the plane joiner region-material seeding, the route_render caption wrap, the net_rescue width histogram, and the `krt_capabilities` #522 fix.

Verified here before pushing: board-floors 43/43 (incl. the live main runs), BGA 13/13, QFN, region-join, npth-seed, the GUI floor-leak test under real KiCad python (0 problems), `--deadline` end-to-end (exit 7 + `complete:false` when armed, byte-inert when not), and `test_tigard_usb_diff` / `test_multipoint_diff_route` at 12/12 and 7/7.

**The one commit I did not take: `diff_pair_custody: a pair the audit demoted to partial kept its credit`.** Its "both members incomplete ⇒ demote" discriminator misfires on multipoint pairs whose extra pads are *deliberately* deferred to single-ended follow-up. Your own run-11 example is the counterexample: on tigard, `/USB_D`'s coupled trunk routes and grades DRC-clean, and both `/USB_DP`/`/USB_DN` carry redundant J1-row pads that the designed flow connects in the follow-up step — so `partial` + both-members-incomplete is the *expected* step-level state there, not a false success. With the commit in, `tests/test_tigard_usb_diff.py` drops 12/12 → 9/12 (pair demoted from `routed_diff_pairs` while the copper is unchanged); without it, 12/12. At the chain level the incomplete members are already counted as pad deficit (the accounting doctrine in CLAUDE.md), so the credit isn't hiding anything — and note your `1152917c` `successful -= len(partial_diff_pairs)` fix, which I *did* take, already handles the narrower engine-side case correctly. If there's a case where a pair kept credit with **no coupled trunk copper at all**, that's worth demoting — but the discriminator needs to be "no coupled copper kept", not member count. Happy to take a revised version.

**One follow-up commit of mine on top** (`dash-hint test: probe argparse instead of assuming it rejects -12V`): Python 3.14's argparse accepts a dash-prefixed value in an `nargs='+'` list, so on those interpreters the `--nets -12V` failure your hint annotates cannot be provoked and three of the six checks asserted a `SystemExit` that never comes. The test now probes the behavior at import and skips those three with a reason; the hint itself stays installed (inert there, active on the older Pythons where you hit it).

You'll want to rebase the branch on current `placement` — the 17 are on it now, plus an unrelated `build_router` change (#615's branch guard, which also removes the need for the local Cargo bump your testing note mentions).

drandyhaas:
Closing: 17 of the 18 commits are merged to `placement` (7d5d142..e3bf258, authorship preserved) — see the comment above for the full accounting, the one commit not taken (the custody demotion, which needs a "no coupled trunk copper kept" discriminator rather than member count), and the test-runner port. Thanks for the thorough work — future fixes at this quality bar are very welcome. If you revise the custody discriminator, open a fresh PR against current `placement`.

Issue #514, opened 2026-07-29
Stress wave sets 21-25: findings for existing issues (roll-up)


drandyhaas:
Companion to #513. Same source: triage of the **318 findings logged by the LLM stress agents across sets 21–25** (75 boards, `sonnet`, v0.19.1 / `9671c55`, 2026-07-28/29; 97.4% net completion, 59/75 fully connected, 88 residual DRC).

#513 collects the findings that need **new** issues. This one collects the findings that belong on **existing** issues — kept as a single list here rather than scattered across nine threads. Each section names the issue it should land on.

---

## → #505 (Corpus sets router-introduced DRC)

Wave DRC totals: **88 residual across 75 boards**, but most nonzero boards *improved* on their baseline. Only three regressed:

| Board | delta | class |
|---|---|---|
| `healthypi_sensor` | **+12** | `VIA-SIZE` |
| `pocat_comms` | **+6** | `pad-via` |
| `ice4pi` | **+1** | `SEGMENT-BOARD-EDGE` |

Both large regressions are now root-caused and belong in their own issues rather than here — see #513 items 1 and 2 (fanout tools discarding `enforce_fab_floors`' return value; the plane-tap path never reading `pad.local_clearance`). Together they are 18 of the 88.

Notable improvements, for the record: `esp32s3_m2_som` −103, `peaksat_comms` −87, `apple2e_mmu` −68, `astrouniboard` −45, `spartan6_4layer` −659. `miniscope_v4_flex` carries the single largest residual (53 `SEGMENT-BOARD-EDGE`) but at delta −23 — a flex board graded at 0.1016 mm that inherited most of them.

Two smaller router-introduced items worth tracking here:

- `usmu_smu` — 2 `SEGMENT-BOARD-EDGE` (0.1 mm graze vs a 0.2 mm floor) on a GND plane-repair trace hugging the bottom-left board corner, introduced by `route_disconnected_planes`' region-join and confirmed real by kicad-cli (kicad=2, matched=2). The region-joiner should respect board-edge clearance when joining near a corner.
- `peaksat_obc_adcs` — 1 persistent `PAD-SEGMENT` graze (0.044 mm) from a `qfn_fanout` stub escape that survives the whole chain. Thinner stubs didn't clear it, and `place_fanout_clearance.py` is a no-op because stub escape places no vias to nudge parts off of. Would need a stub-escape-aware clearance fixer.

---

## → #479 (Remaining net disconnections on the stress corpus)

[...]

PR #623, opened 2026-08-12
placement: the reconstruct ladder, reseat, orbits, and instruments that stopped lying

edgehero:
**Stacked on #616** — this branch includes #616's commits; the placement work is the 9 commits from "reconstruct: the ladder trusted any equal-cost move…" onward. After #616 merges this rebases to show only its own diff.

The placement engine work from perturbed-corpus runs 8–17, re-rooted onto the #522 layout (`py_placer/`, `py_tools/`), plus the stress rig that grades it.

## Engine

- **Reconstruct ladder**: the hole-pattern fit learns per-group families, mid-edge slots and positional at-seat; an orbit detector corroborates seats; an unevidenced equal-cost move is pruned rather than trusted (an equal gate tuple is no tie-break in an unevidenced move's favour); the frame tier keys on the netlist, not `pin_count == 0`; off-board damage witnesses come from a pad-centre census.
- **Reseat**: a repair search that starts from the current pose knows nothing at 40 mm — `place_seed --reseat` lifts a far-from-home part and re-seats it on the fitted family orbit instead of nudging it, behind a 3-conjunct gate.
- **Instrument honesty**: off-outline pad copper measured against the true Edge.Cuts outline per-pad (the placement half's top-priority gate — it converts one-for-one into unrouted nets); `check_assembly` publishes `buildable`/`verdict`/`locked_contacts` and keys its oracle memo on fill cost, not a path every step renames; `check_reachability` reserves exit 1 for CAGED and exits 2 when the question was not answerable; `check_floorplan --require-rules` refuses a vacuous pass; the emit-intent edge band is capped so 160 mm of damage cannot launder itself into "by the edge, by design".
- **Deadlines**: every placement search takes `--deadline` (exit 7, summary guaranteed) through the `krt_deadline` module #616 introduces.
- **converge + check_complete**: the ledger grows lens-verdict grammar with a lens-vs-score contradiction refusal (a PASS beside the numbers that refute it is evidence about the verifier, not the board), commensurability guards, `--exhausted <half>` declarations with mandatory reasons, `--score-file` (a 49 kB score no longer dies on argv limits), and a four-code verdict; `check_complete` is the close-out instrument that fails closed — orphan stubs, weird copper, pad overlaps, and fab floors compared against the AUTHORED project so a board cannot grade itself against floors it rewrote.

## Stress rig (`tests/stress/`)

stage_blind (the dose is checked against the material floor and redrawn — "damage applied" on a clipped dose voids the recovery half before it starts), fence_audit (content-based truth fence; **this commit also repairs the file at tip, which is currently a SyntaxError** — a #522 shim was inserted mid-expression at line 58), arm_report, qualify_subject ("is this board a usable perturbed subject" in seconds), strip_copper_only, calibrate_congestion_ratio (its first result refuted the threshold it was built to calibrate; the gate now reports instead of gating), and `_truth/` staging throughout.

## Live validation

This exact engine + rig closed a blind perturbed board end to end yesterday: piantor staged blind (kind sealed), placement half repaired a ±(72.6, 6.4) mm two-group exchange to all-gates-green with an independent verifier PASS, routing half took it to blocking 0 / DRC 0 at authored floors. The run's report and ledger are reproducible from the rig in this PR.

## Testing

Targeted suites green on this branch: place_seed 37/37, reconstruct (0 errors; corpus-dependent checks skip), slot-fit, stage-blind dose 6/6, fence-and-channels 8/8, arm_report/verdict/check_complete suites, converge 27, render_placement 28, reachability. Five tests in this PR's set are deliberately red until the skill PR stacks on top (they pin `board_score`/loop-driver shares that land there) — each is named in its commit. Full `run_all` on the stack: 294 passed with every red pre-existing at tip on our test box (tip baseline measured; see #615 for why a fresh tip clone cannot run at all).

### DIFF
--- a/.gui-parity-checked
+++ b/.gui-parity-checked
@@ -627,3 +627,19 @@ HEAD 2026-07-30 parity audit 191abc4..d73b1d6 (3 commits). Outcome: no gaps.
     gate runs a fanout step; ea992d9), its stale _path bootstrap
     (3c4fc48), and the #562 rename breaking every recorded manifest's
     repair step (relocate_moved_scripts rename table, 89914ce).
+
+ADDENDUM 2026-08-12 (PR: board floors + false success) -- scope: the one
+GUI-relevant change in this PR is kicad_routing_plugin/swig_gui.py
+_effective_geometry_floor gaining the declared-0 guard (a board writing 0 for
+"not configured" was taken as a real hole-to-hole floor by the GUI while the
+CLI's resolver treated it as unset: measured, GUI drilled h2h 0.1 vs CLI 0.2,
+same board, same settings, under fab-overrides). The plane-edge helper beside
+it already guarded > 1e-9; this closes the divergence in the other helper.
+  * Pinned by tests/gui_parity/test_geometry_floor_leak.py (runs under real
+    KiCad python; "declared-0.0 hole_to_hole, fab floor 0.1 -> GUI 0.2
+    (CLI 0.2) ... 0 problem(s)").
+  * Parity evidence on the source branch (KiCad 10.0, pcbnew 10.0.0,
+    wx 4.2.2): test_gui_engine_parity VERDICT PARITY, copper sets IDENTICAL
+    (segments 1307=1307, vias 144=144, 0 one-sided); test_manifest_plan_parity
+    OK; test_cli_postpass_coverage 7 registered / 14 in use / 1 acknowledged
+    CLI-only (run_drc fanout graze audit, report-only) / 0 unreachable.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -7,6 +7,18 @@ Linux distros). On Windows, if `python3` is missing, fall back to `py -3`
 or `python` — don't retry blindly. Add `-X utf8` when a script prints
 special characters (Ω etc.) to avoid Windows encoding errors.
 
+**On Windows/Git Bash, `export MSYS2_ARG_CONV_EXCL='*'` before any command
+carrying NET NAMES.** MSYS2 rewrites any argument starting with `/` into a
+Windows path, and **every KiCad net name is `/`-prefixed**: `/+1V1` reaches the
+tool as `C:/Program Files/Git/+1V1`. Nothing warns, because a tool cannot tell a
+mangled net name from a net that does not exist. Measured: 61 nets passed to
+`--ignore-nets`, 4 survived (the four not starting with `/`), and the resulting
+render reported 1315 crossings where the truth was 357. It hits `--nets`,
+`--ignore-nets`, `--power-nets`, `--rip-existing-nets` — every net-name argument.
+**The variable is not free**: it also disables conversion for legitimate paths,
+so `~/Documents/...` then arrives as an unusable `/c/Users/...`. Pass
+Windows-style paths (`C:/Users/...`) in the same command.
+
 ## Building the Rust Router
 
 Use `build_router.py` to build the Rust router:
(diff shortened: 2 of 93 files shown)


edgehero:
Three commits added after opening, all from the run 19 exercise of exactly this PR's code (the blind piled board written up in discussion #118):

- check_assembly now counts coincident-origin stacks as blocking: three parts at one coordinate graded buildable / blocking 0 twice in that run because their rotated pads happened to interleave. New JSON channel `coincident_origin_groups`, marker parts (mouse bites, mounting holes) exonerated, pinned by `tests/test_assembly_coincident_stack.py`.
- The pile-degenerate seed no longer licenses pad intersections: every `pads_ok` conjunct is baseline-relative, and a pile baseline (huge `base.pad`, `stack=True`) accepted any smaller intersection — it produced real blocker pairs in the run. Guard at the `seed_baseline` choke point (a seed bucket holding 3+ parts returns `ZERO_SHORTFALL`), two-part by-design overlaps keep their license, pinned by `tests/test_pile_baseline_license.py` including a mutation check.
- Riders: `qualify_subject --json` creates its output directory (the run lost a finished qualification to a missing dir), and the RUNBOOK gains a paragraph on choosing damage kinds the truth fence can actually adjudicate.

The larger placement-engine gaps the same run exposed are deliberately NOT patched here — filed as #628, #629, #630 with the measurements. Full learnings: https://github.com/edgehero/KiCadRoutingTools/blob/run19-assets/LEARNINGS.md

drandyhaas:
Merged as 12 of 13 commits, pushed to `placement` as `e3bf2581..774d8780`. Closing in favour of that — thanks, the placement half of this is good work and the instrument-honesty theme is exactly right.

## What landed

Cherry-picked with `-x`, authorship preserved, applied clean in order:

| yours | on `placement` | |
|---|---|---|
| `4c457b2dd` | `f5cee8777` | reconstruct: the ladder trusted any equal-cost move |
| `bfc2ac812` | `046b43364` | reseat: a repair search that starts from the current pose |
| `203f052dc` | `cc1bbe88e` | placement instruments: measured against a bounding box |
| `724ddaa19` | `c9e2fed70` | deadline: a placement search had no clock |
| `84315259e` | `e8c71ee5b` | converge + `check_complete.py` |
| `38f49096e` | `7b3ded33b` | film: 377 frames over a GIF that holds 348 |
| `cd4530716` | `6a39fb972` | stress rig: the truth fence had three open channels |
| `28a42c4e6` | `8f7feecdd` | CLAUDE.md: what a placement run is FOR |
| `b30fc7af8` | `4f1638e6b` | tests: reconstruct suite imports the #522 layout |
| `f413026d3` | `77ca1885f` | check_assembly: three parts stacked at one origin |
| `9fdbbf192` | `5d35a1157` | legality: a pile seed licensed any smaller pad intersection |
| `d78596b68` | `774d87808` | stress rig: run 19 lost its qualification JSON |

I checked the `fence_audit.py` claim rather than taking it: it really is a `SyntaxError` at line 58 on `placement` tip, and `6a39fb972` really does repair it. Good catch.

## Reading the diff

For anyone finding this later: the PR page showed 29 commits / +13,661, but the real delta against `placement` was 13. #616's routing commits were already cherry-picked onto `placement` on 08-12, so the stack double-counted them. The number that matters is:

```
git log --oneline --cherry-pick --right-only origin/placement...<head>
```

## What did not land: `c90fe328` (`diff_pair_custody`)

This is the one commit held back, and it is byte-for-byte the commit that was dropped from #616 last week for cause.

The rule it adds — `outcome == 'partial'` **and** both members incomplete ⇒ demote from `routed_diff_pairs` — misfires on multipoint pairs whose extra pads are *deliberately* deferred to single-ended follow-up. tigard's `/USB_D` is exactly that shape: a coupled, DRC-clean trunk plus redundant J1-row pads that the flow defers on purpose. It takes `test_tigard_usb_diff` from 12/12 to 9/12 with copper unchanged, verified in both trees.

The commit message cites run-11 `/USB_D` reporting "1/1 routed" with both members in `incomplete_members` — but that is the designed flow, not the #514 defect wearing a new label. Member **count** cannot separate the two cases. The discriminator that can is **"no coupled trunk copper kept"**: a pair with no surviving coupled trunk has nothing to credit and should be demoted; a pair with a clean trunk and deferred extra pads should keep its credit regardless of how many members read incomplete.

Worth recording what this commit does and does not reach, since it is easy to over- or under-estimate: `member_audit_mismatch` feeds only the routed/partial summary buckets and `successful -= len(partial_diff_pairs)`. It does not change copper, and it does not change protected nets — `note_protection_candidates` reads `routed_results` directly and runs before the audit bucketing. So it is a reporting change. But it flips a board's recorded verdict, which moves A/B baselines and `place_route_loop` failure counts, which is why it needs to be right.

A fresh PR with the trunk-copper discriminator and a tigard regression test is very welcome.

## Test state

14 targeted suites green. Four red, all accounted for and none caused by the picks:

- `test_converge` — `loop_driver.final_record_command` missing
- `test_broken_worklist`, `test_run9_check_complete` — `board_score` does not exist on this branch at all

Those three are the skill-PR-side dependencies your description names, so they will clear when that lands. The fourth, `test_orbit_fit_noop`, asserts `>= 30` corpus boards against a local corpus of 29 — environmental, but a fixed threshold on a corpus count will keep doing this to whoever runs it next.

Both no-wx parity gates (`test_manifest_plan_parity`, `test_cli_postpass_coverage`) pass. The `--deadline` entry in `IGNORE_FLAGS` is the right call, and the comment explaining why is appreciated.

## Unrelated find, since you are in this area

`tests/stress/ab_replay_grade.py` is a `SyntaxError` at line 241 on `placement`:

```python
sys.path.insert(0, os.path.join(str(REPO / 'py_router')), 'py_placer'))
```

Same botched-#522-shim bug class you fixed in `fence_audit.py`. Broken at `e3bf2581`, still broken at this PR's head, fine on `main`. That is the primary no-LLM A/B replay tool the runbook points at, so `placement` currently cannot run it. Not yours to fix here — flagging it because it will bite the next replay on that branch.

PR #624, opened 2026-08-12
the combined skill: a loop that refuses to be finished until it is measured to be

edgehero:
**Stacked on #616 and #623** — this branch includes both; the skill work is the 6 commits from "loop driver: L5's printed close-out command could not run as printed…" onward. After they merge this rebases to show only its own diff.

The combined placement-and-routing skill: a loop that refuses to be finished until it is measured to be.

## What this carries

- **`plan-pcb-placement-and-routing`** — SKILL.md + `loop_driver.py` (a near-rewrite of the PR-#576 version): stages L1–L5 with refusal gates on every hand-off, both halves delegated to teammates with verbatim prompts and pre-named contract paths, cycle numbering with `_cN` artifact suffixes so cycle 1's authored baseline survives later cycles, and a close-out that must survive `check_complete` plus a cross-check against the ledger's recorded lens verdicts — the one contradiction that always refuses is DONE-EXHAUSTED beside a failing lens. The driver's self-test pins 147 checks, including every board-binding gate (a close-out for a different board, a score with a foreign `board_sha`, a malformed count — each refusal has a test that fails if it is deleted).
- Fixes from the pre-PR audit, each pinned: L5's printed `converge record --final` command now **runs as printed** (three `--lens` slots; the verdict names STUCK/BUDGED are first-class stop conditions beside a FAIL lens, tested by executing the exact printed text); `_verdict` abspaths its files (a missing ledger reads identically to an empty one — measured — so the wrong cwd was a confident go-round-again derived from nothing); cycle-≥2 close-outs grade against cycle 1's frozen baseline; an unreadable score is a re-score refusal, never a ship ceremony.
- **`plan-pcb-placement`** — read-mandates as gates in the driver; the P-close congestion threshold withdrawn on measurement (the premise inverts on 1 of 3 corpus boards; it reports and demands a disposition instead of gating).
- **`plan-pcb-routing`** — `board_score.py` binds its payload to the graded board by sha, fails closed on DRC parse, and publishes the unrouted shape; references (convergence, evidence-map, verifier-prompts) updated to match.
- **Instruments the drivers consume**: `check_drc --json` (full machine result, per-type contact headers), `check_channels` graded at the board's own floors with `deficit_faces`/`lost_last_lane`, `list_nets.board_floor_declaration`, and `fix_kicad_drc_settings`' `fab_floor_origin` — the FAB FLOOR RELAXED banner baselines on the board's ORIGINAL declaration, not the immediate input (measured: 17 sub-floor vias in silence without it). The key rides the existing `kicad_routing_tools` project namespace, same as #521's `protected_nets`; flagging that format decision here deliberately.

## `routing_driver.py` is not in this PR — deliberately

Upstream deleted it twice by name ("this branch reaches routing through the SKILL, not a driver"), and this PR honors that: the read-mandates-become-gates logic already lives in `placement_driver.py` (9/9 refusal checks green here), the loop driver's eight routing hand-offs point at `/plan-pcb-routing` the way the #576 merge template did, and nothing else is salvaged. `git ls-files | grep routing_driver` is empty on this branch.

## Live validation

This loop closed its first blind perturbed board end to end yesterday, on exactly this driver: piantor, staged blind, blocking 37 → 0, both halves delegated, two waivers taken by name with reasons on the record, placement closed by `--exhausted` declaration, and the final ledger row written by the printed command — verbatim. The refusals fired four times during that run; each one was answered by producing what it asked for.

## Testing

`loop_driver.py --self-test` 147/147; `placement_driver.py --self-test` 58/58; converge 27, broken-worklist 11, handback contract 10, skill commands 10, deadline 5, run12 tools 72, refusal checks 9, unrouted shape 13/13, gate conjuncts 14, channels honesty 14, instruction quality 17, fab-floor origin 9, rule-pair classification 10/10 (routes live). The five tests deliberately red on #623 alone go green on this branch — the stacking proof that their subjects land here.


edgehero:
Two commits added after opening, both from run 19 driving this exact loop end to end (discussion #118):

- `--accept-unclosed` / `--accept-residue` gain `action='extend'`: two occurrences of the same flag silently replaced each other (measured live at L5 — a named waiver vanished and the cross-check re-fired). Repeats now accumulate; bare-flag and unknown-name refusals unchanged; two new self-test pins (147 → 149).
- The fence clause in the delegated teammate prompts now states the tools-only / disclosure duty: the run's placement teammate hand-rolled a placer through five versions before being challenged. The clause names the rule (a disclosed hand-assist is a finding; an undisclosed one invalidates the run), pinned in `tests/test_run15_handback_contract.py` for both emitted prompts. One re-pin disclosed in the commit: the L1 prompt line cap moved 70 → 74 to carry the clause.

drandyhaas:
Merged as 9 of 10 commits, pushed to `placement` as `774d8780..d6d7b6ed` (plus one follow-up of mine). Closing in favour of that. The loop-driver work and the instrument honesty behind it are good, and the refusal-gate design in particular is the right shape.

## What landed

Cherry-picked with `-x`, authorship preserved, applied clean in order:

| yours | on `placement` | |
|---|---|---|
| `f2ba2da4a` | `ee6c3ad1c` | loop driver: L5's printed close-out could not run as printed |
| `3c01b1945` | `96bbad438` | placement skill: read-mandates became gates |
| `1687bf8fb` | `59bc62d78` | routing skill: board_score published scores off the wrong board |
| `d350a1902` | `4ac459cdb` | review skill: the checker roster could not see a bad via |
| `3ae374417` | `ca1055b2b` | instruments: check_drc `--json`, check_channels board-first floors |
| `4ab83db03` | `37bce4919` | fab-floor origin: the RELAXED banner baselined on the immediate input |
| `2b5578ac4` | `ec09c52f0` | board_score: parseable board reported unparseable in-process |
| `54e478e9f` | `c0f860779` | loop driver: a repeated `--accept-*` flag silently REPLACED the first |
| `936690538` | `14792c893` | loop driver: the fence named OPEN, nothing about WRITE |

Plus `d6d7b6edc`, mine — see "Three tests were grading the host" below.

The stacking claim checks out: `test_converge` and `test_broken_worklist` were red on #623 alone (`loop_driver.final_record_command` and `board_score` do not exist there at all) and go green here.

## Reading the diff

The PR page showed +21,067 over 100 files; the real delta against `placement` was **10 commits**, because #616's and #623's shares are already merged. For anyone finding this later, the number that matters is:

```
git log --oneline --cherry-pick --right-only origin/placement...<head>
```

## The routing question, and how it was answered

Three of the ten touch shared routing code, so this got checked rather than assumed:

- `4ab83db03` — floor VALUES and the `changes` computation are untouched; the commit rebaselines the disclosure and makes both fronts list their writeback instead of counting it. No routing change. It does write a new `kicad_routing_tools.fab_floor_origin` key into the output project, and riding the existing namespace rather than a new sidecar is the right call — `copy_board` and the project-carry chain move it for free, exactly as you argue.
- `3ae374417` — detection, counts and exit status untouched. **Verified rather than assumed**: `check_drc` A/B'd pre- vs post-merge across the three tracked boards that carry real copper (`routed_output`, `orangecrab_ext_pll`, `rp2350_fpga_eensy_prePlane`), 2,750 violations total — totals AND per-type breakdowns byte-identical. The new ` -- N in CONTACT` header suffix is compatible with every parser in the repo: the three that read that line (`test_keepout`, `test_guide_corridor`, `test_soft_joint`) pin `violations \((\d+)\)` with no trailing anchor. Nice detail — #623's `grade_final` regex fix already anticipated this exact producer, so the pair closes cleanly.
- `c90fe328` — held back, see below.

One latent gap worth a look if you revisit `4ab83db03`: `fab_floor_origin` is seeded only on a writeback that HAS changes, because the `if not changes:` branch returns before the `setdefault`. A chain whose first step is a no-op therefore seeds the origin from an already-relaxed floor — the same silence the commit exists to close, one step further along.

## What did not land: `c90fe328` (`diff_pair_custody`)

This is its **third** appearance — dropped from #616 for cause, held back from #623, and unchanged here.

The rule it adds (`outcome == 'partial'` **and** both members incomplete ⇒ demote) misfires on multipoint pairs whose extra pads are deliberately deferred to single-ended follow-up. tigard's `/USB_D` is that shape: a coupled, DRC-clean trunk plus redundant J1-row pads deferred on purpose. It takes `test_tigard_usb_diff` from 12/12 to 9/12 with copper unchanged.

Scoping it precisely, since that is easy to over- or under-state: `member_audit_mismatch` feeds only the routed/partial summary buckets and `successful -= len(partial_diff_pairs)`. It does not change copper, and it does not change protected nets — `note_protection_candidates` reads `routed_results` directly and runs before the audit bucketing. So it is a reporting change; but it flips a board's recorded verdict, which moves A/B baselines and `place_route_loop` failure counts, which is why it has to be right.

Member COUNT cannot separate the two cases. **"No coupled trunk copper kept"** can. A PR with that discriminator and a tigard regression test is welcome and will go straight in — but please rebase it out of this stack rather than carrying it forward again, since it now has three declines against the same rationale.

## Three tests were grading the host, not the code

`d6d7b6edc` fixes these. All three were green where they were written and red on a Mac, for reasons unrelated to what they pin — worth knowing about because the same three shapes will keep recurring:

1. **`test_run9_check_complete`** built a relative board path from a `tempfile` directory. macOS puts that under `/var`, a **symlink to `/private/var`**, so the `..` chain ascends one directory too few once the kernel resolves it physically. The board did not exist from the new cwd, `check_complete` exited "no such board", and three checks about the close-out DOCUMENT failed having never produced one. Now `realpath`s both sides, with a fallback for the Windows mirror case (temp on another drive, where no relative path exists and `relpath` raises).

2. **`test_run12_tools`**, twice. `sorted(os.listdir(wd)) == ['board.kicad_pcb']` forbids the sibling carry-over `copy_board` performs on purpose (#441), and `kicad_files/*.kicad_prl` is gitignored — so it passed on a clean checkout and failed on any machine where someone had opened that board in KiCad. It now asserts what the fence is actually about: no control and no record inside the work dir. Separately, `--deadline 0.4` encoded "surely less than a route takes", which is a claim about the HOST: `route_diff.py` on `lvds_converter_dualclk` finishes in well under 0.4 s here and exited 0, failing four checks about budget accounting. `SPENT_DEADLINE_S = 0.001` is spent by construction and a slower box only trips it more surely, so the assertion is one-directional now. (Your own `--deadline 0.001` further down the same file was the precedent.)

3. **`test_orbit_fit_noop`** globbed `kicad_files/` for its corpus. That directory accumulates routing OUTPUTS — `*_routed`, `*_fanout`, `*_plane`, `*_connected` — which `.gitignore:49-51` covers precisely because they are per-machine junk: **22 boards on a clean clone, 29 on this one, and ≥30 wherever the MATRIX counts were pinned.** The numbers could only ever be right in one working copy. It now asks `git ls-files` which boards the repo ships, and skips rather than grading against a set it cannot identify.

   Re-measured over the fixed 22: the ladder's top rows move (28/28/28/14 → 17/17/17/11) and **every row from the occupancy guard down is unchanged (7/2/0)** — including the SHIPPED row the file exists for, still firing on zero boards. The artifact boards were all discarded by occupancy anyway, so the conclusion never rested on them. That is a good result for the guard ladder, not a weakening of it.

## Test state

24 suites green across #623 and #624, plus both no-wx parity gates (`test_manifest_plan_parity`, `test_cli_postpass_coverage`).

One pre-existing failure, **not** from this PR: `test_soft_joint` fails 2 checks identically at the pre-merge baseline `774d8780` (verified in a detached worktree). That is placement-branch test debt, same family as the 37 noted on #616.

## Minor

`wk/calibration/RESULT.md` is committed into `wk/`, which is **gitignored** (`.gitignore:66`). It works — ignore rules only affect untracked files — but the file will never appear in anyone's `git status`, and it is inside a directory whose whole point is that it is disposable. Worth relocating if that content is meant to last.

Issue #602, opened 2026-08-10
route_diff reports a pair as 'coupled' while its own MEMBER AUDIT shows member pads still disconnected


drandyhaas:
## Observed

Sets 21-27 baseline wave (99 boards, 2026-08-09/10, commit `fa1e71f`). On **3 boards** `route_diff.py`'s summary claimed success for pairs its own member audit had just flagged as incomplete.

`muzy_zynq4` (set21) is the clearest:

> `route_diff.py` reported `/HDMI_CK` and `/HDMI_D2` as successfully 'coupled' pairs in its own summary, but its own MEMBER AUDIT flagged member pads still disconnected on both — a false-positive success signal.

Also seen on `lwdo_sdr` (set25) and `spartan6_4layer` (set25).

## Why it matters

`diff_pairs_coupled` is the field a caller gates on — the stress RUNBOOK, `ab_replay_grade`'s summary, and the sweep harness all read it to decide whether the diff-pair stage succeeded. When it disagrees with the member audit in the same run's output, the cheap machine-readable signal is the wrong one and the correct signal is buried in prose.

This is the diff-pair analogue of #8 (`Router reports net fully routed but produces disconnected trace segments`, CLOSED), but that issue concerns segment continuity on single-ended nets, not diff-pair member accounting, so the fix there would not cover this.

## Expected

A pair should not be counted in `diff_pairs_coupled` while the member audit reports any of its member pads disconnected. If the two measurements can legitimately differ, the summary should carry the audit's verdict as its own field (e.g. `diff_pairs_member_incomplete`) so a caller can gate on it without parsing the log.

## Impact

Route-quality reporting / router correctness. Silent over-reporting of diff-pair success: a chain that trusts the summary proceeds believing pairs are routed when member pads are open.

drandyhaas:
Fixed in 3ceb773. The wave's own log made this one worth reading carefully — the summary lists were already right, and there were two *different* problems underneath.

## What #514 already handled

`_partial_by_audit` (c501a80, live at `fa1e71f`) already keeps an audit-contradicted pair out of `routed_diff_pairs`. `muzy_zynq4`'s summary did exactly that:

```
"routed_diff_pairs": ["/USB2t_D","/USB2_D","/HDMI_D0","/USB1_D","/USB1t_D","/HDMI_D1"],
"partial_diff_pairs": ["/HDMI_CK","/HDMI_D2"],
```

## Problem 1 — the per-pair record still claimed success

[...]

PR #632, opened 2026-08-13
diff_pair_custody: demote on no coupled trunk copper kept, not on member count

edgehero:
The revised custody demotion, per your review on #616/#623/#624 — rebuilt from scratch on current `placement` (`d6d7b6ed`), not carried from the old stack.

**The discriminator is the one you named: demote only on no coupled trunk copper kept.** The signal is recorded, not derived: `is_diff_pair` is stamped only by the coupled-route constructors (`diff_pair_routing.py` single-leg, `diff_pair_multipoint.py` leg merge), never by single-ended results, and a nonempty `new_segments` on that shared result dict means the coupled route committed real copper. A `partial` pair keeps its `routed_diff_pairs` credit while any coupled trunk copper exists; it is demoted, with the reason printed by name, only when none does. Member count is not consulted. The demotion feeds the existing member-audit channel inside `batch_route_diff_pairs`, so CLI and GUI inherit identically, plus an additive `no_coupled_trunk` JSON key so graders can tell the two demotion causes apart.

**Verification:**
- `tests/test_tigard_usb_diff.py`: **12/12** — all three variants keep `"routed_diff_pairs": ["/USB_D"]`, the deliberately-deferred multipoint shape from your counterexample keeps its credit.
- Custody suite: **37/37** — two new pins, one per direction (partial + trunk kept + both members incomplete → credit kept; partial + no trunk → demoted with the reason line), and fixture E upgraded to carry a real coupled-trunk result.
- Diff-pair battery at this tip, all green: diff_pair_route 5/5, watchy passthrough + hybrid escape, esp_prog 1/1, multipoint_diff_route 7/7, multipoint_failed_edge_reporting 6/6, qfn_csi_underpad_diff 2/2, diff_se_member_fallback 16/16.

**Stated honestly:** the demotion direction is exercised at unit level only — no board in the tracked corpus produces a partial-without-trunk pair in a live route (the battery boards all keep their trunks, which is why they stay green). And when the demotion fires, `route_diff.py`'s generic MEMBER AUDIT MISMATCH line prints beside the specific reason line; rewording that generic line would touch `route_diff.py`, which this PR deliberately does not.

### DIFF
--- a/py_router/diff_pair_custody.py
+++ b/py_router/diff_pair_custody.py
@@ -162,6 +162,29 @@ def build_pair_reports(state, diff_pair_ids_to_route, member_audit,
         incomplete = [nm for tag, nm in (('p', pair.p_net_name),
                                          ('n', pair.n_net_name))
                       if not aud.get(tag, True)]
+        # 'partial' credit discriminator: a partial pair KEEPS its
+        # routed_diff_pairs credit while it kept ANY coupled trunk copper;
+        # it is demoted ONLY when it kept no coupled trunk copper at all.
+        # (An earlier rule demoted on member count -- 'partial' + both
+        # members incomplete -- which misfires on multipoint pairs whose
+        # extra pads are DELIBERATELY deferred: tigard /USB_D keeps a
+        # coupled, DRC-clean trunk while the redundant J1-row pads close in
+        # the designed single-ended follow-up, expected step-level state,
+        # not false success; that rule dropped test_tigard_usb_diff from
+        # 12/12 to 9/12 with the copper unchanged. Incomplete members are
+        # already counted as pad deficit at chain level, so credit with a
+        # real coupled trunk hides nothing.) Trunk signal, recorded not
+        # derived: 'is_diff_pair' is stamped ONLY by the coupled-route
+        # constructors (route_diff_pair_with_obstacles, the multipoint leg
+        # merge) -- never by single-ended results -- and 'new_segments'
+        # nonempty means that coupled result committed real copper.
+        no_coupled_trunk = bool(
+            outcome == 'partial'
+            and not any(rr.get('is_diff_pair') and rr.get('new_segments')
+                        for rr in (_rr_p, _rr_n) if rr))
+        if no_coupled_trunk:
+            print(f"{RED}DEMOTED {pair_name}: no coupled trunk copper kept "
+                  f"-- 'partial' pair dropped from routed_diff_pairs{RESET}")
         rep = {
             'pair': pair_name,
             'p_net': pair.p_net_name,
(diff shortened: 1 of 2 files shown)


drandyhaas:
Merged as a clean fast-forward, `d6d7b6ed..2b7875cc` on `placement`. Thanks for rebuilding this one rather than rebasing the old commit — the discriminator is the thing that was wrong, and you changed it.

## What made this one mergeable

The previous version demoted on `partial` + both members incomplete, which misfires on multipoint pairs whose extra pads are deliberately deferred to the single-ended follow-up. This one demotes only on **no coupled trunk copper kept**, and member count is not consulted. I checked the signal rather than taking it on trust: `is_diff_pair` is stamped into a result dict in exactly two places, both unconditional on a successful coupled construction — `diff_pair_routing.py:4468` (single leg) and `diff_pair_multipoint.py:984` (leg merge). No single-ended path stamps it, so "`is_diff_pair` + nonempty `new_segments`" really does mean a coupled route committed copper.

## Routing-behavior verdict: no copper change

The thing I had to rule out is #521. If `routed_diff_pairs` shrinking fed the protected-nets writeback, a demotion would make a later chain step willing to rip coupled pair copper — a real routing change hiding behind a reporting patch. It does not, because protection is not keyed on that list:

- `route_diff.py:~1229` `note_protection_candidates` keys on `routed_results[nid]['is_diff_pair'] and not failed` — the raw result, never the audit verdict — and it runs *before* the reports are built.
- The post-route cleanup scope (`dp_scope`, ~1253) is likewise built from `routed_results` + `_member_connected`, also before `build_pair_reports` (~1291).

`build_pair_reports` runs after both, and nothing downstream of it touches geometry. So this is a verdict change, not a copper change: the pair moves `routed_diff_pairs` → `partial_diff_pairs`, `successful` drops by one (the GUI reads it raw), the `DEMOTED` line prints, and `no_coupled_trunk` appears additively. Worth stating explicitly for future readers: because `tests/stress/ab_replay_grade.py` consumes `partial_diff_pairs` as a final status, a firing demotion **can** shift A/B grading tables and could nudge a loop-driven retry decision. It just never moves a track.

Parity needed nothing: `build_pair_reports` is called from inside `batch_route_diff_pairs`, so CLI and GUI inherit identically.

## Verification

Reproduced your numbers in a dedicated worktree at `2b7875cc`, graded against baseline `d6d7b6ed`:

- `test_tigard_usb_diff` **12/12** — the exact regression that sank the previous three attempts.
- `test_diff_pair_custody` **37/37**, vs **32/32** at baseline, so the five new checks are real and both directions are pinned.
- Battery green and identical to baseline: `diff_pair_route` 5/5, `multipoint_diff_route` 7/7, `diff_se_member_fallback` 16/16, `multipoint_failed_edge_reporting` 6/6, `esp_prog` 1/1, watchy passthrough + hybrid escape, `qfn_csi_underpad_diff` 2/2.
- No `DEMOTED` line fired on any real board in that battery.

That last point matches your own disclosure that the demotion direction is exercised at unit level only, and I want to credit stating it plainly — it is exactly the kind of thing that is tempting to leave out. I accepted it anyway because it is the conservative direction: it fires only when there is literally no coupled copper to credit, so the failure mode of being wrong is a pair reported worse than it is, never a false success.

## One flag for whoever merges `placement` into `main`

`main` has independently rewritten the **same hunk** for #602 — it introduces a `mismatch` variable and demotes `outcome` to `'incomplete'` when a `coupled` pair has incomplete members. `placement` + this commit will conflict there. The two changes are **complementary**, not competing: #602 covers *coupled* + incomplete members, this covers *partial* + no trunk. Resolve by keeping both. Your `outcome == 'partial'` test still reads correctly after #602, since that demotion only rewrites `coupled` → `incomplete` and leaves `partial` alone.

Re the generic `MEMBER AUDIT MISMATCH` line printing beside your specific reason line: agreed that rewording it belongs in a `route_diff.py` change rather than here. Worth noting the reason line can print with an empty member list, when a pair has no trunk but both members happen to be connected by other copper — cosmetic, and a fine follow-up if you want it.

Issue #457, opened 2026-07-21
placement: quench output differs across processes, and the writer rounds coordinates to 4 significant digits


edgehero:
Two small defects that both undermine trusting a quench result, plus a related fresh-clone test problem. Quench line refs are against 17d4670 (PR #455 head).

1. Cross-process nondeterminism. `_net_points` iterates `self.net_refs[net_id]`, a Set[str] (placement/quench.py:223, and again at :577 inside the swap evaluation). Point order feeds `compute_mst_edges`, whose tie-break is first-index-wins (connectivity.py:1227-1228), so equidistant ties resolve differently per PYTHONHASHSEED: different airwires, different crossing counts, different accepted moves. Identical inputs should give identical boards (that was the point of the 0.17.7 plane-repair reproducibility fix), and A/B comparison is how changes get judged here; today a placement A/B needs a pinned hash seed to mean anything. The #430 tests already had to scope every equality assertion same-process because of this (tests/test_quench_neighbor_lists.py:15-17). `sorted()` over the refs at both sites fixes it; everything downstream is already insertion-ordered.

2. Writer precision. placement/writer.py:104-106 writes `(at {x:.4g} {y:.4g} ...)`. At coordinates over 100mm that is 0.1mm resolution: format(148.259, '.4g') gives '148.3'. So positions the quench validated get shifted by up to 0.05mm per axis after all clearance checks ran, clearance between two adversely rounded parts erodes by up to about 0.1mm, and a swap that put a part on its partner's file-precision position does not round-trip. kicad_writer.py uses .6f throughout; the placement writer (including the pad-angle rewrite at writer.py:22) should match.

3. Fresh-clone test runs. startup_checks calls sys.exit(1) at import time (startup_checks.py:45 for missing numpy/scipy/shapely, :106 for a missing or version-mismatched rust router), and route.py runs it at module scope, so pytest collection dies with INTERNALERROR on a clean checkout (tests/test_result_shapes.py imports route at module level). Raising when imported, or deferring the check to main(), keeps the suite collectable. Related fresh-clone hazard: fixture boards keep going missing via gitignore patterns (#190 re-committed two; #245, #432 and #433 all note test_enable_used_layers.py failing on a fresh clone because only the .kicad_pro/.kicad_prl are tracked).

edgehero:
[...]

Issue #426, opened 2026-07-17
test kit-out-plane-connected fails check_drc + check_connected


drandyhaas:
tests/test_kit_route.py has a pre-existing failing tail: the kit-out-plane-connected board grades 30 clearance violations / 47 unconnected per the kicad oracle. Confirmed PRE-EXISTING via worktree A/B at base a1e7bac during the 2026-07-16 wave (oracle-identical artifacts old vs new code) — not caused by the wave. Also note: the kit run mutates kicad_files/*.kicad_pro (DRC-floor writeback) — restore with git checkout afterward; consider pointing the test at a temp copy. Follow-up: root-cause the board's plane-connected chain (likely an old recorded-chain artifact rather than an engine bug), fix or re-record, and make the kit suite leave the repo clean.

drandyhaas:
The repo-hygiene half is fixed and merged to main (merge 0fb589b; fix 25de902); the deep routing issue stays open.

**Root cause = genuine routing-completion difficulty, NOT a stale fixture.** Re-running the identical chain to *fresh* temp paths (no sibling `.kicad_pro` to read back) gives **byte-identical 112 DRC / 51 disconnected** (repo checkers; the issue's 30/47 is the same board via the kicad-cli oracle). Step 1 (`route.py`) leaves ~50 fine-pitch U102↔U301 signal nets **boxed in by neighbouring copper and unrouted** (→ the unconnected), plus a few real ~0.02–0.05mm congestion grazes. The input board has no sibling `.kicad_pro`, so it can't be contaminated. Deep completion problem — **left open**.

**Repo-dirtying fixed:** the kit run wrote sibling `.kicad_pro` DRC floors onto tracked `kit-out*` files (and read stale floors back on re-run). The test now routes all `kit-out*` artifacts into a `tempfile.TemporaryDirectory` (input read read-only), and the 6 stale `kit-out*.kicad_pro/.kicad_prl` fixtures are `git rm --cached` (already gitignored). **`git status` is now clean after a full run.**

Remaining scope for this issue: root-cause/relieve the U102↔U301 congestion so the board completes (a routing-engine problem), then optionally re-baseline the expected counts.

PR #634, opened 2026-08-14
fix: pad keep-out generator emitted zero cells on a warm memo cache

theironchef:
`iter_pad_blocked_cells` silently yields **no keep-out cells at all** once
`pad_blocked_cells_array` has run for the same pad geometry.

Introduced in 02f5375a (`perf: pad-offset memo + LRU/budget rework of the raster
caches`), shipped in **v0.20.4**. `tests/test_pad_offset_keepout.py` fails on stock
`main` because of it.

## The bug

02f5375a added the `_PAD_OFFSETS_CACHE` fast-path to `pad_blocked_cells_array`, and
pasted the same six lines into `iter_pad_blocked_cells`:

```python
offs = _PAD_OFFSETS_CACHE.get(key)
if offs is not None:
    out = np.empty_like(offs)
    np.add(offs, np.array([[pad_gx, pad_gy]], dtype=np.int32), out=out)
    return out          # <-- iter_pad_blocked_cells is a GENERATOR
```

`return <value>` inside a generator is not a result — it is a bare `StopIteration`.
So on every cache hit the generator yields nothing and the pad loses its keep-out
entirely, with no error and no warning.

The cache is shared between the two twins and **only the array twin writes to it**,
so the trigger is just "`pad_blocked_cells_array` ran first for this geometry" — the
normal case in any real obstacle build.

## Reproduce it

Save as `repro.py` in the repo root, on stock `main`:

```python
import os, sys
ROOT = os.path.dirname(os.path.abspath(__file__))
sys.path.insert(0, ROOT); sys.path.insert(0, os.path.join(ROOT, 'py_router'))
from routing_utils import iter_pad_blocked_cells, pad_blocked_cells_array

# One ordinary pad: 1.95 x 0.60 mm, 0.30 mm keep-out margin, 0.05 mm grid.
geom = dict(half_width=0.975, half_height=0.30, margin=0.30,
            grid_step=0.05, corner_radius=0.0, off_y=0.016)

print("generator, cold cache :", len(list(iter_pad_blocked_cells(0, 0, **geom))))
print("array twin            :", len(pad_blocked_cells_array(0, 0, **geom)))
print("generator, warm cache :", len(list(iter_pad_blocked_cells(0, 0, **geom))))
```

**Before (stock `main`, 52b006c7):**

```
generator, cold cache : 1214
array twin            : 1214
generator, warm cache :    0     <-- same pad, same args
```

**After (this PR):**

```
generator, cold cache : 1214
array twin            : 1214
generator, warm cache : 1214
```

The existing test catches it too — it calls the array twin before the generator, so
the cache is warm by the time it grades them against each other:

```
$ python3 tests/test_pad_offset_keepout.py     # on stock main
  FAIL  generator blocks the sub-cell-inside cell too
  FAIL  generator matches array twin
2 failure(s)
```

## The fix

**Drop the memo from the generator** rather than repair it.

Repairing it (yielding the cached rows instead of returning them) would fix the
crash-free-but-empty symptom, but sharing the twin's cache is wrong on its own
terms: this scalar rasterizer is the *independent reference implementation* that
`test_pad_offset_keepout.py` grades the vectorized twin against. Serving it out of
that twin's cache makes the cross-check vacuous — and only when the cache happens to
be warm, so it would pass or fail by call order.

The perf case for the memo (2.04M calls / 84s on orangecrab) is entirely
`pad_blocked_cells_array`'s. That is the twin every obstacle builder actually calls
(`obstacle_map`, `obstacle_cache`, `plane_obstacle_builder` all reach for the array
form), and it is **untouched** by this PR — no perf change. A comment at the site
records both reasons so the fast-path doesn't get pasted back in.

I also swept the tree for the same shape; this was the only generator in the
repository with a value-returning `return`.

## Blast radius

Small, and deliberately so:

- `pad_blocked_cells_array` — **unchanged**, so the hot obstacle-build path and the
  02f5375a speedup are unaffected.
- No production caller reaches `iter_pad_blocked_cells` today (the three modules
  that import it have all migrated to the array twin), so this is a latent defect
  rather than an active routing bug — but it is a documented public API
  (`docs/api-routing-config.md`, README) that currently returns nothing, and it
  breaks the suite at HEAD.
- No CLI/GUI parity surface: no flag, no default, no engine parameter.

## Testing

```
tests/test_pad_offset_keepout.py       FAIL on main  ->  PASS  (8/8 checks)
tests/run_all.py --fast                3 failed      ->  2 failed
tests/gui_parity/test_manifest_plan_parity.py    PASS (35 flag-checks, 0 mismatches)
tests/gui_parity/test_cli_postpass_coverage.py   PASS (0 failures, 0 warnings)
```

The two remaining `run_all --fast` failures are pre-existing and environmental on
this box, unrelated to this change:

- `test_gui_finalize_oracle_inrun.py` — wants `kicad_files/kit-out-plane.kicad_pcb`,
  which isn't in the repo.
- `test_live_board_zone_deletion.py` — Windows `cp1252` can't encode `Ω` from
  `plane_resistance.print_single_net_resistance`; passes under `python3 -X utf8`.

The test is also strengthened so this can't regress quietly: it now names the
warm-cache case, asserts the generator is non-empty, and checks both call orders,
instead of only tripping the set-compare as a side effect of test ordering.

### DIFF
--- a/py_router/routing_utils.py
+++ b/py_router/routing_utils.py
@@ -205,13 +205,21 @@ def iter_pad_blocked_cells(
     # sub-grid deviation (diff pair P/N offsets) pass a larger buffer.
     if corner_buffer is None:
         corner_buffer = grid_step / 2
-    key = (half_width, half_height, margin, grid_step, corner_radius,
-           corner_buffer, off_x, off_y, rotation_deg)
-    offs = _PAD_OFFSETS_CACHE.get(key)
-    if offs is not None:
-        out = np.empty_like(offs)
-        np.add(offs, np.array([[pad_gx, pad_gy]], dtype=np.int32), out=out)
-        return out
+    # DELIBERATELY NOT MEMOIZED -- do not paste the _PAD_OFFSETS_CACHE
+    # fast-path from pad_blocked_cells_array in here. Two reasons:
+    #   1. This is a GENERATOR. `return <array>` inside one is not a result,
+    #      it is a bare StopIteration -- the fast path would yield NOTHING
+    #      and the pad would silently get no keep-out at all. That shipped
+    #      once (02f5375a, v0.20.4) and emptied this function on every warm
+    #      cache; tests/test_pad_offset_keepout.py pins it.
+    #   2. Even done correctly it would be wrong to share: this scalar
+    #      rasterizer is the INDEPENDENT reference implementation that
+    #      test_pad_offset_keepout.py grades the vectorized twin against.
+    #      Serving it from the twin's cache makes that cross-check vacuous
+    #      (and only when the cache happens to be warm, so it would pass or
+    #      fail by call order). The perf case for the memo is entirely
+    #      pad_blocked_cells_array's -- that is the twin every obstacle
+    #      builder actually calls.
     rotated = abs(rotation_deg) > 1e-9 and abs(abs(rotation_deg) - 180.0) > 1e-9
     if rotated:
         _rrad = math.radians(rotation_deg)
--- a/tests/test_pad_offset_keepout.py
+++ b/tests/test_pad_offset_keepout.py
@@ -18,8 +18,11 @@ from routing_utils import pad_blocked_cells_array, iter_pad_blocked_cells
 
 def run():
     fails = []
+    total = 0
 
     def check(name, cond):
+        nonlocal total
+        total += 1
         if not cond:
             fails.append(name)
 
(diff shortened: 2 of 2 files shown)


drandyhaas:
Merged, thanks — the diagnosis is exactly right, and dropping the memo rather than repairing it is the correct call for the reason you give (the scalar rasterizer is the independent reference the cross-check grades the vectorized twin against).

**Verified no harm from this in shipped v0.20.4.** I did not want to rely on grep for that, so I checked it two ways:

- **Static:** `iter_pad_blocked_cells` is *imported* by `obstacle_map`, `obstacle_cache` and `plane_obstacle_builder`, but every remaining mention in those modules is a comment — there is no call site in `py_router/`.
- **Dynamic:** wrapped both twins with counters and ran real work on `splitflap_driver`. A **route step** called `pad_blocked_cells_array` 5,050 times and `iter_pad_blocked_cells` **0** times; a **plane step** (the generator’s historic caller) also called it **0** times.

So v0.20.4 ships the defect but never executes it — latent, not an active routing bug, exactly as you described. No re-release needed; the fix rides the next one.

Both follow-ups you flagged are also fixed on `main` (4364d948), and both turned out to be real defects rather than environment quirks:

1. **`test_gui_finalize_oracle_inrun.py`** — worse than a missing file. `kicad_files/kit-out-plane.kicad_pcb` is gitignored *and produced by nothing* since #426 moved `test_kit_route.py`’s outputs to a temp workdir, so it only existed on machines that had run the pre-#426 chain: crash on a fresh clone, and a local "pass" against a board of unknown provenance. It now builds from the tracked root via `fixture_boards.ensure()`. The #457 gate missed it because its consumer list is hand-maintained, so there is now a derived check that scans for any test reading an untracked `kicad_files/` board.

2. **`test_live_board_zone_deletion.py`** — not just a `-X utf8` matter. `route_planes.py` already calls `enable_utf8_console()`, but that test calls `create_plane` **in-process**, so no `__main__` runs and a Windows user hits the same crash from the shared engine. Fixed at the print site with `console_encoding.ascii_safe()`, which keeps the real glyph on UTF-8 and spells it `mohm` only where the stream genuinely cannot encode it. Pinned by a new test against a strict cp1252 stream.

215 passed, 0 failed.

Issue #562, opened 2026-08-03
Pours-first / no-taps architecture: planes -> plane-aware fanout -> one arbitrated route pass -> repair (worktree-proven, needs graduation)


drandyhaas:
## The architecture (Andy's design, 2026-08-03)

Replace the tap-centric flow with clear per-stage ownership:

1. **All pours in stage 1** -- including outer-layer floods -- with **NO
   taps** (`KICAD_PLANE_NO_TAPS=1`). Every via-needed pad (deferred or
   skipped) still **seeds the split-layer Voronoi** as a point seed, so each
   rail's cell reserves its territory before any via exists.
2. **BGA fanout is naturally plane-aware** (pours really exist, no
   declarations needed): pour-direct connects balls the fill reaches with NO
   via; drops barrel only the rest.
3. **One route pass with plane nets included** (`route.py --nets '*'` +
   `KICAD_POUR_LAUNCH=1`): the zone-aware connectivity grouping collapses
   pour-reached pads for free; stragglers get **laddered fill targets**
   (best-12 cells at the first rung of 1/2.5/6/15mm containing fill), so the
   router itself places every remaining tap -- priced by all soft costs and
   rippable/reroutable symmetrically against signal nets.
4. **Plane repair shrinks to a backstop** (true islands, region joins).

## Measured (285-ball 0.5mm 6-layer BGA board, full chains, deterministic)

| chain variant | plane-pad owner | end nets-with-issues | post-pour wall |
|---|---|---|---|
| defer-to-fanout (per-ball drops) | fanout | 19 | 554s |
| no-taps, inner pours only | router | 26 | 807s |
| **no-taps + outer flood (the architecture)** | fill + router | **12 (best)** | **524s (fastest)** |

- Pour-direct covered **126/134 plane balls with no via** (8 barrels vs 128
  in the defer arm -- the barrel-economy problem dissolves rather than needing
  the refuted tap-reuse trade).
- The distance ladder cut pour-launch targets **73,323 -> 78 cells** with an
  identical final grade.
- Outer-pour-late doctrine is obsolete in this flow: zones are outlines and
  fill recomputes, so pour everything first; the fanout and router simply see
  them.

## Where the code is

Branch `se-rip-arbitration` (worktree), env-gated and default-off:
- `KICAD_PLANE_NO_TAPS` + universal Voronoi point-seeding (route_planes.py)
[...]

PR #635, opened 2026-08-14
gate: point the finalize-oracle gate at a TRACKED fixture (it has never run)

theironchef:
`tests/test_gui_finalize_oracle_inrun.py` reads a board that **no clone has**, so
the #562 regression it exists to catch has been unguarded since the day the gate
landed.

## The bug

```
25de9022  (#426, 2026-07-18)  untracked kit-out* and added
                              /kicad_files/kit-out*.kicad_pcb to .gitignore
8bfa9011  (      2026-08-04)  added this gate, reading exactly that path
```

The gate was written **17 days after its fixture stopped being tracked**. It has
only ever passed on a working copy where a stale artifact from an older
`test_kit_route.py` run happened to be left in `kicad_files/` — that test now
writes to a temp dir (which is what #426 was), so nothing regenerates it either.

## Reproduce it

Any fresh clone:

```bash
git clone https://github.com/drandyhaas/KiCadRoutingTools.git && cd KiCadRoutingTools
python3 build_router.py
python3 tests/test_gui_finalize_oracle_inrun.py; echo "exit=$?"
```

**Before (on `main`):**

```
FileNotFoundError: [Errno 2] No such file or directory:
  '.../kicad_files/kit-out-plane.kicad_pcb'
exit=1
```

Confirming it can never be there:

```bash
$ git log --all --oneline -- 'kicad_files/kit-out-plane*' | head -1
25de9022 #426: kit route test writes to a temp dir; untrack stale kit-out fixtures
$ git check-ignore -v kicad_files/kit-out-plane.kicad_pcb
.gitignore:48:/kicad_files/kit-out*.kicad_pcb   kicad_files/kit-out-plane.kicad_pcb
```

**After (this PR):**

```
GUI finalize oracle: runs in-run when a board can be staged, falls back to the
post-apply hand-off when it cannot. OK
exit=0
```

## The fix

Regenerating the old fixture is **not** the fix — it is a full route of
kit-dev-coldfire (minutes), which is exactly the cost `tests/fixture_boards.py`
deliberately declines to cover: *"a test wanting those should use a tracked board
instead."*

So this uses a tracked board. I probed every tracked board in `kicad_files/` for
the two properties the fixture note requires:

| tracked board | zone nets | reaches the oracle leg? |
|---|---|---|
| `flat_hierarchy.kicad_pcb` | `GND` | no — already fully connected, never reaches the finalize |
| `sonde_u.kicad_pcb` | `GND` | yes, but the pour reads complete |
| **`lvds_converter_dualclk_gnd.kicad_pcb`** | `/GND` | **yes, and the pour reads INCOMPLETE** |
| `kit-dev-coldfire-xilinx_5213.kicad_pcb` | `+3.3V`, `GND` | yes, but it is the minutes-long route |

`lvds_converter_dualclk_gnd.kicad_pcb` keeps both properties — a real pour, and one
that reads incomplete (the run's own improvement gate reports it gaining `/GND`,
42 → 41 disconnected pads) — and the whole gate runs in **under a second** on it.

## The new fixture is not a vacuous pass

The fixture note warns that *"two earlier fixture picks here made the gate vacuously
pass"*, so I checked rather than assumed. Re-introducing the #562 deferred-oracle
implementation (forcing the `stage_board_fn` branch off in `route.py`) makes the
gate fail with **both** of its intended messages:

```
FAIL:
  - stage_board_fn was never called -- the finalize did not try to stage a board
    for the oracle
  - deferred spec posted even though staging worked -- the oracle would run twice,
    and the in-run verdict is the one that can feed custody
exit=1
```

Summary:

```
on main                            exit 1   (FileNotFoundError -- gate cannot run)
with this change                   exit 0
with the #562 regression re-added  exit 1   (both intended failures)
```

## Also: stop it recurring silently

Two paths could put this gate back to "reports success while testing nothing":

- **missing fixture** — now fails loudly with a message naming the tracked-board
  requirement, instead of a raw `FileNotFoundError` from `shutil.copy`.
- **fixture with no zone nets** — was `print("SKIP"); return None`, which `main()`
  turned into `exit 0`. A board with no zone nets never reaches the plane finalize,
  so that is a green gate testing nothing — the same silent non-coverage this
  file's own fixture note, and `tests/fixture_boards.py`, warn about. Now a
  failure.

## Blast radius

Test-only. No engine, CLI, or GUI code is touched — no flag, no default, no engine
parameter, so the "Claude-settable end to end" checklist is inert.

## Testing

```
tests/test_gui_finalize_oracle_inrun.py     FAIL on main  ->  PASS
tests/run_all.py --fast                     212 passed, 2 failed, 46 skipped
```

The 2 remaining failures are unrelated to this change:

- `test_pad_offset_keepout.py` — a real bug on `main`, fixed separately in #634
  (this branch is cut from `main`, so it still fails here).
- `test_live_board_zone_deletion.py` — Windows `cp1252` can't encode `Ω` from
  `plane_resistance.print_single_net_resistance`; passes under `python3 -X utf8`.
  It calls `create_plane` as a library, so it never gets the
  `enable_utf8_console()` that the CLI `main()`s install. Happy to file that
  separately if you want it fixed.

### DIFF
--- a/tests/test_gui_finalize_oracle_inrun.py
+++ b/tests/test_gui_finalize_oracle_inrun.py
@@ -31,9 +31,23 @@ REPO = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # Needs real zones AND at least one INCOMPLETE plane net: the finalize is a
 # plane pass, and a board whose pours are already whole exits before it ever
 # reaches the oracle leg (which is how two earlier fixture picks here made
-# the gate vacuously "pass"). This one has both GND and +3.3V pours reading
-# unconnected, and routes in seconds.
-BOARD = os.path.join(REPO, 'kicad_files', 'kit-out-plane.kicad_pcb')
+# the gate vacuously "pass").
+#
+# It must ALSO be a board a fresh clone actually has. The previous pick,
+# kit-out-plane.kicad_pcb, is an OUTPUT of tests/test_kit_route.py and was
+# untracked + gitignored by 25de9022 (#426) on 2026-07-18 -- 17 days before
+# this gate was written -- so it has never existed on any clone, and this
+# file has only ever run where a stale artifact happened to be left behind.
+# /kicad_files/kit-out*.kicad_pcb is still in .gitignore; do not point back
+# at it. Regenerating it is not an option either: it is a full route of
+# kit-dev-coldfire (minutes), which is exactly what tests/fixture_boards.py
+# declines to cover.
+#
+# lvds_converter_dualclk_gnd.kicad_pcb is TRACKED, carries a real /GND pour
+# that reads unconnected (the run's improvement gate reports the pour
+# gaining /GND, 42 -> 41 disconnected pads), and the whole gate runs in
+# under a second on it.
+BOARD = os.path.join(REPO, 'kicad_files', 'lvds_converter_dualclk_gnd.kicad_pcb')
 
 from kicad_parser import parse_kicad_pcb            # noqa: E402
 from route import batch_route                       # noqa: E402
@@ -48,12 +62,23 @@ def _run(with_callback):
         if os.path.exists(src):
             shutil.copy(src, board.replace('.kicad_pcb', ext))
 
+    if not os.path.exists(BOARD):
+        raise SystemExit(
+            f"FAIL: fixture board is missing: {BOARD}\n"
+            "  This gate needs a TRACKED board. A generated/gitignored one "
+            "makes it unrunnable on every fresh clone (see the note above).")
+
     pcb = parse_kicad_pcb(board)
     zone_nets = sorted({pcb.nets[z.net_id].name for z in (pcb.zones or [])
                         if z.net_id in pcb.nets})
     if not zone_nets:
-        print("SKIP: fixture board has no zone nets")
-        return None
+        # Deliberately NOT a skip. A fixture with no zone nets never reaches
+        # the plane finalize, so the gate would report success while testing
+        # nothing -- the silent non-coverage this file's own note warns about.
+        raise SystemExit(
+            "FAIL: fixture board has no zone nets, so the plane finalize -- "
+            "and the oracle leg this gate exists to check -- never runs. "
+            "Pick a fixture with a real, incomplete pour.")
 
     staged = []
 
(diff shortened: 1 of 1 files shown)


drandyhaas:
Landed as 09e12652 — your fixture choice adopted, with one fix. Thanks, and good catch on the 17-day gap between #426 untracking the board and this gate being written; that dating is what makes it clear the gate never guarded anything.

**Your conclusion was right, but not for the stated reason.** The PR argues regenerating `kit-out-plane.kicad_pcb` is off the table because it is "a full route of kit-dev-coldfire (minutes)". That is not what regenerating it requires — pouring planes onto the *unrouted* tracked root takes ~1.5s with no signal routing, which is what I had landed a few hours earlier in 4364d948 (we were fixing this in parallel; apologies for the collision).

The tracked board still wins, for a different reason: **a generated fixture can drift.** This gate needs a pour that reads INCOMPLETE. The day the engine gets good enough to leave that pour whole, the finalize exits before the oracle leg and the gate goes green while testing nothing — exactly the failure mode this file documents ("two earlier fixture picks made the gate vacuously pass"). A static tracked board cannot drift. So `lvds_converter_dualclk_gnd.kicad_pcb` it is.

**One defect fixed on the way in: your missing-board guard was dead code.** It sits after `shutil.copy(BOARD, board)`, which raises `FileNotFoundError` first, so the caller still gets the raw traceback the message exists to replace. Verified on your branch by pointing `BOARD` at a nonexistent file — still a bare traceback. Moved above the copy, where it now reports properly.

**Kept as-is:** no-zone-nets is a failure rather than `print("SKIP"); return None`, which `main()` turned into exit 0. That was a genuine silent-non-coverage hole.

I also reproduced your non-vacuousness check rather than taking it on trust — forcing the `stage_board_fn` branch off in `route.py` makes the gate fail with both intended messages, on the landed version. And I dropped the now-unused `kit-out-plane` recipe from `fixture_boards.py`, since with no consumer it was just rebuilt on every clean run.

Final state: gate passes in ~1s, missing board reports the real message, the #562 regression is caught with both messages, `run_all --fast` 215 passed / 0 failed.

Both loose ends you flagged are already fixed too: `test_pad_offset_keepout.py` via #634 (merged), and the `Ω`/cp1252 one in 4364d948 — that turned out not to be a `-X utf8` matter, since the test calls `create_plane` in-process and so never gets the CLI `main()` guard, meaning a real Windows user hits it from the shared engine. Fixed at the print site with a new `console_encoding.ascii_safe()`.

Issue #628, opened 2026-08-13
Interior milled contours are invisible to placement legality — a pad inside a milled ring passes every placement gate


edgehero:
Measured on run 19 (urchin, piled, see the run record in discussion #118): SW17 at (128.399, 61.001) rot 330 had pad copper spanning x 127.095–129.645 against an interior milled slot at x 127.127–129.667 — the pad sits inside the ring — and every placement gate graded it clean: copper-free `check_drc` 0 at the 0.2 floor, `check_assembly` buildable / blocking 0, channels 0 deficit. The parser already KNOWS about the contour (staging prints "reclassified 1 Edge.Cuts contour(s)... Kept as a milled edge for clearance") but `placement/legality.py` never consumes the classification, and `check_reachability` answered PASSABLE for the stranded pad because it does not model contour/edge clearance either. The defect was found only by the router: a single-net route probe failed at 219k iterations with zero copper competition, which cost a full routing cycle to discover what a placement gate could have said in milliseconds.

Suggested fix: thread the parse-time reclassification into `rect_blocked` / pad legality as keepouts, and into reachability's geometry. A sibling defect measured in the same run: a contour FULLY swallowed by a candidate rect is invisible to the outline gate (`rect_outside_amount` only sees edges crossing the rect boundary).

Full evidence: LEARNINGS.md finding 1 on the run19-assets branch (https://github.com/edgehero/KiCadRoutingTools/blob/run19-assets/LEARNINGS.md).

drandyhaas:
PR #636 is merged to `main` (`b7e0d52a`), and the same fix is now ported to the **placement** branch as `1bf5c312`. Keeping this open, as the PR asked — but with a correction to its scoping.

**The legality half is closed on both branches.** The swallow probe in `BoardOutlineGate.rect_blocked` / `rect_outside_amount` iterated `self.cutouts` only, so it never saw the milled `board_edge_contours` that `drop_pad_containing_cutouts` reclassifies (#505). A rect large enough to contain an entire interior ring passes both exact tests — no corner off-board, no edge within margin — so a part whose courtyard swallowed a milled slot whole passed every gate. Both probes now iterate a precomputed `_swallow_pts` covering cutouts *and* milled contours.

[...]

Issue #505, opened 2026-07-27
Corpus sets' router-introduced DRC errors


drandyhaas:
Every board in corpus sets 1-11 replayed at `dd16527` (wave
`~/Documents/kicad_stress_test/ab_main_0726a`), then **every intermediate step of
every offending board re-graded** with the reconciled grader
(`kicad_drc_compare.compare_board_data`: graded at each board's own routed floor,
**baseline-subtracted** against the unrouted input, #405). The delta between
consecutive steps is what that step introduced; the tool that produced the step is
the source. 80 boards x their full chains.

Companion to #463 (sets 11-20, per board / per type); this one adds the **step and
source** dimension.

## Corpus state at HEAD

- 165 boards, 164 chain-complete
- **nets routed 15390/15632 (98.45%)**, 116/164 boards at 100%
- **diff pairs coupled 463/623 (74.3%)** <- weakest number by far
- router-introduced kicad-cli DRC: **392** across 33 boards, **132 boards clean**
- pre-existing items subtracted: **1856** (83% of raw items were already in the inputs)

## More than half of all violations are TRANSIENT

| | count |
|---|---|
| introduced across all steps | **850** |
| survive to the board's final | **386** |
| **transient (removed by a later step)** | **464 (55%)** |

Removed by: `place_fanout_clearance` 269, `bga_fanout` 95, `route` 62,
`route_diff` 14, `qfn_fanout` 12.

`place_fanout_clearance` alone cleans up 269 - it is doing its #130 job well. This
is why a per-step count is not a per-step indictment: `orangecrab` shows
`bga_fanout+81` then `route-26`, ending at 23.

## Which step INTRODUCES the violations

| step | kicad_drc introduced | connection_width introduced |
|---|---|---|
| `bga_fanout.py` | **469** | 13 |
| `route.py` | **328** | **2459** |
| `route_diff.py` | 23 | 532 |
| `route_disconnected_planes.py` | 16 | 1558 |
| `route_planes.py` | 11 | **2320** |
| `qfn_fanout.py` | 3 | 159 |
| **total** | **850** | **7041** |

Two separate stories:

- **Clearance-class DRC comes from fanout and signal routing** (`bga_fanout` 469 +
  `route` 328 = 94%). Most of the fanout share is transient.
- **connection_width comes from the PLANE steps** (`route_planes` 2320 +
  `route_disconnected_planes` 1558) and `route` (2459). Different rule, different
[...]

Issue #291, opened 2026-07-03
Routers draw copper outside the Edge.Cuts polygon (bbox-only); pre-flight off-board pads


drandyhaas:
## Summary
`route.py`/`route_planes.py`/`route_disconnected_planes.py` do **not constrain routing to the true Edge.Cuts polygon** (only to the bounding box of placed footprints), so when a net's pads sit off-board they draw real tracks/vias far outside the board — producing large SEGMENT/VIA-BOARD-EDGE DRC. Surfaced by three corpus boards whose **source files place components outside the outline**.

## Evidence (both a router gap AND a corpus-tooling need)
- **framework_dock** (set4): 27/48 footprints ~20–50 mm below the real outline → up to **350 DRC** (12–51 mm board-edge overlaps) drawn into empty space.
- **pico_disp** (set5): 35/64 footprints off-board → **165 DRC** (all 124 SEG + 41 VIA-BOARD-EDGE).
- **apple1** (set5): 18 footprints off-board → **84 DRC** across 7 nets; copper extended past bounds chasing unreachable pads.

(These are unfinished/WIP GitHub source boards — the human references also left them unrouted.)

## Fix
1. **Router:** clip the routable region + obstacle model to the actual Edge.Cuts polygon; treat a pad outside the outline as unreachable and refuse to draw copper toward it (as it already does for *some* off-board nets).
2. **Corpus tooling:** pre-flight each board — reject/flag any with footprints outside `board_bounds`. (Done in `tests/stress/validate_candidate.py`, commit d4730c9; these 3 boards are being replaced.)

**Repro:** each board's no-LLM replay is in `~/Documents/kicad_stress_test/runs_set<N>/<board>/redo_commands.sh` (+ per-step `*.log`, `agent_narrative.md`). From the set4/5 stress run.

drandyhaas:
Fixed in e99374f — verified still reproducible at HEAD first, then fixed.

[...]

Issue #456, opened 2026-07-21
placement: clearance model ignores board side and outline; courtyard extraction bleeds into neighboring elements


edgehero:
Three related gaps in the quench's geometric model. All three make candidate_valid / halo decisions wrong on real boards, and the fanout-clearance engine already solved two of them in the same package. Quench line refs are against 17d4670 (PR #455 head); parser/fanout refs match main.

1. No board-side awareness. `_Part` never reads `fp.layer` (placement/quench.py:114-156), and `candidate_valid` (:278) plus `_halo_pair_penalty` (:244) treat all parts as coplanar. A back-side decoupling cap that legitimately sits under a front-side BGA fails the clearance check for every candidate near its seed, so the part is effectively frozen (only candidates are validated, never the current pose), and halo pressure pushes cross-side parts apart for no reason. fanout_clearance keys courtyard/static rects by side for exactly this case (placement/fanout_clearance.py:355-358, 384-386).

2. The board is a bbox. `usable` comes from `board_bounds` (quench.py:171-177), which is an axis-aligned bbox (kicad_parser.py:216). On an L-shaped outline or a board with interior cutouts, quench will nudge parts into the notch or the hole. fanout_clearance gates candidates against the true Edge.Cuts rings via `check_drc.board_edge_geometry` (fanout_clearance.py:280-296, added for #370); the same gate can be ported.

3. Courtyard regex crosses element boundaries. The fp_line/fp_rect patterns in placement/parser.py:96-111 use DOTALL with a lazy `.*?` up to `(layer "[FB].CrtYd")`, so a match can start inside a preceding silk element and swallow the first real courtyard line. With the standard element order (silk, courtyard, fab) I get a bbox of (-2.0,-0.9)-(2.0,0.4) for a true courtyard of (-0.5,-0.4)-(0.5,0.4). Usually this only inflates the part (spurious rejections and halo cost), but a non-rectangular courtyard can lose an extreme and under-estimate, letting real overlap validate. The same function feeds the swap identity guard from #430, which can then spuriously differ for identical footprints that have different neighboring silk. While in there: courtyards drawn with fp_poly/fp_arc/fp_circle fall back to the pad bbox without any warning, and F/B courtyards are unioned into one box.

Items 1 and 2 are ports of existing in-package patterns; 3 is a parser fix plus a regression test against a stock library footprint.

edgehero:
[...]

PR #636, opened 2026-08-14
placement: the swallow probe must see milled contours too (#628, legality half)

theironchef:
Refs #628 — closes the **legality half only**. Please keep the issue open; see
"What this does not close" below.

## The bug

`BoardOutlineGate`'s two exact tests both work from a rect's four corners and four
edges. A rect large enough to **contain an entire interior ring** passes both: no
corner is off-board, and no rect edge comes within the margin of a ring edge. The
swallow probe in `rect_blocked` / `rect_outside_amount` exists precisely to close
that hole — but it iterated `self.cutouts` only, so it never saw the milled
`board_edge_contours` that `drop_pad_containing_cutouts` reclassified (#505).

Result: a part whose courtyard swallows a milled slot whole passes every placement
gate.

## Reproduce it

The new test is the reproduction. Against `main`:

```bash
git checkout main -- py_router/placement/legality.py
python3 tests/test_628_milled_contour_legality.py
```

**Before:**

```
 FAIL gate.milled carries the contour
 FAIL swallowed milled contour -> rect_blocked
 FAIL swallowed milled contour -> rect_outside_amount > 0 (0.0)
 FAIL candidate_valid refuses a pose swallowing the milled slot
 FAIL violation() scores that pose as illegal
FAILED: 5 check(s)
exit=1
```

**After:** `exit=0`, all checks pass.

The fixture is synthetic: outline 0..60 × 0..40, a milled slot 20..24 × 19..21
containing two pads (so `drop_pad_containing_cutouts` reclassifies it), and an SW17
courtyard 18..26 × 16..24 that swallows the slot. It asserts
`board_edge_contours == 1` **and** `board_cutouts == 0` up front, so it cannot pass
for the wrong reason if the fixture ever degrades into a plain cutout.

## The fix

`BoardOutlineGate.__init__` now also captures the milled contours and precomputes
one representative vertex per interior ring:

```python
self._swallow_pts = [r[0] for r in self.cutouts] + [r[0] for r in self.milled]
```

and both probes iterate that instead of `self.cutouts`. `return True` /
`amt += self.margin` are unchanged, so the two methods keep their invariant
(`violation() == 0` still implies `not rect_blocked()`).

Any vertex serves as the representative: a ring wholly inside the rect has every
vertex inside it, and a ring only partly inside is already caught by the corner and
edge tests.

### What this deliberately does NOT do

It does **not** give milled contours containment semantics ("the interior is
off-board"). `board_edge_contours` mixes a real inner **outline** (interior = board:
crkbd's nested half, bus_pirate5's inner ring) with a milled **slot** (interior =
not board), and the parser cannot tell them apart. Declaring the interior off-board
re-opens #291, where bus_pirate5 lost all 870 pads.

*"A part may not swallow a milled ring"* is true under **both** readings — which is
exactly why this fix is safe. There is a test pinning that the interior stays
on-board.

## Blast radius

Placement-only (`py_router/placement/legality.py`); no copper is touched, so DRC and
connectivity deltas are impossible by construction. No flag, no default, no engine
parameter — the "Claude-settable end to end" checklist is inert.

**Inertness on existing fixtures is measured, not asserted.** Parsing every board in
`kicad_files/`:

```
boards scanned: 27
boards with a non-empty board_edge_contours: 0
=> _swallow_pts is byte-identical to the old cutouts-only probe on all of them
```

So every existing test is provably unaffected.

## What this does not close

The issue also names `check_reachability`, `check_assembly` and
`placement/seeder.py`. Those do not exist in this repository:

```
check_reachability       : 0 hits
check_assembly           : 0 hits
placement/seeder.py      : 0 hits
```

They appear to be the reporter's own harness, so I have not touched that half —
hence "Refs", not "Fixes". You may want to split the issue.

Also worth noting: the issue's headline premise ("`legality.py` never consumes the
classification") is **stale** since #505 — the boundary-*crossing* case already
works. The live defect is the swallowed-whole sibling, which is what this fixes.

## Testing

```
tests/test_628_milled_contour_legality.py     FAIL (5 checks) on main -> PASS
tests/run_all.py --fast                       212 passed, 3 failed, 46 skipped
tests/gui_parity/test_manifest_plan_parity.py    exit 0
tests/gui_parity/test_cli_postpass_coverage.py   exit 0
```

The 3 failures are identical to the `main` baseline on this box and unrelated to
this change — `test_pad_offset_keepout.py` (real bug on main, fixed in #634),
`test_gui_finalize_oracle_inrun.py` (missing fixture, fixed in #635), and
`test_live_board_zone_deletion.py` (Windows cp1252 vs `Ω`; passes under
`python3 -X utf8`).

### DIFF
--- a/py_router/placement/legality.py
+++ b/py_router/placement/legality.py
@@ -28,6 +28,14 @@ interior cutout -- a part gets nudged into the hole. `BoardOutlineGate` measures
 against the real Edge.Cuts rings, with the three-level short-circuit that makes
 it affordable: board-level opt-out when the outline IS its bbox, a cached
 per-part reachable-disk prune, then the exact ring test.
+
+**An interior ring can be SWALLOWED WHOLE.** Both exact tests work from the
+rect's four corners and four edges, so a rect large enough to contain an entire
+interior ring passes them: no corner is off-board and no rect edge comes within
+the margin of a ring edge. The swallow probe (`_swallow_pts`) is what closes
+that hole, and it must see BOTH kinds of interior ring -- `board_cutouts` AND
+the milled `board_edge_contours` that `drop_pad_containing_cutouts`
+reclassified (#505). It saw only the former until #628.
 """
 from __future__ import annotations
 
@@ -210,6 +218,16 @@ class BoardOutlineGate:
     Caching note: a key's answer is computed from the pose and budget of the
     FIRST call, so a caller whose budget grows later must use a distinct key.
     Over-estimating `travel` is always safe (more edges, same verdict).
+
+    Three ring collections, and they mean different things:
+
+    * `cutouts` -- real holes. Also drive the on-board (inside-outline,
+      outside-cutouts) point test, so their INTERIOR is off-board.
+    * `milled` -- `board_edge_contours` (#505): Edge.Cuts geometry KiCad mills
+      along that is neither a hole nor an outer ring. They carry an EDGE
+      CLEARANCE role only; they are deliberately NOT given containment
+      semantics -- see `_swallow_pts`.
+    * `rings` -- the flat union, what the distance tests measure against.
     """
 
     def __init__(self, board_info, margin: float):
@@ -221,6 +239,27 @@ class BoardOutlineGate:
         self.rings = rings
         self.outer = outer
         self.cutouts = cutouts
+        # Milled interior contours (#505). Already inside `rings` (that is what
+        # board_edge_geometry does with them), so the distance tests see them;
+        # kept separately because the SWALLOW probe needs them too (#628).
+        self.milled = [c for c in
+                       (getattr(board_info, 'board_edge_contours', None) or [])
+                       if len(c) >= 3]
+        # One representative vertex per interior ring, for the swallow probe in
+        # rect_blocked / rect_outside_amount. Any vertex serves: a ring wholly
+        # inside the rect has every vertex inside it, and a ring only partly
+        # inside is already caught by the corner and edge tests.
+        #
+        # DELIBERATELY NOT a containment rule. A milled contour's interior is
+        # NOT declared off-board here, and must not be: board_edge_contours
+        # mixes a real inner OUTLINE (interior = board -- crkbd's nested half,
+        # bus_pirate5's 60.8x80.2mm inner ring) with a milled SLOT (interior =
+        # not board), and the parser cannot tell them apart. Calling the
+        # interior off-board re-opens #291, where bus_pirate5 lost all 870 pads
+        # and the run broke the chain. "A part may not swallow a milled ring"
+        # is true for both readings, which is exactly why it is safe.
+        self._swallow_pts = ([r[0] for r in self.cutouts]
+                             + [r[0] for r in self.milled])
         self.margin = margin
         self.bounds = getattr(board_info, 'board_bounds', None)
         self.usable = None
@@ -291,8 +330,9 @@ class BoardOutlineGate:
 
     # -- level 3: the exact tests
     def rect_blocked(self, rect, edges=None) -> bool:
-        """True when a rect leaves the real outline, enters a cutout, or comes
-        within the edge margin of either.
+        """True when a rect leaves the real outline, enters a cutout, comes
+        within the edge margin of any Edge.Cuts ring, or SWALLOWS an interior
+        ring (a cutout or a milled contour) whole.
 
         `edges` restricts the distance test to a pre-filtered edge list from
         `edges_near`; omitted, every ring edge is measured.
@@ -308,9 +348,9 @@ class BoardOutlineGate:
                 if _seg_seg_dist_coords(ax, ay, bx, by,
                                         ex1, ey1, ex2, ey2) < self.margin:
                     return True
-        # A cutout ring FULLY INSIDE the rect evades both tests above.
-        for cut in self.cutouts:
-            cx, cy = cut[0]
+        # An interior ring FULLY INSIDE the rect evades both tests above -- no
+        # corner is off-board and no rect edge comes near a ring edge (#628).
+        for (cx, cy) in self._swallow_pts:
             if x0 <= cx <= x1 and y0 <= cy <= y1:
                 return True
         return False
@@ -359,8 +399,10 @@ class BoardOutlineGate:
                     default=float('inf'))
             if d < self.margin:
                 amt += self.margin - d
-        for cut in self.cutouts:
-            cx, cy = cut[0]
+        # Swallowed interior ring (cutout or milled contour), same probe and
+        # same charge as rect_blocked's -- the two must agree on legality or
+        # `violation()==0` stops implying `not rect_blocked()` (#628).
+        for (cx, cy) in self._swallow_pts:
             if x0 <= cx <= x1 and y0 <= cy <= y1:
                 amt += self.margin
         return amt
--- a/tests/test_628_milled_contour_legality.py
+++ b/tests/test_628_milled_contour_legality.py
@@ -1,0 +1,275 @@
+#!/usr/bin/env python3
+"""Regression test for issue #628 (LIVE half) -- a candidate rect that SWALLOWS a
+milled interior contour whole was judged LEGAL.
+
+`BoardOutlineGate`'s two exact tests work from a rect's four corners and four
+edges. A rect large enough to contain an entire interior ring passes both: every
+corner is on-board (a milled contour is not a cutout, so it does not make its
+interior off-board) and no rect edge comes within the margin of a ring edge. The
+"swallowed ring" probe exists precisely for that case -- but it iterated
+`self.cutouts` only, so it never saw `board_edge_contours`, the milled Edge.Cuts
+geometry `drop_pad_containing_cutouts` reclassifies (#505). A part could be
+placed straight over a milled slot and `violation()` reported 0.0.
+
+The boundary-CROSSING case was already handled (the rect's own edges come within
+the margin of the milled ring's edges, and `board_edge_geometry` puts those rings
+in `rings`); this test pins that as a control so the fix is not mistaken for it.
+
+NOT part of the fix, deliberately: milled contours are given no CONTAINMENT
+semantics. `board_edge_contours` mixes a real inner OUTLINE (interior = board:
+crkbd, bus_pirate5) with a milled SLOT (interior = not board) and the parser
+cannot tell them apart, so declaring the interior off-board re-opens #291
+(bus_pirate5 lost all 870 pads). "A part may not swallow a milled ring" is true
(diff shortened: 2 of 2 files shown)


theironchef:
Moving this to **draft** — I found two problems with it after opening it, one of
which is a real regression. Flagging rather than quietly leaving it in your queue.

## 1. The safety argument in the PR body is factually wrong

I claimed `board_edge_contours` mixes an inner *outline* (interior = board) with a
milled *slot* (interior = not board), and that "a part may not swallow a milled ring"
is safe because it holds under **both** readings.

The second reading is unreachable. `board_edge_contours` has exactly one producer —
`drop_pad_containing_cutouts` at `py_router/kicad_parser.py:2105` — and it moves a
contour there only when the contour encloses **≥ 2 pad centres** (`n_inside >= 2`).
The function's own docstring says it: *"A real cutout encloses no pad centres, so a
'cutout' that contains >=2 of them is a mis-classified inner contour."*

So every member is an interior-**is**-board contour with parts living inside it. My
own test fixture demonstrates this — it has to place two pads inside the "slot" to
make the reclassification fire at all.

## 2. It makes the part that *owns* the relief illegal at its own seed pose

Because reclassification requires ≥2 enclosed pads, there is **always** a component
inside such a ring. When that component's courtyard is larger than the ring — a
connector over a milled relief — the patch makes it board-violating at its own
hand-placed position.

Outline 0..60 × 0..40, milled relief 20..24 × 19..21, connector `J1` at (22,20)
whose two pads are the ones inside the ring and whose 8×8 courtyard is
18..26 × 16..24. Margin 0.55:

```
J1 at its own hand-placed seed pose, courtyard (18.0, 16.0, 26.0, 24.0):
                        main      this PR
  rect_blocked          False  ->  True
  rect_outside_amount   0.0    ->  0.55
```

And the reason that is harmful rather than merely noisy — a pose that genuinely
violates board-edge clearance scores **lower** than the seed:

```
A pose 0.45mm from the board edge (required 0.55mm), (0.45, 16.0, 8.45, 24.0):
  rect_outside_amount   0.40
```

`0.40 < 0.55`, so `candidate_valid`'s unfreeze branch (`placement/quench.py:686-700`),
which accepts any overlap-free pose whose board term merely *decreases*, will happily
move `J1` from a correct hand placement to 0.45 mm from the board edge. That is the
bug class the comment at `quench.py:668-675` exists to prevent. It also regresses the
#456 scorecard, which now reports a correct hand placement as `oob_count 1`.

## What still stands

The underlying defect is real: a rect that swallows an interior ring whole passes
both exact tests, and the swallow probe genuinely does not see
`board_edge_contours`. My inertness measurement also holds — 0 of the in-repo boards
have a non-empty `board_edge_contours`, so this is latent, not live, on the corpus.

But the rule needs scoping so it cannot fire on a ring enclosing the tested part's
own pads, and `rect_blocked` currently receives a bare rect with no part identity, so
that is an API change rather than a one-liner. I did not want to guess at your
preference there.

Two options I can see, happy to implement either if you have a view:

1. Give the swallow probe the tested part's pad set and skip rings that enclose them.
2. Apply the milled-ring swallow to `rect_blocked` only (hard gate) and leave
   `rect_outside_amount` alone — which dodges the unfreeze hazard but deliberately
   breaks the documented `violation()==0 ⟺ not rect_blocked()` invariant.

Also worth noting for #628 itself: the issue's headline premise ("`legality.py` never
consumes the classification") is stale since #505 — the boundary-*crossing* case
already works. The live half is only the swallowed-whole one.

theironchef:
⚠️ **This was merged out of draft — I think by accident — and the regression I
flagged above is now live on `main`.** Timeline: I converted it to draft at
16:10:36Z and posted the reproduction at 16:10:37Z; it was marked
`ready_for_review` at 17:26:06Z and merged 2 seconds later. Flagging in case the
draft state and comment weren't seen.

I've now confirmed it end-to-end against the actual merge commit rather than my
branch, driving `candidate_valid` (not just the gate primitives).

## Reproduction on `main`

Fixture: outline 0..60 × 0..40, milled relief 20..24 × 19..21, connector `J1` at
(22,20). J1's own two pads are the ones inside the ring — which is *why*
`drop_pad_containing_cutouts` reclassified it (`n_inside >= 2`) — and its 8×8
courtyard 18..26 × 16..24 contains the ring. `board_edge_clearance=0.55`.

```
### b7e0d52a  (main, #636 merged)
J1 seed violation()                        = 0.55
candidate_valid('J1', x=4.45, y=20.0)      -> True     # 0.45mm from the edge

### 15d11f9c  (its parent)
J1 seed violation()                        = 0.0
candidate_valid('J1', x=4.45, y=20.0)      -> False
```

So on `main` today: a connector sitting over its own milled relief is judged
**board-violating at its own hand-placed seed pose**, and because a genuinely
sub-clearance pose scores *lower* (0.40 < 0.55), `candidate_valid`'s unfreeze
branch (`placement/quench.py:686-700`) now **accepts** moving it to 0.45 mm from
a board edge that requires 0.55 mm. Before the merge that pose was refused.

That is the exact failure the comment at `quench.py:668-675` was written to
prevent — *"`violation()==0` did not imply the hard gate passes, and the unfreeze
branch could walk a part into a pose the ordinary gate forbids."*

Secondary: the #456 scorecard now reports a correct hand placement as
`oob_count 1`.

## Why it is latent rather than actively breaking the corpus

0 of the 27 in-repo boards parse to a non-empty `board_edge_contours`, so nothing
in the test suite or the corpus exercises it — which is also why it merged green.
It bites only on boards that trip the #505 reclassification (crkbd, bus_pirate5
shapes), and there it hits the part that *owns* the relief.

## Suggested action

A plain `git revert b7e0d52a` restores the previous behaviour and loses only a
latent, never-exercised improvement — I'd suggest that as the immediate move, and
I'm happy to open it as a PR if you'd like.

For the real fix, the rule needs scoping so it cannot fire on a ring enclosing the
tested part's own pads. `rect_blocked` currently takes a bare rect with no part
identity, so that is a signature/API change rather than a one-liner, which is why
I stopped rather than guessing. Two shapes:

1. Give the swallow probe the tested part's pad set (or a precomputed per-part
   exclusion list) and skip rings enclosing them.
2. Apply the milled-ring swallow in `rect_blocked` only and leave
   `rect_outside_amount` alone — dodges the unfreeze hazard, but deliberately
   breaks the documented `violation()==0 ⟺ not rect_blocked()` invariant.

Tell me which you prefer and I'll build it. Sorry for the noise — I should have
closed this rather than drafting it.

theironchef:
Fix is up as #637 — scopes the milled-ring swallow rule to non-owners, so a part is exempt only on rings its own pads sit in.

`J1` seed `violation()` back to 0.0 and `candidate_valid('J1', 4.45, 20.0)` back to False, with #636's guarantee intact for non-owners (`tests/test_628_milled_contour_legality.py` passes **unmodified**). Full suite 217 passed / 0 failed; both wx-free parity gates plus `test_settings_roundtrip` green.

That's an alternative to reverting `b7e0d52a` — your call which you prefer. The `placement` port of the same regression (`1bf5c312`) is still outstanding either way; I've confirmed it's identical there and can port whichever you pick.

edgehero:
Flagging that the regression @theironchef reported above is **also live on the `placement` branch**, and that it is now fixed there by scoping rather than reverting — in case this thread is the only place it is being tracked.

## It rode along in the port

`1bf5c312` ported this verbatim into `py_placer/placement/legality.py` (the placement branch's own 1352-line copy), including the "holds under BOTH readings" safety argument that had already been retracted in this thread. So the branch where placement work actually happens carries the same defect, and its port commit records the withdrawn rationale as justification.

## Reproduced on `placement`, end to end

Independent fixture, deliberately the realistic shape: a connector `J1` with a **milled relief in its middle** — two pads inside it (which is *why* `drop_pad_containing_cutouts` reclassifies the ring, `n_inside >= 2`) and four pins around it, so both its courtyard and its pad extent swallow the ring.

Measured at `1bf5c312`, `board_edge_clearance = 0.55`:

```
violation('J1')  at its own hand-placed pose      = 0.55     (should be 0)
candidate_valid('J1', 0.45mm from the board edge) = True     <-- the harm
legality_metrics()['oob_count']                   = 1
grade_pad_legality()['oob_pad_count']             = 1   refs [['J1', 0.55]]
```

That confirms the walk end-to-end through `candidate_valid`, not just the gate primitives: the seed scores 0.55, the sub-clearance edge pose scores 0.40, and the unfreeze branch accepts any overlap-free pose whose board term merely decreases.

**One addition to the analysis in this thread:** there is a *second*, independent scorecard path. `grade_pad_legality` grades the **pad extent**, not the courtyard, and it reaches the same false verdict whenever the owner has pins *around* its relief — which feeds `place_seed` / `place_optimize` / `place_reconstruct` / `check_assembly`, not just the #456 scorecard. A fix that only scopes the quench path leaves that one reporting phantom `oob_pad_count`.

## The fix taken (option 1, scoped)

- `BoardOutlineGate._swallow_pts` gains ring identity (it was a flat point list, which is why scoping needed an API change at all).
- `rings_enclosing(points) -> frozenset` names the **milled** rings enclosing a part's own pad centres, using `check_drc._point_in_poly` — the same containment test the parser used to reclassify the ring.
- Both probes take `skip_rings=None`; every existing call site is source-compatible.
- Ownership is `>= 1` enclosed pad (two parts with one pad each are both partial owners), computed **from the seed pose and cached** — candidate-pose ownership would let any part claim a ring by moving onto it.
- Threaded into the four sites that charge at a non-zero margin: `violation_parts`, `candidate_valid`, `legality_metrics`, `grade_pad_legality`. The margin-0 gates are numerically inert (the probe charges `+= self.margin`), and `labels.py`'s silk-label box deliberately does **not** inherit a part's exemption.

Only the `milled` half is ever skippable, which keeps the genuine-cutout half of the probe safe *by construction*: `drop_pad_containing_cutouts` moves every pad-enclosing ring out of `board_cutouts`, so no cutout id can ever land in a skip set.

After: `violation` 0.0, `candidate_valid` refuses the edge walk, both scorecards 0 — and the control part that owns nothing is still refused over the ring, which is the anti-over-scoping check.

Fix is up as #638 (onto `placement`), with the four measurements above as regression tests and a control part that owns nothing, which must stay refused over the ring.

## `main` still has it

The fix above is against the placement copy. `py_router/placement/legality.py` on `main` carries the same code and the same defect; happy to send the identical patch there if you want it.

## For #628 itself

PR #636 reported `check_reachability`, `check_assembly` and `placement/seeder.py` as "not in this repository" and inferred they were the reporter's harness. That is a `main`-only view — all three exist on `placement` (`py_tools/check_assembly.py`, `py_tools/check_reachability.py`, `py_placer/placement/seeder.py`). The rest of #628 should be assessed against that branch.

Issue #263, opened 2026-07-01
route_disconnected_planes: forced-via re-check hangs on daisho pad U1.W18 (pure-Python min/max loop, ~20min @ 99% CPU)


drandyhaas:
## Summary
`route_disconnected_planes.py`'s **forced-via re-check** stage hangs on daisho pad **U1.W18 (GND, F.Cu)** — ~20+ min at 98–99% CPU, deterministic and reproducible across runs. The board's normal tap phase and region-connection all complete in ~4 min; the hang is purely in the last-resort forced-via re-check.

## Where
The fill-aware re-check (`route_disconnected_planes.py` ~L578–606) forces a real via for any pad still floating:
```python
result = tap_pad_with_escalation(
    pad, pad_layer, net_id, pcb_data,
    replace(tap_config, via_size=vtry, via_drill=dtry),
    max_search_radius=max_search_radius, via_size=vtry,
    via_drill=dtry, verbose=verbose, fine_for_all=True,
    distant_trace_radius=0.0, disable_reuse=True)   # <- hangs here for U1.W18
```
The **same pad U1.W18 succeeds in the normal tap phase** ("reused via at (132.50, 83.50), 53 trace segment(s)"). Only the forced-via re-check (`disable_reuse=True`, `distant_trace_radius=0.0`, full `max_search_radius`) hangs — so it goes straight to `find_via_position` (multi-source, 397399c) + obstacle-map builds on a ~13 mm window at grid 0.05.

## Evidence
- `ps`: 99.3% CPU, 20+ min on one pad; A\*-call counter frozen (stuck **before/between** `route_with_frontier` calls, i.e. in Python, not the Rust router).
- `sample`: a **single deep `_PyEval_EvalFrameDefault` frame** in a tight loop of `builtin_max` / `builtin_min` / `float_richcompare` — a pure-Python min/max-over-floats loop, not numpy/Rust.
- `lldb`: confirms one deep Python frame (no nested Python calls) — the hot function's own loop.

## Likely regression
[...]

PR #642, opened 2026-08-14
#162: resolve same-net self-crossings in the shared cleanup pipeline (delete-only)

### LINKED ISSUE #162: Same-net redundant copper (self-crossings + stacked duplicates): find the producers, remove delete-only
> **#606 was combined into this issue on 2026-09-05.** Both were the same
> defect class — same-net redundant copper that KiCad's DRC permits and nothing
> removes — and both needed the same delivery vehicle. #606's content is carried
> below in full; its two comments (the built-but-inert trim pass, and the cleanup
> perf profile) remain readable on that issue and their load-bearing conclusions
> are restated here.

## The class

Same-net copper that is redundant: no electrical effect, invisible to KiCad's
DRC (which permits same-net overlap), but real material on the board and extra
segments in every downstream file. It comes in **three populations**, and the
first thing to know is that **the two detectors we have are disjoint by
construction** — neither finds the other's population, and one of the three is
detected by nothing at all.

| # | population | detector today | status |
|---|---|---|---|
| **A** | **self-crossing** — a net's own copper crossing itself transversally (an X) | `check_drc.segments_cross` (`py_router/check_drc.py:333`), spatially-indexed same-net sweep | detected, **447 on 51/74 corpus boards** |
| **B** | **exact stacked duplicate** — identical endpoints, layer, net | `check_weird._check_stacked` (`py_router/check_weird.py:478`), O(n) hash bucket on `(layer, rounded endpoint pair)` | detected, disclosed in `JSON_SUMMARY.stacked_copper`, 10 of 99 boards |
| **C** | **partial collinear overlap** — parallel, overlapping, *not* identical endpoints | **nothing** | undetected; see below |

**Why A and B cannot see each other** (verified by reading both, 2026-09-05):
`segments_cross` returns `False, None` for parallel segments
(`abs(cross) < 1e-10`, line 362) **and** for segments sharing an endpoint
(line 351). An exact duplicate is therefore doubly excluded from the A detector.
Conversely a transversal crossing pair has different endpoints, so it never
shares a bucket in B's hash key. A single acceptance criterion cannot cover
both — hence the per-population acceptance at the bottom.

**Population C is the gap this merge exposes.** #606's literal request was to
"merge collinear overlaps, drop fully-contained duplicates". Only the second
half is detected. A partial collinear overlap is parallel (invisible to A) and
has different endpoints (invisible to B). Closing A and B does not close C, and
C is the one that needs new detection rather than new removal.

---

### Origin (why FSI is not the answer)

`fix_self_intersections` (and its `collapse_appendices` shell) was **removed** —
it was the confirmed source of #159's off-grid long non-orthonormal diagonals
(it "fixed" a self-crossing by *extending* a segment to a far off-grid endpoint,
creating a tilted track that crossed foreign copper). A/B on the set-3 corpus
showed removing it is net-positive:

| keks_planes_repaired | with FSI | without FSI |
|---|---|---|
| long non-ortho (`check_orthonormal`) | 37 | 24 |
| real DRC (`check_drc -c 0.1 --clearance-margin 0.1`) | 5 | 2 |
| connectivity issues | 35 | 34 |
| **same-net self-crossings** | **3** | **6** |
| `/SRAM_A6` 232um via-cross (#159) | present | gone |

(tigard: identical/neutral either way.) So FSI traded ~3 harmless cosmetic
self-crossings for 3 real cross-net DRC violations + 13 off-grid diagonals — a
bad deal. Removed. **Any fix here must therefore be delete-only and must never
synthesize new geometry.**

### Current scale (re-measured 2026-09-05, supersedes the keks-era numbers)

Still live, and **trending worse**, not better:

- Recorded runs, sets 1–5 (era): 362 on 50/75 boards (+7 via-via).
- `cloud_final3` arm (`10982fc6`, default knobs): **447 on 51/74 boards**, 0 via-via.
- Paired on the 74 shared boards: **354 → 447**; 11 better, 24 worse, 39 same.
  zynq 45→81, orangecrab 2→19, glasgow 5→18, ottercast 9→18; crkbd 25→18, keks 19→3.
- `pourweld` / `slivertrim` arms (`d0278273`): identical on the 72 shared boards
  (310/310/313) — deterministic, unmoved by those knobs.

### Producers (evidence, not hypothesis)

- **Stage.** Every crossing appears in the main `route.py` **signal step**.
  Per-step grade of recorded chains (dilemma / ulx3s / zynq / keks): planes,
  fanout and diff pairs all score 0. ulx3s: 153 at the signal step, 56 after the
  retry step's rip-up.
- **By net.** 328/447 power (GND/RFGND/VCC, multipoint), 86 signal (64
  multipoint), 33 diff-pair members.
- **Producer 1 — the route landing rule.** For the signal class (dilemma
  `SCL_L` / `M_COL3_L` / `/SCLK_L`, all F.Cu, w 0.16 × 0.15): the route's
  terminal leg **rides along the net's pre-existing fanout stub** (nearly
  collinear, ~30–70 µm off-axis) and terminates at the stub's FAR endpoint, so
  it crosses the stub's kink. Redundant overlap, ~0.8 mm long.
- **Producer 2 — region-join straps.** 104 crossings involve a ≥0.4 mm segment:
  the in-run **plane finalize's region-join straps** crossing the net's own
  power routing (dilemma `+3V3_L`: 0.4 strap × 0.3 route, 5 crossings, identical
  coordinates era vs HEAD).

### Severity

min-trim (the smaller tail — how far to snap the nearer endpoint to remove the X):
207 crossings < 50 µm, 166 < 150 µm, 57 < 0.5 mm, **14 ≥ 0.5 mm**.
Layers: F.Cu 272, B.Cu 124, inner 51. See the severity comment below for the
per-crossing keks table.

---

### Observed

Sets 21–27 baseline wave (99 boards, 2026-08-09/10, commit `fa1e71f`).
**10 boards** shipped with same-net duplicate segments that `route.py` itself
disclosed as `stacked-copper`:

| board | count |
|---|---|
| `pocat_comms` (set24) | 11, from the final GND reconciliation pass |
| `universal_badge_led` (set21) | 10 on GND, both F.Cu and B.Cu |
| `cm4_edge_ai` (set21) | 6 (1 GND, 5 `+5V`) |
| `granit_cm5` (set23) | 5 on `+5VP`/F.Cu |
| `peaksat_comms` (set22) | 3 (GND B.Cu, GND F.Cu, `/RF/uhf-frontend/5V_RX_U` B.Cu) |
| +5 more | |

The common origin the boards report is repeated reconciliation / rescue /
plane-finalize passes across multiple `route.py` invocations laying copper over
copper that is already there. **These numbers have not been re-measured since
2026-08-10** — and population A's re-measure moved 354→447 over a comparable
window, so treat the table as indicative, not current.

### This half is disclosure working as designed

`CLAUDE.md` documents the contract: the writer drops exact via re-emissions
before writing, and whatever still stacks is *named in the summary rather than
shipped silently*. The reporting works. The gap is that nothing then removes it.

### The trim pass was already built once, and was inert (#606, @theironchef)

The natural shape — a segment twin of `route.py`'s VIA DEDUP, keyed on
`(net_id, layer, min(a,b), max(a,b), width)`, exact duplicates only — works
mechanically and survives review, with four guards: width must be in the key
because `remove_segments_from_content`'s match key has none; `graphic=True`
skipped per #337; uuid-sharing arc-derived segments skipped because the writer
only scans `(segment` blocks; `locked` is the keeper per #521.

**It is inert on the boards this issue names**, for two reasons that constrain
the fix:

1. **Scope.** Hooked into `batch_route` beside the via dedup, the trim inherits
   `strip_scope_net_ids` = the run's `--nets`. The duplicates above are
   overwhelmingly `/GND`, `+5V`, `+5VP` — nets a later signal step never
   carries. Reproduced end-to-end: duplicate an exact `/GND` block, run
   `route.py … --nets /CLK /CLK+ /CLK-` → the WARNING still prints, the
   `stacked_copper` key is still in `JSON_SUMMARY`, and both blocks ship.
2. **Ordering.** A trim at the via-dedup site (`route.py` ~:2648) runs *before*
   the plane finalize (~:3336) and the final reconciliation (~:4003) — the very
   passes named as the origin. Nothing re-censuses after them, so the copper is
   created after the trim already ran.

**This is the same conclusion population A reached independently** (producer 2
is the plane finalize), which is the main reason the two issues are now one:
they demand the same position in the pipeline.

---

## The shared vehicle

One delete-only **same-net redundancy trim**, positioned **after the plane
finalize and the final reconciliation**, operating on **the nets those passes
touched — not `--nets`**. Either (a) re-run at the end of those passes, or (b) a
standalone post-pass over the written board; (b) is simpler and testable against
the boards named above, but it is a new CLI post-pass, so per `CLAUDE.md` it
needs a GUI counterpart registered in `tests/gui_parity/test_cli_postpass_coverage.py`.

Combining buys exactly this: **one** placement decision, **one** GUI
counterpart, **one** post-pass registration, **one** summary census key, and
**one** mop-up of the freed arm (`sweep_dead_ends` / `prune_redundant_cycles`)
— instead of two of each. Detection stays split, because the populations are.

### Perf constraint (#606, @edgehero; measurements in #786)

Cleanup is already **71% of a route's wall time** — `run_post_route_cleanup`
37.3 s of `batch_route`'s 52.2 s on splitflap_driver, against 3.2 s (6%) for the
actual routing. The existing passes bottom out in scalar Python geometry
(`_pt_seg` 2,300,968 calls, `geometry_utils.union` 1,679,580, `_seg_seg_min_dist`
575,242 on that one board). Fold into a pass that already walks the same-net
segment list rather than adding an independent sweep.

**One correction to that note now that the issues are merged:** it describes the
duplicate trim as "an all-pairs same-net test" that numpy would vectorise well.
That is not true of population **B** — `_check_stacked` is an O(n) hash bucket,
already linear, and vectorising it would buy nothing. The all-pairs shape is
population **A**'s, and that sweep already runs inside `check_drc`. Apply the
vectorisation advice to A, not B.

---

## Acceptance — per population, because the risk tiers differ

**B (exact duplicates) — provably neutral, can default on.**
Deleting one of two identical segments leaves the board's geometry
bit-identical, so the trim is connectivity- and DRC-neutral *by construction*
and zeroes the reported findings by construction. Scope it to exact duplicates
explicitly. Acceptance: the 10 named boards ship `stacked_copper` absent from
`JSON_SUMMARY`, with `check_connected` counts unchanged.

**A (self-crossings) — real geometry change, opt-in until a corpus A/B clears it.**
A snap-to-crossing trim *moves an endpoint*. Acceptance:
- Identify the producing stage with evidence — **done**, recorded above.
- A delete-only resolver (or a `prune_redundant_cycles` extension) that drives
  the count toward 0 **without** adding any non-orthonormal or long segment —
  verify `check_orthonormal` stays **flat** — and with no connectivity
  regression. Candidate: uniform snap-to-crossing trim of the shorter tail,
  with `prune_redundant_cycles` / `sweep_dead_ends` mopping up the freed arm.
  No new geometry, no clearance check.
- Graded by a **corpus A/B**, not a single replay pair.

**A's upstream half — separate work, and where the 447 actually goes.**
The trim is mop-up. The two producers are the fix: (1) the route landing rule
when the target is a same-net stub (stop at first contact / trim the leg back to
the stub), (2) region-join straps in the plane finalize crossing same-net routed
copper. This half has no counterpart in B and should not be folded into the
shared pass.

**C (partial collinear overlap) — needs a detector first.**
Not covered by A's or B's acceptance. Either build the detector and extend the
trim, or split C out to its own issue once A and B land. Do not let the merge
imply it is covered.

theironchef:
Closes #162.

`fix_self_intersections` was deleted in d37a0074 (#159) because it "fixed" a
same-net X by *extending* a segment to a far off-grid endpoint, manufacturing
long non-orthonormal diagonals across foreign copper. Nothing replaced it, so
today **no cleanup pass resolves a same-net self-crossing** and `check_drc`
only warns. This adds a **delete-only** pass that resolves the ones that can be
resolved without moving a single endpoint.

## Root cause, exactly

`_prune_net_cycles` (`py_router/pcb_modification.py`) builds its node set from
segment **endpoint clusters** (`rep_items`). A pure X has *no endpoint at the
intersection*, so no node lies there and the redundant loop the crossing closes
is invisible to the union-find tree invariant. Every other subtractive pass is
whole-segment / endpoint-graph based too.

Measured in isolation on the two in-repo boards that carry the defect — the
cycle prune is **blind to the shape, not idle**:

```
routed_output.kicad_pcb                    prune_redundant_cycles     removed   1 seg(s); crossings 2 -> 2
routed_output.kicad_pcb                    collapse_strict_redundant  removed  18 seg(s); crossings 2 -> 1
      resolved (106, 'In1.Cu', 217.6, 107.8)
rp2350_fpga_eensy_prePlane.kicad_pcb       prune_redundant_cycles     removed   2 seg(s); crossings 4 -> 4
rp2350_fpga_eensy_prePlane.kicad_pcb       collapse_strict_redundant  removed   2 seg(s); crossings 4 -> 4
```

Note the second line: on `routed_output` the existing **strict collapse already
takes the arm this pass would take**. That fact shapes the whole evidence
section below.

## What this PR does

Adds `pcb_modification.prune_self_crossing_segments`, called from the **shared**
`cleanup_pipeline.run_post_route_cleanup` at step 7b, immediately after the
cycle prune (its blind spot) — so **both fronts (CLI and GUI) get it**. No flag,
no new engine parameter, so no `FLAG_PARAMS` / dialog-control plumbing and
neither parity gate's registry changes.

**Delete-only. No endpoint ever moves.** For each crossing the shorter arm is
tried first, and a removal is accepted only if every guard the cycle prune
already uses still passes:

* connectivity still connected, `num_components` not increased,
  `disconnected_pads` not increased;
* no via loses its support (the #209 gate);
* `_restore_soft_joint_bridges` (#319) does not put the segment back.

Copper graphics (#337) are never candidates, and a **whole net** is skipped when

* it is zoned (planes are meshes, not trees);
* **any** of its segments or vias is KiCad-locked — #521 is a *net*-level rule
  ("never-rippable with NO override"), and the lock need not be on the arm being
  tried;
* it is net 0, or has fewer than two pads — `check_net_connectivity` calls a
  pad-less net *connected* with `num_components` 0, so every gate above would
  pass vacuously.

**A crossing whose both arms are load-bearing is left alone.** I deliberately
did **not** implement the "snap the shorter tail back to the crossing point"
trim: that is #159's failure mode one step down — a `segment-endpoint-gap` above
`connectivity.COINCIDENCE_TOL` is a *counted* DRC violation in `check_drc`, so
the trim would clear a KiCad-permitted warning by creating a real one.

### Scope is not a protection list

The pipeline hands this pass the same `_sub_scope` as the cycle prune and the
strict collapse, which subtracts **only** `protect_net_ids` (route.py's
`_protect_unfinished`). #521-protected matched groups, impedance nets and diff
pairs **are** in scope — `route_diff.py` calls the pipeline with the diff-pair
nets *as* the scope. That is not new exposure. On **`main`**, with a net
registered protected *and* impedance-declared and passed as the entire scope:

```
protection_map / notes see it as protected: True
segments 4 -> 2; counts={'cycles_pruned': 1, 'strict_collapsed': 1}
   Cycle prune: removed 1 redundant loop segment(s) across 1 net(s)
   Strict collapse: removed 1 redundant segment(s) (superseded parallel/buried copper)
```

Identical on this branch. The docstrings now say this plainly instead of
claiming an exclusion that never existed.

## Before / after — measured through the path that actually ships

Production never calls the pass on a raw board; it calls
`run_post_route_cleanup`, where six passes run first and four run after. So the
A/B below is the **same pipeline with the pass stubbed out vs live**, on a fresh
parse each leg, graded with `check_drc.py` at 0.09 and `check_connected.py`:

| board | same-net self-crossing (check_drc) | other DRC | check_connected |
|---|---|---|---|
| `routed_output.kicad_pcb` | **1 → 1** | `VIA-DRILL-HOLE (3)` → `VIA-DRILL-HOLE (3)` | 188 issues, **report byte-identical** |
| `rp2350_fpga_eensy_prePlane.kicad_pcb` | **6 → 4** | `NO DRC VIOLATIONS FOUND!` → `NO DRC VIOLATIONS FOUND!` | 2 issues, **report byte-identical** |

```
$ diff <(check_connected stubbed | tail -n +3) <(check_connected live | tail -n +3)   # both boards
DIFF_EXIT=0
```

Two things in that table need saying out loud rather than glossing:

* **On `routed_output` this pass changes nothing that ships.** The shipped
  copper multiset (segments *and* vias) hashes identically with and without it —
  `47466f7887b2747d` both legs — because `collapse_strict_redundant` removes the
  same arm on its own (`strict_collapsed` just moves 17 → 16). It fires
  (`self_crossings_pruned: 1`) but the board is the same board.
* **The 6 on `rp2350` is not a typo.** The raw board has 4. `close_soft_joints`
  (step 10) *manufactures* crossings with the bridge copper it adds, on `main`
  and on this branch alike — a stage trace hooked to the pipeline's own
  `progress_callback` puts every added crossing after the last callback. This
  pass runs at 7b and cannot see them. So the shipped board comes back to its
  input count of 4; it does not go below it. That is a separate defect and it is
  filed as such, not fixed here.

**Real coverage: 2 of the 6 in-repo crossings, on 1 of the 2 boards.** On the
other four (`/RP2354A/1V1`, `Net-(U6-VREG_AVDD)`, `Net-(U2A-~{TXE})`, and the
strict-collapse-reachable one) *both* arms are load-bearing bridges, or another
pass gets there first.

`rp2350` ships **3** fewer segments, not 2: the two arms plus one dead end the
sweep then trims (`dead_ends_swept` 0 → 1 on the live leg only).

### Reproduce (copy-paste, tracked in-repo board, no corpus, no route run)

```bash
python3 py_router/check_drc.py kicad_files/rp2350_fpga_eensy_prePlane.kicad_pcb -c 0.09 | tail -7

python3 - /tmp/rp2350_trimmed.kicad_pcb <<'PY'
import sys
sys.path[:0] = ['.', 'py_router', 'py_tools', 'rust_router']
from kicad_parser import parse_kicad_pcb
from kicad_writer import remove_segments_from_content
from pcb_modification import prune_self_crossing_segments
src, dst = 'kicad_files/rp2350_fpga_eensy_prePlane.kicad_pcb', sys.argv[1]
pcb = parse_kicad_pcb(src)
n, nets, strip = prune_self_crossing_segments([], pcb, None)
print(f'prune_self_crossing_segments: removed {n} segment(s) on {nets} net(s)')
for s in strip:
    print(f'  {pcb.nets[s.net_id].name}  {s.layer}  '
          f'({s.start_x},{s.start_y})->({s.end_x},{s.end_y})  w={s.width}')
txt = open(src, encoding='utf-8').read()
txt, k = remove_segments_from_content(
    txt, strip, net_id_to_name={i: v.name for i, v in pcb.nets.items()})
open(dst, 'w', encoding='utf-8', newline='').write(txt)
PY

python3 py_router/check_drc.py /tmp/rp2350_trimmed.kicad_pcb -c 0.09 | tail -7
python3 py_router/check_connected.py /tmp/rp2350_trimmed.kicad_pcb | tail -8
```

Real captured output:

```
NO DRC VIOLATIONS FOUND!
WARNINGS (4, not DRC failures):
  same-net self-crossing: 4 (same-net copper overlap; permitted by KiCad DRC)

prune_self_crossing_segments: removed 2 segment(s) on 2 net(s)
  /RP2354A/SWDIO  F.Cu  (149.65,111.15)->(149.7,111.2)  w=0.09
  /RP2354A/SCL  F.Cu  (153.2,104.9)->(153.1,104.9)  w=0.09

NO DRC VIOLATIONS FOUND!
WARNINGS (2, not DRC failures):
  same-net self-crossing: 2 (same-net copper overlap; permitted by KiCad DRC)

FOUND 2 ISSUES:
  Unrouted nets (2):
    +3V3 (48 pads)
    GND (87 pads)
```

The two `Unrouted nets` are the board's pre-existing un-poured planes (it is a
`prePlane` fixture) — identical before and after, and the full
`check_connected.py` report is byte-identical (`DIFF_EXIT=0`).

## Inertness — the default path is unchanged where there is no crossing

There is no flag, so "default path unchanged" has to mean *byte-identical
output*. Three independent measurements:

**1. All 33 tracked `kicad_files/*.kicad_pcb` through the pipeline, base vs
branch** (17 have copper):

```
BYTE-IDENTICAL shipped copper AND identical other-pass counts: 15
CHANGED: 2
   routed_output.kicad_pcb: segs 517->517, self_crossings_pruned=1
        strict_collapsed: 17 -> 16          (copper hash still identical)
   rp2350_fpga_eensy_prePlane.kicad_pcb: segs 903->900, self_crossings_pruned=2
        dead_ends_swept: 0 -> 1
boards that errored (base, fix): 0
cycles_pruned identical on every board: True
```

**2. A real end-to-end `route.py` of `splitflap_driver.kicad_pcb`**, base and
branch, to fresh output paths, comparing the UUID-free copper multiset (never
file hashes — CLAUDE.md):

```
A: sf_base  segments=1437 vias=168
B: sf_fix   segments=1437 vias=168
segment multiset identical: True   (A-only=0, B-only=0)
via     multiset identical: True   (A-only=0, B-only=0)
```

Every other pass line is character-identical (`Cycle prune: removed 149 ...`,
`Strict collapse: removed 11 ...`, `Octolinear smoothing: 83 span(s)`,
`Bridged 2 soft joint(s)`), and the new pass prints nothing.

**3. Strict object-level no-op** pinned in the test on four crossing-free
tracked boards: nothing removed, `nets_pruned == 0`, `strip == []`, and
`id(pcb_data.segments)` unchanged — the list object is not even reassigned.

**Determinism**: one distinct removal set across six shuffled segment/via
orderings, on both boards.

## Cost

One connectivity graph per crossing-carrying net (`return_graph=True`), then
`analyze_conn_excluding` per candidate (#263: O(net_size + candidates) instead
of O(net_size × candidates)), an O(vias)-per-query index twin of the #209
via-support model, and the #319 guard batched per net exactly as
`collapse_strict_redundant` batches it. Back-to-back on one machine, a naive
per-candidate rebuild vs what ships:

| segments | crossings | per-candidate rebuild | shipped |
|---|---|---|---|
| 516 | 4 | 0.40 s | **0.06 s** |
| 1516 | 4 | 0.57 s | **0.23 s** |
| 3016 | 4 | 1.03 s | **0.41 s** |
| 1660 | 40 | 5.95 s | **0.49 s** |
| 2100 | 150 | 32.44 s | **1.63 s** |

`analyze_conn_excluding` replays a graph built from the *full* segment list, so
it can diverge from a true recompute if a removal empties a copper layer (its
documented caveat). Every net that actually loses copper therefore gets **one
authoritative `check_net_connectivity` recompute** at the end; if it disagrees
with the graph model, every removal on that net is reverted. Nothing ships in a
state the real checker has not graded.

The #209 via-support model is now a single shared helper (`_via_support_model`)
used by both `_prune_net_cycles` and this pass instead of two copies —
`cycles_pruned` is unchanged on every tracked board.

## Where the crossings come from (issue acceptance item 1)

Measured, not speculated: on `rp2350_fpga_eensy_prePlane` **all 4** crossings
have one arm at width exactly `0.0889` — `single_ended_routing._fab_track_floor`
for a 6-layer board, i.e. a terminal connector necked by `_neck_terminal_grazes`
— even though only 59 of the board's 904 segments (6.5 %) carry that width. The
matching code path is the `_merge_terminal_to_exact` bailout at
`py_router/single_ended_routing.py:749`. I have the 4/4 correlation and the code
path, **not** a captured `Segment.__init__` stack — producing one needs
regenerating the fanout chain. Flagging that as unfinished rather than claiming
it.

## Tests

`tests/test_162_self_crossing_prune.py` (standalone, no wx, picked up by
`run_all.py --fast`). 48 checks:

* **POS / NEG / SJ** synthetic fixtures — the redundant-link-crossed-by-a-
  connector shape measured on the real boards; an X whose arms are both sole
  bridges (nothing removed, and it survives the *whole* pipeline); and the
  #319 fixture where the short arm is connectivity-safe but manufactures a soft
  joint while the long arm strands two pads (both gates asserted *in isolation*
  first, so the fixture cannot pass for the wrong reason).
* **Net-level skips** — a lock on either arm, and a locked *via*, take the whole
  net off-limits; net 0, a zero-pad net and a one-pad net are untouched at the
  documented default `scope_net_ids=None`.
* **Via-support equivalence** — the index twin matches the list original on all
  32 subsets of a fixture.
* **Pipeline wiring** — the fix arrives through `run_post_route_cleanup`, is
  counted, and the removed input segment lands in `input_strip_segments`.
* **Real boards through the shipping path** — `run_post_route_cleanup` with the
  pass stubbed vs live, asserting the pass fires, that the shipped board never
  gains a crossing, that `routed_output`'s copper is **identical** (pinned as
  identity on purpose — claiming a win there would be false), that `rp2350`'s
  crossings drop, and that every net the pass changed is no worse than without
  it (connectivity, components, disconnected pads, soft joints).
* **Inertness** — strict no-op on four crossing-free tracked boards.

Red on `main` (engine files from `main`, test file from this branch) — 8 failed,
including a genuine *behavioral* red, not only the missing symbol:

```
  FAIL  pipeline: run_post_route_cleanup resolves the crossing  crossings=1 segs=4
  FAIL  pipeline: the removal is counted  counts=None
  FAIL  pipeline: the removed input segment is in input_strip_segments  strip=0
  PASS  pipeline: net still connected, no pad stranded
  PASS  pipeline: load-bearing X survives the whole pipeline  crossings=1 segs=3
EXIT=1
```

And a **mutation red**: with the pass neutered to `return 0, 0, []`, the
real-board section goes red on *both* boards —

```
5 FAILED: routed_output.kicad_pcb: the pass fires inside the pipeline,
routed_output.kicad_pcb: DELETE-ONLY -- no survivor moved,
rp2350_fpga_eensy_prePlane.kicad_pcb: the pass fires inside the pipeline,
rp2350_fpga_eensy_prePlane.kicad_pcb: the pass is what removes the crossings,
rp2350_fpga_eensy_prePlane.kicad_pcb: DELETE-ONLY -- no survivor moved
```

A direct-call test would have stayed green through that on `routed_output`,
which is exactly why the section drives the pipeline.

## Regressions

```
$ python3 tests/run_all.py --fast -j 4 --timeout 240
217 passed, 0 failed, 46 skipped in 198.6s

$ python3 tests/gui_parity/test_manifest_plan_parity.py
Converter parity: 35 flag-checks across 1 manifest(s), 0 mismatch(es).
Param->control resolution: 12 params checked, 0 unresolved.

$ python3 tests/gui_parity/test_cli_postpass_coverage.py
CLI post-pass coverage: 8 registered, 15 in active CLI use, 1 CLI-only acknowledged,
0 unreachable call(s), 0 failure(s), 0 warning(s).
```

No parity registration needed: the pass lives inside the shared engine function,
not a CLI `main()`, and adds no flag.

`run_doc_examples.py`: 4 failures, **identical on `main`** (missing fixture
boards in `api-impedance.md` / `api-net-analysis.md`). Confirmed by reverting
only my doc file: the same 4 fail, 31 run, only the skip count moves 77 → 78.

## Known residuals

* **Coverage is 2 of 6, on 1 of 2 boards** — and on `routed_output` nothing
  ships differently. Four crossings have two load-bearing arms or are already
  reachable by another pass.
* **The pipeline still manufactures crossings after this pass runs.**
  `close_soft_joints` takes `rp2350` from 4 to 6; this pass brings it back to 4,
  not below. A second, post-pipeline run of the same pass would take it to 1
  (measured) — but that inverts the documented #319 ordering ("`close_soft_joints`
  is the LAST copper step … so every joint close sees is router-born, never one
  a cleanup pass manufactured"), which deserves its own change and its own
  grading. Filed separately.
* **The cost table is a synthetic worst case.** Its residual driver is
  `_soft_joint_pairs`' O(dangles²) pair loop, which the fixture makes
  adversarial (300 free dangles). Real boards have far fewer. 1.63 s is a
  measurement, not a bound.
* **The graph-model caveat is covered by a net-level revert, not by exactness** —
  a divergence reverts the whole net rather than the single bad candidate. None
  was observed on any in-repo board.
* **The batched #319 guard loses one fallback**: if the shorter arm fails *only*
  the soft-joint gate, the longer arm is no longer tried. Measured identical
  removals on both in-repo boards and on the SJ fixture; no in-repo case
  exercises the difference.
* **The 8 crossings tabulated in the issue are on a corpus board I cannot
  measure here**, and acceptance item 1 is a width correlation plus a read code
  path, not a captured stack.
* **The 46 `--fast`-skipped integration tests were not run.** The end-to-end
  `route.py` A/B and the 33-board pipeline sweep cover the same ground for this
  change, but a full `run_all` is worth a CI slot.
* **Unrelated pre-existing bug found while working here, deliberately not fixed
  in this PR**: `check_connected.analyze_conn_excluding` computes `copper_roots`
  twice (the block is duplicated verbatim) and never returns
  `num_copper_components`, so the island-stranding gate at
  `pcb_modification.py:1809` and `check_weird.py:393` is vacuous —
  `analyze_conn_excluding` returns only
  `['connected', 'disconnected_pads', 'num_components']`. Filed separately;
  turning a dead gate live changes how much copper `collapse_strict_redundant`
  removes and needs its own grading.

## Post-review addendum (second adversarial pass)

A second independent review re-ran everything and confirmed all prior findings
genuinely fixed, the pass never disconnects anything, and the limitations
section holds. It surfaced one new latent defect, fixed in the head commit:

- **The 1mm spatial hash indexed each segment's bounding box, not its traversed
  cells** — a net of board-spanning diagonals paid ~O(length²) cell inserts and
  measured **11× slower** than the naive pairwise scan the old docstring claimed
  it avoided (identical pairs both ways). Latent — the worst in-repo net is 130
  inserts — but the docstring claim was false. It now states the truth, and the
  scan falls back to the plain double loop when the bbox fanout exceeds the
  pairwise cost. Both paths verified equivalent over 200 random tangles plus a
  long-diagonal board (same pairs, same deterministic order).

Three corrections to numbers quoted earlier in this description, from the same
pass: the sweep statement is precisely **22 tracked boards, 11 with copper, 9
byte-identical**; the red run's real failure line is `12 FAILED:` of which the 5
named checks are the load-bearing subset; and the doc-example failure count is
environment-dependent — the claim that holds is base-parity (identical failures
on both arms), not the absolute count.

### DIFF
--- a/docs/api-pcb-modification.md
+++ b/docs/api-pcb-modification.md
@@ -95,7 +95,44 @@ the routing pipeline applies both so the model and the output file agree.
 > redundant with `sweep_dead_ends`). `add_route_to_pcb_data` now commits routed
 > segments directly; connectivity cleanup is handled by `prune_redundant_cycles`
 > and the whole-net dead-end trim `sweep_dead_ends` (#84). The residual same-net
-> self-crossings are tracked in #162.
+> self-crossings are handled by `prune_self_crossing_segments` (#162) — by
+> DELETION only, never by moving an endpoint.
+
+```python
+prune_self_crossing_segments(results, pcb_data, scope_net_ids=None,
+                             keep_input_copper=False)
+    -> (segments_removed, nets_pruned, original_segments_to_remove)
+```
+
+Resolves same-net **self-crossings** (issue #162) — two segments of one net
+crossing in an interior X, which `check_drc` reports as a
+`same-net self-crossing` warning. `prune_redundant_cycles` cannot see them:
+its node set is built from segment *endpoint* clusters, and an X has no
+endpoint at the intersection, so the loop it closes is invisible to the tree
+invariant.
+
+**Delete-only.** For each crossing the shorter arm is tried first, and a
+removal is accepted only if every guard `prune_redundant_cycles` uses still
+passes: `check_net_connectivity` still connected with no extra components and
+no extra disconnected pads, no via loses its support (#209), and
+`_restore_soft_joint_bridges` (#319) does not put the segment back. A crossing
+whose *both* arms are load-bearing is left alone — the pass never extends or
+snaps an endpoint to the crossing point, which is what `fix_self_intersections`
+did wrong (#159). Copper graphics (#337) are never candidates, and a whole net
+is skipped when it is zoned, when **any** of its copper is KiCad-locked (#521 —
+the lock need not be on the arm being tried), or when it is net 0 / has fewer
+than two pads (`check_net_connectivity` calls a pad-less net *connected*, so
+every gate would pass vacuously).
+
+`scope_net_ids` is a scope, **not** a protection list: the pipeline passes the
+same `_sub_scope` the cycle prune and the strict collapse get, which subtracts
+only the caller's `protect_net_ids`. Protected matched groups, impedance nets
+and diff pairs are in scope, exactly as they are for those two passes.
+
+Runs inside `cleanup_pipeline.run_post_route_cleanup` at step 7b, so both the
+CLI and the GUI front get it; there is no flag. Strictly inert on a net with no
+self-crossing. It runs *before* `close_soft_joints`, so a crossing that a later
+pipeline pass manufactures is out of its reach.
 
 ## `geometry_utils.py`
 
--- a/py_router/cleanup_pipeline.py
+++ b/py_router/cleanup_pipeline.py
@@ -42,6 +42,7 @@ from pcb_modification import (
     nudge_grazing_microshift,
     nudge_grazing_vias,
     prune_redundant_cycles,
+    prune_self_crossing_segments,
     collapse_strict_redundant,
     remove_orphan_islands,
     sweep_dead_ends,
@@ -118,6 +119,9 @@ def run_post_route_cleanup(results, pcb_data, scope_net_ids, config, *,
       6. nudge_grazing_vias    -- re-bend / micro-shift load-bearing grazes the
                                   prune must keep (#224/#276/#280).
       7. prune_redundant_cycles -- per-net tree invariant (RAM_A9 loops).
+      7b. prune_self_crossing_segments -- #162: the cycle prune's blind spot,
+                                  a same-net X with no endpoint at the
+                                  intersection. DELETE-ONLY, same guards.
       8. sweep_dead_ends       -- trim dead-end spurs and unsupported vias.
       9. neck_wide_segments_grazing_pads -- width-only fix for wide power
                                   trunks overlapping a fine-pitch foreign pad.
@@ -347,6 +351,31 @@ def run_post_route_cleanup(results, pcb_data, scope_net_ids, config, *,
         if _cy_segs:
             print(f"{label}Cycle prune: removed {_cy_segs} redundant loop "
                   f"segment(s) across {_cy_nets} net(s)")
+
+    # Same-net SELF-CROSSING prune (#162), right after the cycle prune because
+    # it is the cycle prune's blind spot: a pure X has no endpoint at the
+    # intersection, so the endpoint-cluster tree invariant never sees the loop
+    # it closes. DELETE-ONLY -- no endpoint ever moves (the #159 lesson: the
+    # geometric "fix" that extends/snaps a segment to the crossing trades a
+    # KiCad-permitted warning for real DRC violations).
+    #
+    # Scope: the same _sub_scope as the cycle prune and the strict collapse,
+    # which subtracts ONLY protect_net_ids (route.py's _protect_unfinished --
+    # nets with failed pads, plus unrouted single-ended nets). It does NOT
+    # exclude #521-protected matched groups, impedance nets or diff pairs:
+    # those reach the pipeline through sweep_scope_ids (name-filtered) and,
+    # from route_diff.py, as the scope itself. Same exposure as its two
+    # neighbours here -- the pass carries its own net-level guards instead
+    # (locked copper, zoned nets, pad-less nets; see its docstring).
+    _prog("self-crossing prune")
+    _xc_segs, _xc_nets, _xc_strip = prune_self_crossing_segments(
+        results, pcb_data, _sub_scope, keep_input_copper=keep_input_copper)
+    counts['self_crossings_pruned'] = _xc_segs
+    _trace('self_crossings')
+    strip.extend(_xc_strip)
+    if _xc_segs:
+        print(f"{label}Self-crossing prune: removed {_xc_segs} redundant "
+              f"same-net crossing segment(s) across {_xc_nets} net(s)")
 
     # Strict-redundant collapse (#217 classes 1-2): superseded parallel
     # chains and pad/via-buried tails that are redundant under the strict
--- a/py_router/pcb_modification.py
+++ b/py_router/pcb_modification.py
@@ -2254,6 +2254,70 @@ def neck_wide_segments_grazing_pads(results, pcb_data, config) -> int:
     return necked
 
 
+class _ViaSupport(NamedTuple):
+    supported_vias: object          # (seglist) -> set of via indices
+    supported_excluding: object     # (excluded_seg_indices) -> same, or None
+
+
+def _via_support_model(net_vias, net_pads, net_segs=None) -> _ViaSupport:
+    """The #209 via-support model shared by the subtractive per-net passes.
+
+    A via survives ``sweep_dead_ends`` only if a kept same-net segment ENDPOINT
+    lands within 0.05mm of it, or a pad covers it -- so a removal that drops a
+    via's last support would be followed by the sweep culling that via and
+    disconnecting the net. Both ``_prune_net_cycles`` and
+    ``prune_self_crossing_segments`` gate on it; this is the single copy.
+
+    ``supported_vias(seglist)`` -> set of indices into ``net_vias``.
+
+    ``supported_excluding(excluded_indices)`` is the same answer for
+    ``net_segs`` minus those indices, but costs O(vias) per query instead of
+    O(vias x segments) -- available only when ``net_segs`` is passed, and
+    exactly equal to ``supported_vias([s for i, s in enumerate(net_segs)
+    if i not in excluded])`` (pinned over all subsets of a fixture by
+    tests/test_162_self_crossing_prune.py).
+    """
+    vias = list(net_vias or [])
+    # Pad coverage points, mirroring sweep_dead_ends' via-support model exactly.
+    pad_cover = []
+    for p in (net_pads or []):
+        px = getattr(p, 'global_x', getattr(p, 'x', 0.0))
+        py = getattr(p, 'global_y', getattr(p, 'y', 0.0))
+        ps = max(getattr(p, 'size_x', 0.5), getattr(p, 'size_y', 0.5))
+        pad_cover.append((px, py, ps))
+    pad_backed = {i for i, v in enumerate(vias)
+                  if any(math.hypot(v.x - cx, v.y - cy) < cs / 2 + 0.05
+                         for cx, cy, cs in pad_cover)}
+
+    def _touches(v, sg):
+        return (math.hypot(v.x - sg.start_x, v.y - sg.start_y) < 0.05
+                or math.hypot(v.x - sg.end_x, v.y - sg.end_y) < 0.05)
+
+    def supported_vias(seglist):
+        out = set(pad_backed)
+        for i, v in enumerate(vias):
+            if i in out:
+                continue
+            if any(_touches(v, sg) for sg in seglist):
+                out.add(i)
+        return out
+
+    supported_excluding = None
+    if net_segs is not None:
+        by_via = [{k for k, sg in enumerate(net_segs) if _touches(v, sg)}
+                  for v in vias]
+
+        def supported_excluding(excluded=()):  # noqa: F811 (index-based twin)
+            ex = set(excluded)
+            out = set(pad_backed)
+            for i, sup in enumerate(by_via):
+                if i not in out and (sup - ex):
+                    out.add(i)
+            return out
+
+    return _ViaSupport(supported_vias, supported_excluding)
+
+
 def _prune_net_cycles(net_id: int, net_segs: List[Segment], net_vias, net_pads,
                       fgrid, fcell: float, fmax_rad: float, clearance: float,
                       protected_ids=None):
@@ -2419,25 +2483,8 @@ def _prune_net_cycles(net_id: int, net_segs: List[Segment], net_vias, net_pads,
         return net_segs, []
     base_comps = base.get('num_components') or 1
     base_disc = len(base.get('disconnected_pads') or [])
-    # Pad coverage points, mirroring sweep_dead_ends' via-support model exactly.
-    pad_cover = []
-    for p in (net_pads or []):
-        px = getattr(p, 'global_x', getattr(p, 'x', 0.0))
-        py = getattr(p, 'global_y', getattr(p, 'y', 0.0))
-        ps = max(getattr(p, 'size_x', 0.5), getattr(p, 'size_y', 0.5))
-        pad_cover.append((px, py, ps))
-
-    def supported_vias(seglist):
-        # A via survives sweep_dead_ends only if a kept same-net segment endpoint
-        # lands within 0.05mm of it, or a pad covers it (see sweep_dead_ends).
-        out = set()
-        for i, v in enumerate(net_vias or []):
-            if any(math.hypot(v.x - sg.start_x, v.y - sg.start_y) < 0.05 or
-                   math.hypot(v.x - sg.end_x, v.y - sg.end_y) < 0.05 for sg in seglist) \
-               or any(math.hypot(v.x - cx, v.y - cy) < cs / 2 + 0.05
-                      for cx, cy, cs in pad_cover):
-                out.add(i)
-        return out
+    # The #209 via-support model (shared with prune_self_crossing_segments).
+    supported_vias = _via_support_model(net_vias, net_pads).supported_vias
 
     cur = list(net_segs)
     cur_supported = supported_vias(cur)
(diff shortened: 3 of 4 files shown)


drandyhaas:
Thank you for this — and apologies for leaving it open for three weeks. I'm
going to close it, but **not** on the merits: the work is sound and the
diagnosis is the most useful thing anyone has written on #162. Closing because
the issue underneath it changed shape today, in ways that move where this pass
has to live.

## What the root-cause finding bought

This is the part I want on the record, because it survives the reshuffle intact:

> `_prune_net_cycles` builds its node set from segment **endpoint clusters**
> (`rep_items`). A pure X has *no endpoint at the intersection*, so no node lies
> there and the redundant loop the crossing closes is invisible to the union-find
> tree invariant. Every other subtractive pass is whole-segment / endpoint-graph
> based too.

That explains why every existing pass is blind to the shape rather than merely
idle, and the isolation measurement backing it — including the observation that
`collapse_strict_redundant` already takes the arm this pass would take on
`routed_output` — is exactly the kind of evidence the issue was missing. It is
being carried into #162 rather than lost with this PR.

## Why the issue moved

Three things, all after this PR was opened:

1. **#606 was combined into #162 today.** The issue is now *same-net redundant
   copper* across **three** populations, not self-crossings alone: transversal
   crossings (this PR), exact stacked duplicates, and partial collinear overlaps
   that no detector finds at all. `segments_cross` returns `False` for parallel
   segments and for segments sharing an endpoint, so the two detectors we have
   are disjoint by construction — one acceptance criterion can't cover them, and
   #162 now states acceptance per population.
2. **The position changed.** #606's own review established that a trim hooked
   into `batch_route` is inert on the boards that matter, because the plane
   finalize (~:3336) and final reconciliation (~:4003) create the copper
   afterwards, and it inherits `strip_scope_net_ids` = `--nets` while the copper
   sits on `/GND` and `+5V`. Independently, #162's producer attribution named
   that same plane finalize — its **region-join straps** crossing the net's own
   power routing — as the source of 104 of the crossings. Step 7b of
   `run_post_route_cleanup` is the right answer to the *cycle prune's* blind
   spot, which is what this PR set out to fix; it is not yet the right answer to
   *where the copper is produced*. The current plan is one pass positioned after
   finalize and reconciliation, scoped to the nets those passes touched.
3. **Re-measured 2026-09-05, and it got worse, with producers named.** Corpus
   sets 1–5, paired on 74 boards: **354 → 447** crossings on 51/74 boards (11
   better, 24 worse, 39 same). Every one appears in the `route.py` **signal
   step** — planes, fanout and diff pairs all grade 0. Two producers now have
   evidence: the region-join straps above, and a route landing rule where the
   terminal leg **rides along the net's pre-existing fanout stub** (~30–70 µm
   off-axis) and terminates at the stub's far endpoint. That upstream half is
   where the 447 actually goes; a delete-only trim is the mop-up behind it.

There's also a constraint from #786 worth knowing before this is rebuilt:
`run_post_route_cleanup` is already **71% of route wall time** (37.3 s of
`batch_route`'s 52.2 s on splitflap_driver, against 3.2 s of actual routing), so
the intent is to fold into a pass that already walks the same-net segment list
rather than add an independent sweep.

## What happens next

#162 stays open with the combined scope and per-population acceptance. When the
trim pass gets built it will be as part of that larger piece — the shared
post-pass plus the two producer fixes — and this PR is the reference for the
delete-only design: shorter arm first, every guard the cycle prune already uses
(connectivity, `num_components`, `disconnected_pads`, the #209 via-support gate,
`_restore_soft_joint_bridges`), #337 graphics skipped, and whole-net skips for
zoned nets and #521 locks. The `check_orthonormal`-flat requirement is now
written into #162's acceptance for exactly the reason your description opens
with.

The branch is untouched, so nothing here is lost. Sorry to spend your effort
this way, and thanks again for the diagnosis — it made the issue much better
than it was.

Issue #266, opened 2026-07-02
Add polarity (P/N pad) swap to the hybrid diff-pair path (side-flipping pairs wrap around instead)


drandyhaas:
## Summary

The **hybrid** diff-pair path (`_route_direct_coupled_middle`) cannot perform a
physical **polarity (P/N pad) swap**. Only the *standard* coupled route does. So
when a coupled leg falls through to the hybrid — common for a congested BGA escape
— and the pair's P/N ordering **flips between its two terminals**, the hybrid can't
swap; instead one polarity takes an ugly **wrap-around detour** to reach the
flipped side without crossing. We should add polarity-swap capability to the hybrid
path (or to its single-ended escape legs).

## Two different "polarity" operations today

1. **Physical pad polarity swap** — `apply_polarity_swap` / `undo_polarity_swap`
   (`polarity_swap.py`): reassigns which physical pad is + vs −. Resolves a
   required side-swap cleanly. Reached ONLY from the standard coupled route, via
   `route_diff_pair_with_obstacles(..., swap_allowed_ends=...)` in the multipoint
   leg loop (`diff_pair_multipoint.py` `_route_chain_attempt`).
2. **Coupled-ribbon relabel** — the `[polarity flipped: cleaner/shorter legs]`
   step inside `_route_direct_coupled_middle` (#248): assembles both P/N
   orientations of the parallel middle and keeps the cleaner legs. This only
   *labels* the two offset tracks; it **cannot** resolve a side-swap.

The hybrid has (2) but not (1).

## Worked example — fpga_sdram `/D1` (leg `U1:J1/J2 -> J3:4/6`)

```
Leg 1/2: U1:J1/J2 -> J3:4/6
  Error: source - no valid position at any setback/direction   <- standard route FAILS at the FPGA
  DIRECT HYBRID: coupled middle on F.Cu ... [polarity flipped: cleaner/shorter legs]
  Leg 1 via hybrid
```

The standard route (which *could* polarity-swap) fails to escape the congested
FPGA, so the leg falls to the hybrid. The pair's vertical order flips between the
two terminals:

- FPGA: `D1+ = U1.J1`, `D1- = U1.J2`  (D1− escapes on the lower side)
- J3:   `D1- = J3.6` (y 106.37, **above**), `D1+ = J3.4` (y 107.37, below)

With no polarity swap available, the hybrid routes **D1− the long way around** — a
[...]

Issue #381, opened 2026-07-13
Tier D CLI/GUI parity gaps D1-D9 (from review #368): GUI/plan-replay routes differently from CLI


drandyhaas:
Consolidated tracking issue for **Tier D — CLI/GUI parity gaps** from the codebase review in discussion #368. An agent diffed every argparse surface against its GUI call sites. These are behavioral divergences where a GUI (or plan-replayed) run routes differently from the equivalent CLI command — several are silent SI/connectivity regressions. Directly in scope of the "Keep CLI and GUI routing in sync" contract in CLAUDE.md.

Each item is an independent fix; check it off as it lands.

---

### D1 — Differential tab silently drops ~30 engine params it collects
`differential_gui.py:901-939`: `ordering_strategy` (GUI diff orders pairs DIFFERENTLY from a default CLI run — engine default `inside_out` vs CLI default `mps`), max rip-up, turn/heuristic costs, bus/proximity params, length matching (CLI diff can length-match, GUI diff cannot), the keepout checkbox is a no-op for diff, teardrops, etc. Root cause: no kwargs-dump harness for `batch_route_diff_pairs` (`KICAD_DUMP_BATCH_KWARGS` only covers `batch_route`).

### D2 — Plan-executor reset turns diff GND return vias OFF
`swig_gui.py:2204` resets to `False`; CLI default is ON (the negative flag `--no-gnd-vias` means manifests record nothing when ON) → every plan-replayed diff step routes without GND return vias. Silent SI regression.

### D3 — Polarity-swap default inverted between fronts
GUI/reset default `'*'` (allow all) vs CLI deny-all (#279); and `polarity_swap_nets` has no `claude_plan` alias, so a scoped allowlist silently widens to `'*'` on replay. Connectivity-affecting.

### D4 — Diff width defaults differ 3× and manifest `--track-width` is misrouted on plan replay
CLI diff width 0.3 vs GUI 0.1; manifest `--track-width` is routed to the Basic-tab control on plan replay, and `manifest_to_plan.py:46` also carries a nonexistent `--diff-pair-width`.

### D5 — claude_plan can't reach several route.py controls
`rip_existing_nets`, `impedance` (route action), `ordering`, `direction`, `time_matching`, `keepout`, `guide_corridor`, `length_match_groups`, `swappable_nets` all fall to "no control, ignored" → GUI plan replay of #284-class rip-existing chains silently routes without the rescue.

### D6 — Planes tab parity
[...]

Issue #215, opened 2026-06-26
route_diff: coupled differential-pair P/N traces overlap below clearance at bends (SEGMENT-SEGMENT)


drandyhaas:
Found during the #209 DRC step-attribution. On **cparti_fpga**, 8 of the 14 final DRC violations are the coupled diff pair `/USB/USB_D_P ↔ /USB/USB_D_N` overlapping each other below clearance. Introduced by `route_diff` (step3_diff_pairs: +9).

### Evidence (cparti_fpga, clearance 0.1)
```
/USB/USB_D_P <-> /USB/USB_D_N   In1.Cu  Overlap 0.017mm
  Seg1 (165.20,90.20)-(165.00,90.20)   Seg2 (164.70,89.90)-(164.80,90.00)
/USB/USB_D_P <-> /USB/USB_D_N   In1.Cu  Overlap 0.017mm
  Seg1 (154.70,97.60)-(154.70,96.90)   Seg2 (154.50,95.90)-(154.50,96.70)
```
The P and N traces of the coupled pair come 0.017 mm too close (effectively touching) at corners/bends — the coupling geometry isn't keeping the intra-pair gap >= clearance at turns. Clusters at two bends near (165,90) and (154.7,97) on In1.Cu.

This is the diff-pair-coupling clearance class flagged in the #209 redo SUMMARY ("route_diff couples more aggressively → via/segment clearance violations"); here it's the **P/N segment** spacing at bends specifically.

Reproduce: replay `runs_set2/cparti_fpga/redo_commands.sh`; board at `~/Downloads/unconnected/cparti_fpga.kicad_pcb`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

drandyhaas:
**Fold-in from the 45-board redo (HEAD 05b22a8): the SEGMENT-CROSSING class belongs here too.**

Beyond the sub-clearance overlap at bends, the redo found **3 outright P/N
crossings** of a coupled pair — the severe form of the same route_diff
trace-coupling defect:

- schoko: `/USB1_DP <-> /USB1_DN` cross at (128.903,122.200) F.Cu; `/DVI_CK_P <-> /DVI_CK_N` cross at (132.103,131.800) B.Cu
- butterstick: `/SYZYGY0.D3_P <-> /SYZYGY0.D3_N` cross at (126.700,102.900) F.Cu

Same root cause (the two coupled traces routed without keeping P clear of N), just
[...]

Issue #530, opened 2026-07-30
Make default routed constraints equal what KiCad's DRC


GreenFoot:
**Type:** enhancement · **Area:** design-rule ingestion (`kicad_dru.py`, `routing_config.py`, `list_nets.py`, `check_drc.py`)
**Audited at:** `aa0c7cf` (v0.19.1-77-gaa0c7cf)

## Goal

**Acceptance criterion: for every constraint, the value KRT routes and grades to
equals the value KiCad's DRC engine enforces.** Not "KRT implements a precedence
order that we believe matches KiCad" — that is a means, and an unreliable one,
because the order is only knowable by asking KiCad.

This reframing changes the shipping order: the **agreement harness comes first**,
before any new constraint is honored. It turns every precedence question into a
measurement instead of a recollection, and leaves behind a permanent regression
gate.

## Two problems

1. **Coverage.** `kicad_dru.py` honors exactly one constraint (`clearance`, `min`
   only, layer-scoped only). `via_diameter`, `hole_size`, `annular_width`,
   `track_width`, `hole_to_hole`, `edge_clearance`, `diff_pair_gap`,
   `physical_clearance` and `disallow` are ignored — silently.
2. **Structure.** Precedence in KRT is *emergent, not expressed*: the tiers are
   pre-baked into different structures at different times, and 84 call sites are
   each responsible for composing them correctly.

### Coverage

- `kicad_dru.py:155` — `if item[0] == "constraint" and item[1] == "clearance"`.
  Everything else falls through `clearance_mm is None: continue` (line 159-160)
  **with no note** — not even the "skipped, KiCad will still enforce it" warning
  that out-of-subset *scopes* get. A board with `(constraint via_diameter (min
  0.6))` gets zero indication the router ignored it.
- Only `(min ...)` is read; `opt` / `max` are dropped.
- `_rule_layers` accepts only pure-layer scopes (`(layer <name>|outer|inner)`
  and/or a `condition` of `||`-joined `onLayer()` terms). Netclass, netname,
  `insideArea`, `intersectsCourtyard`, `&&`, negation → skipped with a note.
- `(disallow ...)` is not parsed at all.

Two gaps are more than plumbing:

- **Via sizing has no per-net or per-layer path.** `config.via_size` /
  `via_drill` are scalars. A per-net clearance map and a per-net track-width map
  exist; no via equivalent — so even a **netclass** `via_diameter` off Default
  isn't applied per net.
[...]

PR #644, opened 2026-08-14
Add polarity (P/N pad) swap to the 2-terminal hybrid diff-pair path (Refs #266)

theironchef:
Refs #266 — this is **part 1**: the 2-terminal hybrid path only. It does **not** close the issue; see *Scope* below.

## The bug

`_route_direct_coupled_middle` (the DIRECT HYBRID / coupled-middle + point-to-point-legs escape) already computed whether a pair is **side-flipped** — the P pad on the `+` side of the coupled centerline at one end and the `-` side at the other — and threw the answer away with the comment *"costs one crossing either way"*. Whichever offset ribbon it labels P, one terminal leg then has to wrap around its partner. Its success dict carried no `polarity_fixed` / `swap_target_pads`, so `apply_polarity_swap` bailed at its first line: **the hybrid could never produce a physical pad swap.**

## What changed

**Engine (`py_router/diff_pair_routing.py`)**

* Name the flip: `side_flipped = (src_plus is not None and tgt_plus is not None and src_plus != tgt_plus)`.
* `_assemble(pol, swap_end='tgt')` exchanges the two TARGET terminals, so the P leg routes to the N pad and vice versa.
* Teach the gates that read pad/via nets about the **pending** swap, mirroring the existing `_pn_tracks_cross_full(p_swap_positions=…)` pattern:
  * `_leg_partner_copper` stops stamping the pad (and stub copper) a swapped leg has to land on;
  * `_connector_grazes_foreign_copper` re-buckets the two swapped **pads** *and* re-labels the **vias** at the swap positions;
  * `_hybrid_route_connects` grades post-swap connectivity.
* Swapped candidates are **appended** to `cands`; the existing selector (fewest intra-pair overlaps → via-weighted length → first-on-tie) picks one only when it is the better candidate. When the pair is not permitted a swap the list is byte-identical to before — that is the inertness guarantee.

**Call site (`py_router/diff_pair_loop.py`, the HYBRID ESCAPE)** opts in with `allow_polarity_swap=True` and applies the swap **before** `add_route_to_pcb_data`. The engine kwarg defaults to `False`, so the other three hybrid call sites — which would drop a `polarity_fixed` result on the floor and ship copper on the wrong pads — cannot request one.

**Disclosure.** A side-flipped pair routed *without* a swap is named on the `DIRECT HYBRID` line and in `JSON_SUMMARY.polarity_flip_unswapped_pairs`, recorded at **all three** hybrid commit sites (first-pass, multipoint leg, reroute).

### The via half of the third gate

`_connector_grazes_foreign_copper` re-bucketed the swapped **pads** but read every **via** at its pre-swap net. A swapped leg landing on a via-fanned target was therefore graded as grazing a "foreign" via that the pending swap hands to its own net — so on the ordinary BGA/QFN escape topology the feature was silently unreachable. Fixture: the in-test watchy fanout with a 0.6 mm stub + via at each U5 USB_D target. Candidate dump, `--polarity-swap-nets '*'`:

```
BEFORE
[PROBE266] plain cand pol=1  -> REJECT: terminal legs could not attach to the coupled middle
[PROBE266] plain cand pol=-1 -> OK len=32.892 ov=0 vias=2
[PROBE266] swap  cand pol=1  -> REJECT: assembled route grazes foreign via (net USB_D-) by 375um
[PROBE266] swap  cand pol=-1 -> REJECT: assembled route grazes foreign via (net USB_D-) by 375um

AFTER
[PROBE266] plain cand pol=1  -> REJECT: terminal legs could not attach to the coupled middle
[PROBE266] plain cand pol=-1 -> OK len=32.892 ov=0 vias=2
[PROBE266] swap  cand pol=1  -> OK len=30.690 ov=0 vias=4
[PROBE266] swap  cand pol=-1 -> REJECT: assembled route grazes foreign via (net USB_D-) by 38um
```

The surviving `pol=-1` rejection is a genuine 38 µm graze, not the 375 µm structural one — the gate still bites. Watchy's own U5 has bare SMD pads with no via, which is why the original two test arms never caught this.

## Verification

**Red / green** (`tests/test_hybrid_polarity_swap.py`, three arms):

```
$ git checkout main -- py_router/ && python3 tests/test_hybrid_polarity_swap.py ; echo EXIT=$?
  FAIL side flip disclosed in the summary (got None)
  FAIL the hybrid produced a polarity swap (got [])
  FAIL pair copper shrinks: 36.710 mm < 36.710 mm (-0.000 mm, 36 segs vs 36)
  FAIL U5 pad nets actually exchanged: {'1': 'USB_D+', '3': 'USB_D-'}
  FAIL with it, the via the swap hands over is this leg's own (got TypeError: ...unexpected keyword argument 'n_swap_positions', want None)
  FAIL the exemption is scoped to the swapped positions (got TypeError: ..., want ('via', 375))
  FAIL the declined swap is disclosed, not silent (got None)
FAIL (7)
EXIT=1

$ python3 tests/test_hybrid_polarity_swap.py ; echo EXIT=$?
  ... 25 ok, 0 FAIL ...
All checks passed (#266: hybrid polarity swap fires under --polarity-swap-nets and is inert without it)
EXIT=0
```

Arm 2, watchy USB_D with the flag: **36.710 mm / 36 segments → 34.953 mm / 24**, `ALL NETS FULLY CONNECTED`, `NO DRC VIOLATIONS FOUND` at the routed 0.1 mm.
Arm 3, the via fixture: the gate itself — `('via', 375)` without the pending-swap view → `None` with it → `('via', 375)` again when the *other* target's position is named, so the exemption is position-scoped rather than a blanket partner-via pass.

**The disclosure now agrees with the console.** Same board, same command, before/after — the extra names come from the two commit sites that printed the `[side-flipped …]` line and then dropped the key:

```
glasgow_revC, --polarity-swap-nets '*'   (the reroute commit site)
  BEFORE  console=2  HYBRID ESCAPE (reroute)=1  summary=["/IO_Banks/Z5"]
  AFTER   console=2  HYBRID ESCAPE (reroute)=1  summary=["/IO_Banks/Z5", "/IO_Banks/Z8"]

orangecrab_ext_pll, no flag              (the multipoint leg commit site)
  BEFORE  console=2  leg-hybrid=2              summary=["RAM_UDQS"]
  AFTER   console=2  leg-hybrid=2              summary=["RAM_CK", "RAM_UDQS"]
```

The same gap is larger on `routed_output` (61 pairs): pre-fix, 12 console prints against 11 names without the flag, and 4 prints (all four from `HYBRID ESCAPE (reroute)`) against `[]` with it — those two are pre-fix measurements only, the matched post-fix run on that board was not completed. Copper is byte-identical across each before/after pair, so the disclosure moves no tracks: glasgow with the flag 703.351 mm / 672 segs / 70 vias on both arms, orangecrab 481.880 mm / 1020 segs / 149 vias with the same 27 pre-existing DRC violations on both.

**Inertness** — no `--polarity-swap-nets`, whole-board totals:

| run | copper | segments | vias |
|---|---|---|---|
| `glasgow_revC --nets '*'` main | 289.031 mm | 245 | 24 |
| `glasgow_revC --nets '*'` patched | 289.031 mm | 245 | 24 |
| `glasgow_revC --nets '*' --clearance 0.1` main | 460.445 mm | 442 | 34 |
| `glasgow_revC --nets '*' --clearance 0.1` patched | 460.445 mm | 442 | 34 |
| `glasgow_revC` full geometry¹ main | 689.721 mm | 648 | 64 |
| `glasgow_revC` full geometry¹ patched | 689.721 mm | 648 | 64 |
| `orangecrab_ext_pll` full geometry¹ pre-fix | 481.880 mm | 1020 | 149 |
| `orangecrab_ext_pll` full geometry¹ patched | 481.880 mm | 1020 | 149 |

¹ `--layers F.Cu In1.Cu In2.Cu B.Cu --track-width 0.1 --diff-pair-gap 0.15 --clearance 0.1 --via-size 0.3 --via-drill 0.2 --no-gnd-vias --grid-step 0.05`

**The discard path.** `apply_polarity_swap` returns `False` *without mutating* when the pair is already in `polarity_swapped_pairs` — swapped earlier in the run, then ripped and re-reaching this hybrid. Dropping the whole hybrid there sent a pair that base couples to the single-ended follow-up; it now re-asks without a swap. Forced `False` return at that call site, watchy fixture:

```
BEFORE                                           AFTER
DIRECT HYBRID: ... [polarity pad swap at target]  DIRECT HYBRID: ... [polarity pad swap at target]
[INJECT] apply_polarity_swap -> False             [INJECT] apply_polarity_swap -> False
Hybrid polarity pad swap could not be applied     Hybrid polarity pad swap could not be applied
  - discarding the hybrid route                     - retrying the hybrid without a swap
ROUTE FAILED - no rippable blockers found         DIRECT HYBRID: coupled middle on In1.Cu + 28 leg seg(s)
Deferring USB_D to the single-ended follow-up     HYBRID ESCAPE: direct coupled middle + ...
Diff pairs:    0/1 routed                         Diff pairs:    1/1 routed
```

The re-asked route is the base hybrid exactly — `USB_D+ 17.889 mm / 13 segs`, `USB_D- 18.821 mm / 23 segs`, fully connected, DRC-clean.

**Suites**: `run_all.py --fast` 216 passed / 0 failed / 47 skipped (172.9 s); `run_all.py diff hybrid polarity` 22 passed / 0 failed; `tests/gui_parity/test_manifest_plan_parity.py` 35 flag-checks, 0 mismatches (exit 0); `tests/gui_parity/test_cli_postpass_coverage.py` 0 unreachable, 0 failures (exit 0); `tests/run_doc_examples.py` 31 run / 77 skipped / 5 failed — identical to `main` (a missing `kicad_files/test_diffpair_ram.kicad_pcb` fixture, pre-existing and unrelated). The exported patch applies cleanly to a fresh `main` checkout, where the new test passes on its own.

CLI/GUI parity: no new flag and no new dialog control. The gate is the existing `--polarity-swap-nets` / `polarity_swap_allowed`, the swap lands in `state.pad_swaps` which both fronts already consume, and the new summary key is built inside the shared `batch_route_diff_pairs`, so the GUI gets it free. `docs/differential-pairs.md` is also corrected: the GUI's "Polarity-swap allowed nets" field has defaulted to **EMPTY**, not `*`, since #381 D3 (`kicad_routing_plugin/differential_gui.py:586`) — the inertness argument rests on deny-by-default at both fronts.

## Scope, and what this does NOT do

**The issue's own worked example is not fixed by this PR.** `Leg 1/2: U1:J1/J2 -> J3:4/6` is the multipoint leg hybrid (`diff_pair_multipoint.py`), which passes `terminal_pads` and is therefore double-gated out of building a swap candidate. On that path this change contributes the **disclosure and nothing else**. Follow-ups: the multipoint leg loop (needs source/target end policy plus `undo_polarity_swap` on the rip path) and `reroute_loop.py`.

**A swap that wins its own candidate list is not a board-level win.** The `min()` ranks candidates for *one pair on one layer combination*; the pad rename is permanent and re-shapes the board for every pair routed afterwards. Measured on `glasgow_revC`, flag off vs flag on, same code, `failed=[]` and DRC-clean on every row — it goes **both ways** depending on the chain:

| chain | pairs | swaps | copper | segments | vias |
|---|---|---|---|---|---|
| full geometry¹, no flag | 14/16 | 0 | 689.721 mm | 648 | 64 |
| full geometry¹, `--polarity-swap-nets '*'` | 14/16 | 4 | **703.351 mm** (+13.630) | 672 | 70 |
| `--clearance 0.1`, no flag | 9/16 | 0 | 460.445 mm | 442 | 34 |
| `--clearance 0.1`, `--polarity-swap-nets '*'` | 9/16 | 4 | **447.404 mm** (−13.041) | 403 | 32 |
| `--track-width 0.1 --diff-pair-gap 0.15 --clearance 0.1 --via-size 0.3 --via-drill 0.2`, no flag | 13/16 | 0 | 572.605 mm | 490 | 32 |
| same, `--polarity-swap-nets '*'` | 13/16 | 1 | 571.981 mm (−0.624) | 486 | 32 |

Every row routes the *same set* of pairs with the same `member_incomplete_pairs`, so the copper delta is the swap and nothing else. The first pair of rows is the case a maintainer would not expect from "strictly cleaner/shorter": four pairs at J5 swap and the board gains 2 % copper and 6 vias. `--polarity-swap-nets` is a per-pair permission, not an optimisation switch, and the wording now says so.

**Other residual limitations**

* **Target end only.** `swap_end` accepts only `'tgt'`; `apply_polarity_swap`'s contract is target-shaped.
* **The swapped legs cannot reuse the partner's existing target via as a launch island**, so on a via-fanned target the swap pays two extra vias and usually loses the selector's comparison even though its candidate is now legal (30.690 mm / 4 vias vs 32.892 mm / 2 on the arm-3 fixture). The gate fix is what makes it *reachable*; making it *win* there is a separate change, and arm 3's end-to-end half is explicitly not a gate on it.
* **The test fixture is electrically wrong on purpose.** watchy USB_D is USB D+/D−, which this repo's own `identify-diff-pairs` skill classifies as polarity-critical and never to be swapped. Forcing `--polarity-swap-nets '*'` exercises the mechanism; it is not a swap a user should make on this board.
* **Cost when the flag is on**: a permitted, side-flipped pair runs up to two extra `_assemble` calls per layer combination — negligible on the fixture, untimed on a large flag-enabled board.

## Post-review addendum (second adversarial pass: VERDICT SHIP)

The second review verified all 8 prior findings fixed by execution — including
rebuilding the via-fanned fixture from scratch and reproducing the before/after
candidate dumps verbatim, all six glasgow/orangecrab table rows, and the
byte-identical no-flag board hashes. Two follow-ups from its notes are in the
head commit:

- The **fourth** hybrid commit site (`_maybe_swap_to_hybrid`, the #215 pinch
  swap) now records the disclosure too — dormant on every in-repo board, but the
  same console-vs-summary class as F4.
- The `plan-pcb-routing` skill now surfaces `polarity_flip_unswapped_pairs`
  next to the denied list it already reads, so the consumer of `JSON_SUMMARY`
  actually shows the disclosure this PR adds.

One correction to a line above, from that pass: of the 5 pre-existing
`run_doc_examples` failures, 3 (not all 5) are the missing
`test_diffpair_ram.kicad_pcb` fixture; 1 is a missing `fanout_output.kicad_pcb`
and 1 a non-UTF-8 doc block. All identical on `main`.


drandyhaas:
Thank you for this, and apologies for the three weeks. Closing it — again not on
the merits. The diff-pair engine has moved a long way underneath this branch,
and #266 is better served by being revisited whole than by landing part 1 into a
tree it no longer applies to.

## The finding I want to keep

The via half of the third gate is a genuinely good catch, and it's the kind of
thing that stays hidden for a long time:

> `_connector_grazes_foreign_copper` re-bucketed the swapped **pads** but read
> every **via** at its pre-swap net. A swapped leg landing on a via-fanned target
> was therefore graded as grazing a "foreign" via that the pending swap hands to
> its own net — so on the ordinary BGA/QFN escape topology the feature was
> silently unreachable.

And the reason the original arms missed it — watchy's U5 has bare SMD pads with
no via — is the useful half. That's the fixture note that makes this reproducible
later, and the before/after candidate dump showing the surviving `pol=-1`
rejection is a real 38 µm graze rather than the structural 375 µm one is exactly
the right way to show a gate still bites. Both are being carried onto #266.

## Why now isn't the moment

**The branch is conflicting.** Thirteen commits have touched
`diff_pair_routing.py` / `diff_pair_loop.py` since 14 Aug, several of them right
through the surfaces this PR modifies:

- `1f9a4f1f` (#806) — the diff-pair engine keeps its working obstacle map in step
  with the copper it commits, which is the same commit-time bookkeeping the
  polarity swap has to interleave with;
- `44d3c240` / `5cae0abc` (#530) — a pad clearance override now *replaces* the
  class/rule value, and multipoint pairs resolve per-pair geometry. The graze
  gates this PR teaches about pending swaps price clearance differently now;
- `33e9f43d` (#766) and `9e1d562c` (#764) — the hybrid escape gained its own
  disclosure and needs-fanout paths, which overlap the
  `polarity_flip_unswapped_pairs` disclosure added here.

A mechanical rebase would resolve the text, but the three gates this PR extends
(`_leg_partner_copper`, `_connector_grazes_foreign_copper`,
`_hybrid_route_connects`) are precisely where those commits landed, so the
re-verification cost is close to rebuilding it — and the red/green arms would
need re-establishing against the current engine before any of it could be
trusted.

**#266 is still open and still wants the whole thing.** This is part 1 of a
multi-part change by design, and the remaining hybrid call sites still can't
request a swap. Landing one part into a substantially changed engine means the
inertness guarantee ("when the pair is not permitted a swap the candidate list is
byte-identical") has to be re-argued once per part against a moving base, rather
than once for the finished feature.

## What happens next

#266 stays open. When it's picked up it'll be as one piece across the hybrid call
sites, and this PR is the design reference for it: `side_flipped` named rather
than discarded, `_assemble(pol, swap_end='tgt')`, the three gates taught about
the pending swap mirroring the existing `_pn_tracks_cross_full(p_swap_positions=…)`
pattern, swapped candidates *appended* so the existing selector picks one only
when it wins, `allow_polarity_swap` defaulting to `False` so the other three
hybrid sites can't request a swap they'd drop on the floor, and disclosure at all
three commit sites. I'll link this PR from the issue so the candidate dumps and
the watchy fixture are one click away.

The branch stays as it is, so none of the work is gone. Thanks for the care in
the writeup — the scope section being explicit that this does *not* close #266 is
what made it easy to see where it fits.

Issue #621, opened 2026-08-12
Fanout escapes should be cancelable from the GUI, not bounded by a wall clock


edgehero:
The four routing mains stop on their own budget (exit 7, `complete:false`
summary); the fanout mains cannot — an external timeout on Windows is
`TerminateProcess`, which leaves no summary and no exit code of the tool's
own, so a harness cannot tell "killed mid-write" from "crashed". Run 9 lost an
entire campaign to two non-terminating steps with nothing in the log to say
so. Unlike `create_plane`/`repair_planes`, the fanout engines have zero
`cancel_check` call sites (verified at tip), so this is not one closure into
existing plumbing but a real engine change to
`generate_bga_fanout`/`generate_qfn_fanout` loop heads — flag-shaped and
design-heavy, hence an issue for the maintainer rather than a patch.

drandyhaas:
## Reframed and implemented on both branches — retitled to match

The premise here ("the tools must stop themselves") has been **deliberately reversed**: no result any of these tools produces may depend on a wall clock, so nothing self-budgets, and the only cancel in the system is a human clicking Cancel. Retitled accordingly.

### 1. `--deadline` is gone repo-wide, and `krt_deadline.py` is deleted (`placement`)

PR #643 gave the fanout mains a wall-clock budget. That budget is now removed, along with the whole facility it belonged to — 10 mains (`route`, `route_diff`, `route_planes`, `repair_planes`, both fanouts, `place_seed`, `place_reconstruct`, `place_portfolio`, `place_optimize`, `beautify_labels`, `converge`), `KRT_DEADLINE_S`, exit 7, and the partial/`deadline_skipped` reporting. It reached the engine API too: `seeder.py`'s three entry points dropped their `deadline`/`progress` parameters and the `deadline_skipped`/`complete` keys from their result contracts, and `plane_fragility` no longer shortens a KiCad refill against a run clock.

Two things worth knowing for anyone reading the old code:

- **`krt_deadline.emit()` had become the sole `JSON_SUMMARY` emitter** for `route_planes`, `repair_planes`, `place_reconstruct`, `place_seed`, `place_portfolio`, `place_optimize` and `beautify_labels`. Each now prints directly, preserving the `complete`/`status` defaults consumers read — `route_summary`'s sticky-incompleteness merge and `place_route_loop`'s partial guard still work.
[...]

Issue #326, opened 2026-07-07
check_drc ignores footprint/pad-level clearance overrides (hackrf: 52 KiCad-only clearance items)


drandyhaas:
From the #316 sweep: hackrf_one's final grades 0 by check_drc but 52 clearance items by kicad-cli at the equalized routed clearance — every item citing a LOCAL override, e.g. "footprint U17 clearance 0.1524 mm; actual ..." / "pad clearance 0.1778 mm".

KiCad honors per-pad `(clearance ...)` AND footprint-level clearance attributes; check_drc grades one global clearance. The parser carries `pad.local_clearance` (per-pad) but check_drc doesn't apply it pairwise, and footprint-level clearance isn't parsed at all. The ROUTER has the same blindness (obstacle margins are global), so these are real KiCad-rule violations the router was never told to avoid.

Fix sketch: parse footprint-level clearance into Footprint (inherited by its pads when pad.local_clearance is 0); grade pad-involving checks at max(global, pad_effective_clearance); stamp pad obstacle margins with the same effective clearance so routing complies. Needs a corpus A/B (routability cost on boards with generous overrides).

Repro: `python3 tests/stress/kicad_drc_compare.py ~/Documents/kicad_stress_test/ab_320final_0706/set1/hackrf_one/step5e_final.kicad_pcb` (pass the routed clearance).

---

## Folded in from review #368 (Tier B — netclass/local-clearance modeling gaps)

Same root family as above: the router and check_drc model a single global clearance, so higher per-net-class / per-pad clearances are honored inconsistently across the base map, the per-net cache, the post-passes, and the grader. Confidence markers: **[V]** verified with a minimal repro; plain = traced but one link unverified.

### B5 — Netclass clearances enforced only against static copper
`obstacle_map.py:58-59` vs `:906,:1139,:1175`: same-run nets stamp each other at bare `config.clearance` — two 0.3mm-netclass nets in one GUI batch route 0.1mm apart. check_drc (single-value) is blind; KiCad flags it. (The routing side of this issue's "obstacle margins are global" blindness, now against *same-run* copper rather than static.)

### B6 — `pad.local_clearance` honored by the base map but by NO post-route pass, NOT by the per-net obstacle cache, NOT by check_drc
[...]

PR #679, opened 2026-08-20
route.py: --deadline with a reserve band, and machine-readable run output

edgehero:
Deadline handling and machine-readable run output, split out of #679 after review. Two commits, one PR, because they depend on each other: the short summary line is how a deadline verdict gets read, and each one's test needs the other.

## What changes for a user

**Commit 1, deadline.** `route.py --deadline SECONDS` (or `KRT_DEADLINE_S`): the routing loops stop at deadline minus a reserve, the tail legs (plane finalize, oracle, reconcile) run inside the reserve on their own closure, a reconcile that would start on a spent budget is refused with the reason printed, and the board written is a real partial route with `complete=false / status=deadline` in its summary and exit code 7. Without a budget nothing changes. The motivation is Windows: an external kill there is `TerminateProcess`, which leaves no summary and no exit code of ours, and a whole lap was once lost that way.

**Commit 2, output.** One `JSON_SUMMARY_MIN:` line per outermost route with the merged tally; the "boxed in by static obstacles" verdict becomes a summary key (`boxed_in`, next to `blockers`) and plane-finalize exclusions become `finalize_excluded_nets`, both previously only sentences in the log; `check_reachability` gets auto-widen and `--json-out`, `net_forensics` gets `--json`.

## Files

`py_router/krt_deadline.py` (new, stdlib only), `route.py` (deadline plumbing and the summary keys), `route_summary.py`, `routing_diagnostics.py`, `reroute_loop.py`, `single_ended_loop.py`, `py_tools/check_reachability.py` (plus the two placement helpers it is built on and whose changes its auto-widen and `--defect-json` relief need: `placement/reachability.py` and `placement/routability.py`), `py_tools/net_forensics.py`, and four tests. 14 files. Standalone on main: compiles, all four tests green (`test_deadline`, `test_run23_deadline_reserve`, `test_run23_summary_min`, `test_boxed_in_summary`), plus upstream's `test_json_out_summary` still green.

### DIFF
--- a/py_placer/placement/reachability.py
+++ b/py_placer/placement/reachability.py
@@ -44,6 +44,7 @@ because a silently-degraded reachability answer is worse than none.
 from __future__ import annotations
 
 import math
+import os
 from dataclasses import dataclass, field
 from typing import Dict, List, Optional, Sequence, Tuple
 
@@ -212,7 +213,8 @@ def slack_field(pcb, target_net_id, layer, view, base_clearance,
 # widest path
 # --------------------------------------------------------------------------
 
-def widest_path(slacks, targets, via_ok, seed_xy, seed_layer, view, step):
+def widest_path(slacks, targets, via_ok, seed_xy, seed_layer, view, step,
+                return_throat=False):
     """Kruskal widest path over a multi-layer grid; returns mm or None.
 
     Nodes are (cell, layer). In-layer edges join adjacent ACTIVE cells; a via
@@ -223,6 +225,13 @@ def widest_path(slacks, targets, via_ok, seed_xy, seed_layer, view, step):
     Returns the largest t such that a path exists using only cells of slack
     >= t. Exact for the raster: no search order, no grid alignment, no
     heuristic. None means no positive-slack path exists at any width.
+
+    `return_throat=True` returns `(value, (layer, x_mm, y_mm))` instead. The
+    cell that CLOSED the path is the throat -- the narrowest point on the widest
+    route -- and it was already in hand here (`lay`, `x`, `y` at the moment the
+    seed joins the target) and thrown away at the bare `return float(val)`.
+    Every consumer that wanted to say WHERE the bottleneck was had to go and
+    find it again by eye. `None` throat when there is no path.
     """
     L = len(slacks)
     h, w = slacks[0].shape
@@ -274,8 +283,13 @@ def widest_path(slacks, targets, via_ok, seed_xy, seed_layer, view, step):
                 if other != lay and active[other * n + cell]:
                     union(idx, other * n + cell)
         if active[seed] and find(seed) == find(TARGET):
-            return float(val)
-    return None
+            if not return_throat:
+                return float(val)
+            # `lay`, `y`, `x` are this cell -- the LAST one activated, i.e. the
+            # narrowest on the widest path. Cell centre, not corner.
+            return float(val), (lay, view[0] + (x + 0.5) * step,
+                                view[1] + (y + 0.5) * step)
+    return (None, None) if return_throat else None
 
 
 @dataclass
@@ -295,6 +309,17 @@ class Reachability:
     grid: Tuple[int, int]
     note: str = ''
     open_room_mm: float = 0.0     # the cap free space was clamped to
+    # WHERE the bottleneck is: (layer_index, x_mm, y_mm), or None when there is
+    # no path. Everything below exists because run 20 measured a throat, printed
+    # the number, and then had to hand-assemble the coordinates, the blocking
+    # refs and the relief distance into an English paragraph in a ledger string
+    # -- three tools' stdout, none of it consumable by anything downstream.
+    throat_cell: Optional[Tuple[int, float, float]] = None
+    #: The base clearance the field was built at. Needed to convert between the
+    #: two SPACES this class reports in -- see `gap_mm`.
+    clearance_mm: float = 0.0
+    #: Foreign copper nearest the throat: [{ref, pad, kind, x, y, layer, dist}].
+    near: Tuple[dict, ...] = ()
 
     @property
     def wide_open(self) -> bool:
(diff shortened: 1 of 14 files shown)


drandyhaas:
Reviewed commit 3033bcdc ("check_drc and parser fixes: custom pad circles, per-pad clearance, dup-ref warning vs bare-JSON stdout") in detail, and we're declining this part of the PR.

The headline fixes are stale after the rebase: the custom-pad-circle and per-pad-clearance fixes are already on main (test_run4_custom_pad_circle.py and #326 local_clearance landed earlier), so what the commit actually adds is the poly-index speedup, the via-annular rung, size-floor provenance, fab_floor_origin hole-clearance grading, and the duplicate-reference warning. None of these are important bug fixes or features we want to add.

Please drop this commit from the PR, along with its riders elsewhere: tests/test_point_in_poly_index.py (in 305089d9, it asserts this commit's internals) and the list_nets docstring hunk. The rest of the PR is still under review.

drandyhaas:
To be explicit about the parser half of 3033bcdc (the duplicate-reference warning + `PCBData.duplicate_references`): this is also declined, same grounds.

The "warning printed to stdout and broke bare-JSON tools" bug being fixed never existed on main — main has no duplicate-reference warning at all, so that was a defect introduced and fixed within the fork's own history. What the rebased commit actually adds to main is a new advisory feature for a legal-in-KiCad, rare situation (duplicate references), not an important bug fix or a feature we want. Please drop the kicad_parser.py hunks along with the rest of 3033bcdc.

drandyhaas:
Also declining the board-declared fab floors group: c02af774 (`--board-floors` + the writeback hold), e8ecc9a7 (under-bind warning), b5c1a861 (the test_run12/tigard.kicad_pro rider), and the `--board-floors` dialog/settings wiring in 4977b8a1.

Reason: we're going to have all floors enforced with kicad_dru files. The repo already auto-reads the sibling `.kicad_dru` in every routing step and in check_drc (#498, replacement semantics, outranks `--clearance`), and that rules file is the single source of truth we want to extend — not a parallel binding mechanism reading floors out of the `.kicad_pro`. Please drop these commits (including the kicad_files/tigard.kicad_pro fixture) and the dependent writeback-hold hunks in fix_kicad_drc_settings.py.

The manifest_to_plan `--fab-tier`/`--fab-overrides` retention fix in 4977b8a1 is separate from this and still under review, as are the deadline, summary-tracking, and test-infra commits.

edgehero:
Thanks for the detailed look, and fair on all three counts.

The check_drc/parser commit was described by what the original commits intended, not by what the rebase actually added on top of today's main. That is on me. It is gone, with its riders (test_point_in_poly_index, the list_nets hunk, the duplicate-reference parser change).

The board-declared floors group is gone as well: --board-floors and the writeback hold, the under-bind warning, the GUI wiring, the tigard project file, the fix_kicad_drc_settings hunks, and the fanout/net_rescue hooks that rode on it. Understood on .kicad_dru being the one mechanism.

While doing that I went through the rest hunk by hunk against current main with the same question, and cut more: the check_complete provenance changes (same family), and a set of test edits that duplicated the #522 path repair main already has in its own form.

This PR is now just the deadline handling and the machine-readable summary line, 12 files, 2 commits, re-run green on the branch alone. The small unrelated fixes moved to #681 so they can be judged on their own. #680 will get the same treatment before it costs you a review.

drandyhaas:
On the manifest_to_plan `--fab-tier`/`--fab-overrides` piece: we took the substance as a native fix on main (ece1e011), so please drop that hunk from 4977b8a1 too — along with the `--board-floors` line (rejected group) and the `place_plan`/`REFUSED_TOOLS` rewording, which belongs with #680 if anywhere.

One correction to the PR description, from running the baseline control: the flags were not silently dropped. The converter's unknown-flag fallthrough carries both into the plan — `--fab-tier` actually worked on main all along, because the fallthrough name coincides with the dialog control's name. The real defect was only `--fab-overrides`, and at the APPLY layer: it falls through as `fab_overrides`, the control is `fab_overrides_path`, so ai_plan silently ignored it (losing the override file and re-enabling escalation on replay). Our fix maps both explicitly, adds a `fab_overrides` alias in ai_plan for plans converted before the fix, adds both flags to the parity gate's expectation table, and makes the gate's fixture manifest always run — no corpus manifest uses `--fab-overrides`, which is why the gate could never have caught this from corpus data alone. Gate verified: 8460 flag-checks / 403 manifests / 0 mismatches, and with the mapping reverted it fails with exactly "[route] --fab-overrides: want 'my_fab.txt' got None".

Status recap of this PR: rejected so far — 3033bcdc (check_drc/parser) and the board-floors group (c02af774, e8ecc9a7, b5c1a861, the --board-floors GUI wiring). Absorbed into main — the fab-flag manifest parity fix. Still under review — deadline handling (67ed16e8), summary tracking (e8ebc7fa), and the test-infra commits.

drandyhaas:
Thanks for the restructure — this shape is much easier to judge.

The deadline commit (50833389) is declined: wall-clock deadlines are not a feature we want. This isn't a quality call on the implementation — it's a standing design decision. This exact facility (krt_deadline.py, `--deadline`, `KRT_DEADLINE_S`, exit 7) existed here and was deliberately removed repo-wide on 2026-08-16 (#621): nothing should depend on timing for its results, and cancellation is a manual GUI click, not a clock. An opt-in flag doesn't change that; it reintroduces the facility and its maintenance surface. The Windows TerminateProcess problem it targets is real, but the answer we've chosen is to not run tools under external kill budgets rather than to teach every tool to race one.

The machine-readable output commit (067c574d) is still under review — but it has to stand without the deadline commit to be considered. Please drop 50833389 and rebase 067c574d standalone (without the `status=deadline` plumbing); JSON_SUMMARY_MIN, `boxed_in`/`finalize_excluded_nets` as summary keys, and the check_reachability/net_forensics JSON outputs all look separable from the deadline machinery on their own terms.

drandyhaas:
Closing this one too: with the deadline commit declined (a standing design decision, #621 — not a quality call), nothing in the PR's current shape is headed for merge, and everything else the original ten-commit version carried has been resolved — rejected with reasons, absorbed onto main (ece1e011, 9745cf50, 72caae0a, e268146d via #681), or split out by you.

The machine-readable output work (JSON_SUMMARY_MIN, boxed_in / finalize_excluded_nets as summary keys, check_reachability --json-out, net_forensics --json) is still of interest: if you'd like it considered, open a fresh PR with just that commit rebased standalone on current main, without the status=deadline plumbing, and it will get a prompt review. Thanks for the thorough work and the gracious restructuring throughout.

PR #680, opened 2026-08-20
Placement legality: courtyard blocking gate and body containment (closes #628)

### LINKED ISSUE #628: Interior milled contours are invisible to placement legality — a pad inside a milled ring passes every placement gate

### LINKED ISSUE #629: A no-legal-pose verdict should name its blockers — the census exists and converts directly into the next move
Measured on run 19: three consecutive seeder/reseat sweeps returned bare "no legal pose anywhere on the board" for SW17+SW34+REF_PUCK_R (the third was a complete null — a passthrough board written with all three refs unmoved). When the question was finally asked in scoped form, the engine answered precisely: with D14 in place, 0 legal thumb-pocket poses; with D14 lifted, 46. With D31: 0 → 32. One eviction each and both switches seated, which closed the run.

Suggested fix: when a sweep ends at 0 legal poses for a ref, run a bounded second pass that lifts each nearby incumbent in turn and counts poses freed; return `no_pose_blockers: {ref: {blocker: poses_freed}}` in the JSON summary and print the top entries. A verdict that names its blockers is the next move; a bare verdict is a dead end that cost this run one wasted invocation and arguably the whole hand-seeding detour.

Full evidence: LEARNINGS.md finding 2 (link above).

edgehero:
Placement legality: courtyard blocking and body containment. Split out of the original #680 after the review of its companion; the first of a short stack, each one claim. The rest open ONE AT A TIME as each lower one merges, so no PR ever shows a cumulative diff: next is the seeder (#682, 11 files, closes #629/#630), then review instruments (#683, 16), the recovery harness (#684, 35), and the skill docs (#685, 5). All four are built and verified on their bases already; they are closed only to keep the diffs honest.

## What changes for a user

- `check_assembly` catches parts whose courtyards interpenetrate and reports them as a census; with `--baseline BOARD` it GATES the pairs the repair moved (absolute gating flips healthy corpus boards, so the gate is moved-vs-baseline). Edge-class waivers stand only for edge-live members; a locked part voids class waivers.
- Legality measures body CONTAINMENT: a part placed inside another part is NOT BUILDABLE, where before it passed every check. `quench` and `reconstruct` refuse to create such poses (the repair of an inherited one lands with the seeder PR, whose census it uses).
- Interior milled contours are visible to placement legality, with the ring's owner exempted so the part whose pads caused the cutout may sit on its own ring.

Closes #628.

## Why the two big fixtures

The tests grade real damaged geometry (the run-23 tigard board before and after repair, 26k lines each) rather than synthetic rectangles, because synthetic rectangles are what passed while a human rejected the real board at a glance. They are committed as bytes: regenerating the damaged one from the tracked source already produces a different file, and the repaired one is the output of actual placement work.

## Testing

Standalone on main: compiles; green: `test_run22_seat_containment_gate`, `test_run6_body_overlap`, `test_run6_check_assembly`, `test_run20_ring_ownership`, `test_outline_prefilter`, `test_part_class`, `test_457_determinism`. (`test_run6_body_overlap`'s corpus guard now asserts the 22 tracked boards, not a developed tree's generated surplus.)

### DIFF
--- a/kicad_files/watchy.kicad_pro
+++ b/kicad_files/watchy.kicad_pro
@@ -1,0 +1,645 @@
+{
+  "board": {
+    "3dviewports": [],
+    "design_settings": {
+      "defaults": {
+        "apply_defaults_to_fp_fields": false,
+        "apply_defaults_to_fp_shapes": false,
+        "apply_defaults_to_fp_text": false,
+        "board_outline_line_width": 0.15,
+        "copper_line_width": 0.2,
+        "copper_text_italic": false,
+        "copper_text_size_h": 1.5,
+        "copper_text_size_v": 1.5,
+        "copper_text_thickness": 0.3,
+        "copper_text_upright": false,
+        "courtyard_line_width": 0.05,
+        "dimension_precision": 4,
+        "dimension_units": 3,
+        "dimensions": {
+          "arrow_length": 1270000,
+          "extension_offset": 500000,
+          "keep_text_aligned": true,
+          "suppress_zeroes": false,
+          "text_position": 0,
+          "units_format": 1
+        },
+        "fab_line_width": 0.1,
+        "fab_text_italic": false,
+        "fab_text_size_h": 1.0,
+        "fab_text_size_v": 1.0,
+        "fab_text_thickness": 0.15,
+        "fab_text_upright": false,
+        "other_line_width": 0.1,
+        "other_text_italic": false,
+        "other_text_size_h": 1.0,
+        "other_text_size_v": 1.0,
+        "other_text_thickness": 0.15,
+        "other_text_upright": false,
+        "pads": {
+          "drill": 0.0,
+          "height": 1.0,
+          "width": 1.4
+        },
+        "silk_line_width": 0.15,
+        "silk_text_italic": false,
+        "silk_text_size_h": 1.0,
+        "silk_text_size_v": 1.0,
+        "silk_text_thickness": 0.15,
+        "silk_text_upright": false,
+        "zones": {
+          "45_degree_only": false,
+          "min_clearance": 0.2032
+        }
+      },
+      "diff_pair_dimensions": [
+        {
+          "gap": 0.0,
+          "via_gap": 0.0,
+          "width": 0.0
+        }
+      ],
+      "drc_exclusions": [],
+      "meta": {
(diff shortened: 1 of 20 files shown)


drandyhaas:
Merged — thank you. The measured, fixture-hunted test style here (the gate ON/OFF tables, and refusing fixtures that prove nothing) is exactly what we want; this went through review far faster for it.

One follow-up commit on main right after the merge: fa72000e removes `kicad_files/watchy.kicad_pro`. Nothing in this PR consumed it (verified by grep and by running all seven tests without it), and it looked like residue from the declined board-floors campaign — `kicad_files/` boards are deliberately project-less, and a sibling project there silently changes what future routing runs on that fixture auto-read. Measured before removal: both watchy routing tests produce byte-identical output with and without it, so nothing of this PR's substance was affected.

Per your stack plan, feel free to reopen the seeder PR (#682) rebased on current main whenever ready — and the standing invitation from #679 for the machine-readable output commit as a standalone PR remains open.

Issue #647, opened 2026-08-15
Plane fragility: 'NoneType' object has no attribute 'items' on every route after route_planes.py (degrades to zone outlines)


danwillis-aethl:
## Summary

Every `route.py` / `route_diff.py` invocation that follows `route_planes.py` logs:

```
Plane fragility: exact refill unavailable ('NoneType' object has no attribute 'items'); using zone outlines
```

An `AttributeError` is caught and the plane-fragility field silently degrades from KiCad's exact fill to plain zone outlines. It fired on **every** routing invocation in my run (24 occurrences across two chains).

## Environment

- KiCadRoutingTools **0.20.4**, `grid_router` **0.20.1** (prebuilt binary via `build_router.py`)
- KiCad **10.0.3**, Python **3.11.5**, Windows 11
- 4-layer board, 317 footprints, 802 nets. 11 zones: GND on `In1.Cu`, 4 power nets sharing `In2.Cu` (Voronoi), plus 2 pre-existing keepout rule areas.

## Reproduce

```bash
python py_router/route_planes.py in.kicad_pcb planes.kicad_pcb \
  --nets GND +3.3V +5V VLED VDD_CORE \
  --plane-layers In1.Cu In2.Cu In2.Cu In2.Cu In2.Cu

python py_router/route.py planes.kicad_pcb routed.kicad_pcb --nets "*"
```

The message appears right after `Net obstacle cache built`, immediately before
`Plane fragility field: <N> cells near the edges of 5 pour(s)`.

## Suspected cause

`route_planes.py` intentionally writes pours **unfilled** (the #562 bare-pour behaviour — output correctly notes *"Open in KiCad and press 'B' to refill zones"*). When `route.py` then asks pcbnew for the exact fill in order to build the fragility field, there is no fill data to read, and something downstream indexes `.items()` on the resulting `None`.

If that is right, the documented planes-then-route chain **always** lands in the degraded path, which I don't think is the intent.

## Impact

The fragility field is what steers the router away from necking the pours — in my run it covered 333,677 cells around 5 pours. Building it from zone *outlines* rather than the actual fill means the router's model of the pours is wrong at exactly the boundaries the feature exists to protect.

Functionally the run still completes; this is about the quality of that guidance, plus a swallowed exception that hides a real code path failure.

## Notes

Happy to supply the full 46k-line log, or a redacted board that reproduces it.

drandyhaas:
[...]

PR #681, opened 2026-08-21
Small fixes: Windows execv, .kicad_dru in sibling lists, manifest keeps --fab-tier/--fab-overrides

edgehero:
Four small fixes, split out of #679 after review.

- `headless_plan.py` used `os.execv`, which on Windows returns to the caller immediately; it now runs the child and exits with its code.
- Four hand-written board-sibling lists omitted `.kicad_dru` (`gui_utils`, `fixture_boards`, `redo_diff_stage`, and three tests); they use `copy_board.SIBLING_EXTS` now.
- `manifest_to_plan` had `--fab-tier` and `--fab-overrides` in none of its tables, so a recorded run replayed through the GUI at the default tier. Eight lines.
- A missing `import os` in `fix_mixed_net_refs.py` and one stale docstring.

Standalone on main: compiles, the parity and sibling tests green (`test_457_fresh_clone_fixtures` is red on main already, unrelated).

### DIFF
--- a/kicad_routing_plugin/gui_utils.py
+++ b/kicad_routing_plugin/gui_utils.py
@@ -855,8 +855,9 @@ def run_kicad_oracle_on_live_board(board, net_names, *, clearance,
                                     if hole_to_hole_clearance is not None
                                     else defaults.HOLE_TO_HOLE_CLEARANCE),
             progress_callback=progress_callback)
-        for _p in (tmp, os.path.splitext(tmp)[0] + '.kicad_pro',
-                   os.path.splitext(tmp)[0] + '.kicad_prl'):
+        from copy_board import SIBLING_EXTS as _sib_exts
+        for _p in (tmp,) + tuple(os.path.splitext(tmp)[0] + _e
+                                 for _e in _sib_exts):
             try:
                 os.unlink(_p)
             except OSError:
--- a/py_router/blocking_analysis.py
+++ b/py_router/blocking_analysis.py
@@ -919,7 +919,7 @@ def filter_rippable_blockers(
     With `pcb_data`, PROTECTED nets are also dropped (run-6 fix): this is the
     one choke point all six in-run rip ladders flow through, and none of them
     consulted protection_map -- so a .kicad_pro protected net, KiCad-locked
-    copper, or a --protect-nets match could be ripped mid-run by the phase-3
+    copper, or a protected net could be ripped mid-run by the phase-3
     cascade even though every PRE-run rip channel refused it (test-board run
     5, journal [10]: the #444 seam re-ask and the tap ladder churned a
     committed net's copper repeatedly). Refusals are recorded in
--- a/py_router/headless_plan.py
+++ b/py_router/headless_plan.py
@@ -72,7 +72,14 @@ def reexec_into_kicad(script=None, argv=None):
             continue
         if subprocess.run([cand, '-c', 'import pcbnew, wx'],
                           capture_output=True).returncode == 0:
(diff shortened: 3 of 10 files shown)


drandyhaas:
Reviewed all four pieces against current main. Verdicts:

- **Windows execv fix: taken.** Absorbed onto main as 9745cf50 (with co-author credit) — verified the CRT space-splitting mechanism is real and the posix path is untouched. It pairs with #647's Windows interpreter-discovery fix: discovery found the right python, then the exec tore its path apart.
- **Manifest `--fab-tier`/`--fab-overrides`: superseded.** Main already has this as ece1e011 (now pushed), which went further than the eight lines here — explicit mappings plus an `ai_plan` alias for previously-converted plans, gate expectation entries, and the gate's fixture manifest now always runs (no corpus manifest uses `--fab-overrides`, so corpus-only gate runs could never catch its loss). Please drop this hunk on rebase.
- **`.kicad_dru` sibling lists + the `import os`/stale-docstring fixes: look good**, still under review — likely accepted once the PR is rebased down to just them. The SIBLING_EXTS consolidation is exactly the drift-prevention shape we want (one list, imported everywhere).

So: rebase this PR to the dru-sibling + import/docstring changes and it should go quickly.

drandyhaas:
The `.kicad_dru` sibling-list fix is also taken onto main, as 72caae0a (co-author credit) — with one repair worth knowing about:

Your `redo_diff_stage.py` hunk did a function-local `from copy_board import SIBLING_EXTS`, but that script drives everything by subprocess and nothing ever puts `py_router/` on its sys.path — executing the module shows the import raises `ModuleNotFoundError` at the copy-out stage. No test reaches that line, which is why the branch's suite stayed green over it. On main the import is module-level behind an explicit path insert, and executing the module confirms it resolves. (The `gui_utils` local import is fine as you wrote it — swig_gui puts the engine dir on sys.path at plugin load — and `fixture_boards`' own insert was already correct.)

That leaves this PR with just the `import os` in `fix_mixed_net_refs.py` and the `blocking_analysis` stale-docstring line — both fine; feel free to rebase down to those two, or we can absorb them the same way and close this out. Everything is on pushed main once these land, so rebasing #679/#680 on current main picks all of it up.

drandyhaas:
Final two pieces judged, which completes this PR's review:

- **Stale docstring: taken** as e268146d — confirmed `--protect-nets` exists nowhere as a flag (#521 deliberately ships no such CLI flag or GUI control; the override is naming the net exactly).
- **`import os` in `fix_mixed_net_refs.py`: declined** — on current main that file contains zero references to `os`, so the import would land unused. Presumably something on your branch uses it; on main it's a fix for the fork's own history, the same class as the custom-pads and dup-ref-stdout bullets in the original #679.

Full disposition of this PR: execv fix → 9745cf50, dru sibling lists → 72caae0a (with the redo_diff_stage import repair noted above), manifest fab flags → superseded by ece1e011, import-os → declined as main-inapplicable. Everything taken carries your co-author credit. Nothing left to merge here, so this PR can be closed once you've confirmed — and thanks: three of the four were real improvements we now have.

drandyhaas:
Closing: review complete, everything here is resolved on main. Taken with your co-author credit — 9745cf50 (Windows execv), 72caae0a (dru sibling lists, with the redo_diff_stage import repair), e268146d (stale docstring). Superseded — the manifest fab-flag hunk (ece1e011). Declined — the import os (unused on main). All of it is on pushed origin/main now, so rebasing your other branches on current main picks everything up. Thanks Rob — three real improvements landed from this one.

PR #682, opened 2026-08-21
Seeder: eviction rung, and blockers named on a no-legal-pose verdict (closes #629, #630)

edgehero:
Seeder: eviction rung, and blockers named on a no-legal-pose verdict. Second in the placement stack, on top of #680 (its diff here includes #680's until that merges).

## What changes for a user

- A piled-up board is reachable when legal space exists: the seeder gains an eviction/restructuring rung (`place_lift`, eviction with ordering) instead of giving up.
- A no-legal-pose verdict and a park now NAME what is in the way, and a zero-pose census carries a near-miss report ("overlaps only U3 by 1.7 mm2, move that and the seat exists") instead of a bare refusal.
- `legalize` repairs a containment it did not create (the part-inside-a-part that the legality PR refuses to create), through the seeder's census.
- `place_seed --deadline` stops between parts and keeps its seats (through the routing PR's `krt_deadline`, included here as an identical copy so this branch runs alone).

Closes #629, closes #630.

## Testing

On top of the legality PR: compiles; green: `test_630_seeder_eviction`, `test_place_edge_containment`, `test_run22_containment_repair`, `test_portfolio_determinism`. `test_compare_seeds` self-budgets an hour and is not run here (it behaves the same on main).

PR #683, opened 2026-08-21
Review instruments: review sheet, defect panels, pocket census, connector class

edgehero:
Review instruments. Third in the placement stack, on top of #682 (its diff here includes #680 and #682 until those merge).

## What changes for a user

- `render_placement --review-sheet` composes one image for a human review: both sides, the courtyard census, and each connector's distance to its nearest edge. `--quiet` keeps the numbers off the terminal so the reviewer can look at the picture before reading them.
- Defect records (from `check_reachability --defect-json`) are drawn at a scale the measurement itself fixes, so a 0.04 mm shortfall is visible instead of a pixel.
- `check_pockets` reports routing demand against free area per window BEFORE routing (one census definition shared with the router's congestion cost), naming the pocket where nets will die.
- `connector_affinity` becomes an intent class with an interior setback.

`check_reachability`, `net_forensics` and `routability` are identical copies of the routing PR's modules (the defect-record channel runs through them); whichever merges second resolves clean.

## Testing

On top of the seeder PR: compiles; green: `test_run23_courtyard_channel`, `test_defect_render_scale`, `test_defect_record_reachability`, `test_render_view_and_drc_panels`, `test_run23_pockets`, `test_run23_connector_affinity`, `test_504_ratsnest_metrics`, `test_run23_reachability_widen`.

### DIFF
--- a/kicad_files/watchy.kicad_pro
+++ b/kicad_files/watchy.kicad_pro
@@ -1,0 +1,645 @@
+{
+  "board": {
+    "3dviewports": [],
+    "design_settings": {
+      "defaults": {
+        "apply_defaults_to_fp_fields": false,
+        "apply_defaults_to_fp_shapes": false,
+        "apply_defaults_to_fp_text": false,
+        "board_outline_line_width": 0.15,
+        "copper_line_width": 0.2,
+        "copper_text_italic": false,
+        "copper_text_size_h": 1.5,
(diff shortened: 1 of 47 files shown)


edgehero:
Stacked on #682 (and #680). Closing until those merge; it reopens then with only its own 16 files. Branch pr/placement-review-instruments stays ready.

Issue #411, opened 2026-07-15
placement: manufacture known-bad seeds by perturbing the corpus placements


edgehero:
#110 asks for two things that the current corpus can't supply:

> Run it across a corpus of real boards ... covering a range of sizes, densities, and **placement quality**.

> **Characterize when it helps.** Document which board/placement regimes benefit (**bad seeds**) vs. where it's neutral or harmful (good hand placements)

The stress harness (#117) downloads real boards, strips tracks/vias/pours, and re-routes. It **doesn't strip the placement**. So every board in the corpus arrives with a floorplan that a human already iterated until the board routed end to end and shipped. There are no bad seeds in it to characterise, which makes #110's acceptance criteria hard to meet from the corpus as it stands.

**Proposal:** add a perturb step, and get bad seeds with free ground truth.

1. Take a corpus board at its original (good) placement.
2. **Perturb the placement** by a controlled amount: displace whole blocks or subsystems by 5 / 10 / 20 / 40 / 80 mm, and/or move a subsystem to the wrong side of the board.
3. Run `place_route_loop.py` on the perturbed board.
4. **Grade the recovered placement against the original.** The original placement is free ground truth, exactly the way the original routing already is for routing.

That gives a **dose response curve on displacement** and a direct measurement of the loop's recovery radius: how far can a block be wrong before the loop can no longer bring it back?

It also lets the existing grader do the work, since "did routing complete, with how many vias, at what router effort" is already implemented, and now has a known-correct placement to compare against.

**A prediction, from n=1, that I would genuinely like to be wrong about:** at block scale (a 40+ pin part displaced tens of mm) `--max-target-pins 40` recovers none of them, because the part that needs to move is never a passive. If that holds it measures the gap precisely. If it doesn't hold, I've learned my own stop-loss is unnecessary and I'll take that trade happily.

Background and the two real-board failures that motivated this are in #407.

edgehero:
**Assessment vs. the OmniEDA benchmarks ([omnieda.com](https://www.omnieda.com/)) — this issue is one of the two biggest beneficiaries.**

[...]

PR #684, opened 2026-08-21
Recovery-run harness and loop gates (serves #411)

edgehero:
The recovery-run harness and loop gates. Fourth in the placement stack, on top of #683 (its diff here includes the three below it until they merge). This is the part that serves #411 (manufacture known-bad placements by perturbing the corpus, then measure recovery) and the agent loop around it; say so if you would rather it lived in the fork.

## What is in it

- Staging with the control board fenced OUTSIDE the work dir and a content-based leak audit: a pose record is caught by its content whatever it is named.
- `tee_cmd`: per-command timing ledger and a `.done` completion marker that is last-attempt-wins (absent while a retry runs).
- `run_watch` and a lever-provenance audit: every tool that writes poses declares itself, so a hand-written pose can be told from an engine's.
- Loop driver gates: the handoff and the close refuse until someone has LOOKED at a render and written observations before reading the numbers; `RUN_STATE.json` as the machine-readable NOW view. (Upstream already has `converge --exhausted`; this adds only RUN_STATE.) `check_complete` prints the components it found NOT APPLICABLE instead of folding them into the verdict silently.

`routing_diagnostics` is an identical copy of the routing PR's module; one test here reads its verdict.

## Testing

On top of the instruments PR: compiles; `loop_driver --self-test` green; green: `test_431_board_gates`, `test_run23_tee_done`, `test_run23_run_state`, `test_run20_not_applicable`, `test_provenance_audit`, `test_board_score_vacuity`, `test_run20_oob_channel`, `test_run23_budget_honesty`, `test_run22_driver_workdir`, `test_run21_watcher_self_scan`, `test_run8_gate_conjuncts`, `test_doc_flag_liveness`, `test_run20_run_watch`.

### DIFF
--- a/.claude/skills/plan-pcb-placement-and-routing/SKILL.md
+++ b/.claude/skills/plan-pcb-placement-and-routing/SKILL.md
@@ -145,6 +145,23 @@ the failure, and without `--accept-cmd` its comparator cannot see the thing you
 are trying to fix (a length, a width, a clause). An ACCEPTED round is not a
 verdict -- grade the output with the same battery either skill would use.
 
+## Eyes at the boundaries (run-23)
+
+Numbers gate legality; nothing gated LOOKING, and run 23 shipped a board a
+human rejected at a glance (15 courtyard interpenetrations, four mid-board
+connectors) while every key read clean — its orchestrator viewed ONE image in
+4.7 hours, after the failure. Two driver gates now refuse without a written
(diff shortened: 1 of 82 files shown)


edgehero:
Stacked on #683. Closing until the stack below merges; it reopens then with only its own 35 files. Branch pr/placement-run-harness stays ready.

PR #685, opened 2026-08-21
Skills: the routing doctrine from runs 20-24 (docs, plus the tests that pin its phrases)

edgehero:
Docs only: the routing skill's doctrine from runs 20 to 24, and the placement write-up. Last in the placement stack, on top of #684 (its diff here includes the four below it until they merge); the drivers' tests pin several of these phrases.

Checked against 76fc1eef: no corpus board names in user-facing guidance, and no reference to the declined floors mechanism.

## Testing

Green on top of the harness PR: `test_run15_handback_contract`, `test_run20_defect_reentry`. Two tests here, `test_run20_boxin_lever_guard` and `test_run20_tooling_gaps`, also check summary keys that #679 adds (`boxed_in` merged across summaries, `finalize_excluded_nets`) and go fully green once #679 lands; everything else in them passes now.


edgehero:
Stacked on #684. Closing until the stack below merges; it reopens then with only its own 5 files. Branch pr/skill-docs stays ready.

PR #690, opened 2026-08-22
review instruments: the review sheet, defect panels, pocket census, connector class (replaces #683)

edgehero:
Review instruments, rebuilt on current main. This replaces #683 (closed while it was stacked on #680 and #682). It now carries only its own files on top of the merged legality (#680), run-output (#686) and seeder (#687) work, so every hunk here is new against main. The reachability instrument that writes defect records is a separate PR; this one reads a plain JSON file and imports nothing from it.

## What changes for a user

- `render_placement --review-sheet PATH` writes one composite image for the boundary review: the F and B panels side by side over a facts strip carrying the courtyard census, every connector-family part's distance to its nearest edge (INTERIOR flagged past 3 mm), and the front/back stacks that look like collisions in the panels but sit on opposite faces. Long lines wrap at the `  |  ` separator so a ten-pair blocking list stays on the sheet. With `--pair` the sheet composes the AFTER panels, which is the board the facts strip describes. The path it wrote is echoed back as `review_sheet` in the JSON, or `null` when the compose failed.
- The courtyard channel reaches the render. Courtyard pairs come from `legality.grade_body_overlap`, the same function `check_assembly` calls, and are drawn as a filled intersection with the pair named on the image. The JSON checklist gains `b_courtyard_advisory_pairs`, `b_courtyard_blocking_pairs`, `b_courtyard_overlap_mm2`, `b_courtyard_census_error` and `b_cross_side_stacks`. `b_body_overlap_pairs` was pad intersections all along, which is how a board with one part 0.90 mm inside another's courtyard rendered "clean". With `--gate` and `--before`, a blocking pair with a moved member fails the gate; without `--before` the census is report-only.
- `render_placement --defect-json PATH` (repeatable) draws defect panels. A defect record is a plain JSON file; the schema is documented on `load_defect_records`, which is the consumer contract and now enforces it: a document that is not a JSON object, or whose `version` is not one this reader knows, is refused with a note instead of read hopefully. Each defect gets its own panel, cropped so the measured shortfall spans at least 16 px (`DEFECT_MIN_PX`), the measured gap drawn solid and the required gap dashed alongside it. Records whose `board_sha` does not match the board are skipped with a note, and every note is carried in the document as `instrument.defect_notes`. The caption and `defect_panels[].shortfall_px` report the scale actually achieved and say INVISIBLE AT THIS SCALE when it is not enough. Giving `--defect-json` SUPPRESSES the `--focus` legality-pocket panels (a measured coordinate beats a cluster derived from pad positions); `--focus` panels for failed nets, which need `--summary-json`, are unaffected. With `--gate`, a defect render now also FAILS (exit 4) when a record was refused or cannot be proven to describe this board -- including one that merely lacks `board_sha` -- when a defect produced no panel, or when a panel could not reach `DEFECT_MIN_PX`. That is a new verdict channel, and it fixes the contract before its writer exists: a gate asked to judge a defect render should not pass one that could not show the defect. Nothing in this tree writes such a record yet; the reachability PR does.
- `render_placement --side F|B` renders one face only. `--quiet` together with `--json-out` now also suppresses the stdout JSON_SUMMARY echo and the describe text (the file carries both), so a reviewer can look at the image before reading the keys; without `--json-out` the summary still prints.
- `check_pockets.py` is new: routing demand against free copper area per window, before routing. The board is binned at `--bin` mm, each bin's distinct demanding nets are set against its free area, and the top windows are printed with their nets and parts, plus the lane arithmetic (track + 2 x clearance at the board's own floors). Report-only, exit 0 always; there is no threshold by default because the metric is too young to own one. `--bin` is floored at 0.25 mm, which is the census's own floor, and the effective value is announced and recorded rather than silently accepted. It uses the router's own census: `congestion_field.build_congestion2` is split so that `congestion_bins` holds the bin computation and both consumers share one definition. The A* cost path is unchanged; a dead loop in the moved body (`for p in pads: pass`) is dropped.
- `connector_affinity` becomes an intent class. `check_floorplan --emit-intent --declare-classes` records connector-family parts that claim no edge with `"class": "connector_affinity"` and no `edge`, and the `edge_connector` rule flags such a part seated more than 3 mm from every edge at warn only, whatever the configured severity; `pass` never flips on it. An author upgrades the entry by writing `max_setback_mm` or `edge`.
- `--emit-intent` withholds the `overlap_area` budget when the emitting board carries unwaived courtyard interpenetrations past the blocking floors (it already did for blocking body pairs), and `context.budget_withheld` names what was withheld and why. A withheld key is now ABSTAINED at grade time rather than silently ungraded: `check_floorplan --intent` prints `N legality budget key(s) NOT DERIVABLE -- not graded, not passed` and carries `budget_abstained` in `--json` and `budget_abstained` / `budget_abstained_keys` in JSON_SUMMARY.

There is no wall-clock budget in any engine or tool path. (`tests/test_defect_render_scale.py` passes `timeout=900` to `subprocess.run`, which is a test-harness guard against a hung child, not a budget any shipped code observes.)

## Two populations, named

Three courtyard numbers reach the render and they are not nested, so each key names its own population. Measured on `tests/fixtures/run23/tigard_placed.kicad_pcb`: `b_courtyard_overlap_mm2` = 26.98, summing EVERY courtyard pair including waived ones; `b_courtyard_advisory_pairs` = 5, the unwaived ones; `b_courtyard_blocking_pairs` = 10, the ones past the blocking floors, of which 6 are absent from the advisory list because a waiver-voided pair (dead edge waiver, locked member) blocks while staying labelled waived. The first of those keys was called `b_courtyard_overlap_pairs`, which invited a reader to treat the advisory list as a superset. The sheet's headline now says which population each number is.

`edge_connectors` in an emitted intent also holds two populations, and that one is a real hazard rather than a naming problem. An `edge_receptacle` entry is a seat claim the placement engines act on: `place_seed` LOCKS those refs for its polish quench, `place_reconstruct` and `seeder.reseat_scope` give each a banded off-outline allowance, `place_reconstruct` excludes them from the exchange stage, `reconstruct.classify` forces them into the anchor tier, and `seeder.repair_placement` routes them to an edge seat instead of the ordinary repair loop. A `connector_affinity` entry is a declaration with no claim. `Intent.edge_claims()` is the split, in one place rather than as a filter repeated at eleven call sites, and it is what every engine read goes through; the `edge_connector` rule reads the whole key, because flagging an interior pose is the entry's only purpose. On tigard_placed the emitted `edge_connectors` is 7 entries and `edge_claims()` is 1, which is exactly what upstream main emits.

The `repair_placement` site was a measured regression, caught in review and fixed here rather than shipped. Its dispatch refuses every entry in its edge map that carries no `edge`, before the ordinary `_try_place` loop, so reading the raw key took the declarations out of repair entirely. On `tests/fixtures/run23/tigard_damaged.kicad_pcb` with its own `--declare-classes` intent, J5, J6 and J7 came back as "declared edge part misplaced, but no edge is declared" -- a label they had not earned -- where upstream main tries them like any other violator and reports "no legal pose within any cap". With the split in place the run is note-for-note identical to main's (29 notes each) and the output board is pose-identical. One raw read is left in `seeder.reseat_scope`, marked `edge-claims-exempt`, because the sub-intent it builds must stay a faithful copy of the declaration list; the source guard in the tests accepts exactly that one and fails on any other.

## What the render's census is NOT

It calls the same function as `check_assembly`, and it is not the same census. Two differences, in the code comments and here:

- `check_assembly` passes `intent_waivers`; `render_placement` reads no intent, so an operator-waived pair is unwaived here and can reach `--gate`.
- `moved_parts` (the gate's currency) compares position and rotation over refs present in BOTH boards. `check_assembly`'s moved set also counts refs the baseline does not have, and side flips. On a board whose parts were added, removed or flipped the two lists differ.

Where they disagree, `check_assembly` is the authority. `--gate` here is a render's own verdict about its own picture.

`connector_edge_facts`, which feeds the sheet's connector row, is a footprint-NAME heuristic (`conn`, `header`, `jst`, `usb`, ...) plus the two `part_class` edge classes. It is deliberately broader than `part_class`'s gating classes and it is presentation only: nothing gates on it, and a connector whose footprint is named otherwise is missing from that row. The sheet says so on its face.

## Files

- `py_tools/render_placement.py`: courtyard, cross-side and census-error keys in `legality_findings`; `load_defect_records` (with the schema refusals), `defect_view`, `draw_defects`, `DEFECT_MIN_PX`, `DEFECT_SCHEMA_VERSIONS`; the courtyard overlay in `draw_legality` and its legend row; `connector_edge_facts` and `write_review_sheet`; the `--defect-json`, `--side`, `--review-sheet` flags and the `--quiet` change; the defect panels, the `defects` / `defect_panels` / `instrument.defect_json` / `instrument.defect_notes` / `review_sheet` keys, the review-sheet write and the moved-member courtyard gate in `main`.
- `py_tools/check_pockets.py`: new.
- `py_router/congestion_field.py`: `congestion_bins` extracted from `build_congestion2` so the pocket census and the router's A* cost share one definition, returning the bin it actually used alongside the bins. The extraction is behaviour preserving: the `max(0.25, bin)` and `max(1, layers)` clamps moved inside the new function and the caller still applies its own, and the env gate stays in `build_congestion2`. It also drops a `for p in pads: pass` no-op loop that folded nothing (the per-net pad loop below it does the work).
- `py_placer/placement/floorplan.py`: the `connector_affinity` branch of `rule_edge_connector`, its declaration in `emit_intent`, `Intent.edge_claims()`, `Intent.budget_withheld` and the `budget_abstained` reporting, and the courtyard budget withholding.
- `py_placer/place_seed.py`, `py_placer/place_reconstruct.py`, `py_placer/placement/reconstruct.py`, `py_placer/placement/seeder.py`: the engine reads go through `edge_claims()`.
- `docs/floorplan-intent.md`: the rule row, the class declaration, the two populations, and the withheld budget (`context.budget_withheld`, and what the grade does with it).
- Tests: `tests/test_run23_courtyard_channel.py`, `tests/test_defect_render_scale.py`, `tests/test_run23_pockets.py`, `tests/test_run23_connector_affinity.py` (new); `tests/test_render_view_and_drc_panels.py` gains the `--quiet` test.

Scope note: `test_run23_courtyard_channel.py` also pins `check_assembly`'s moved-vs-baseline courtyard gate and the courtyard floors, which #680 merged without a test; that code is not touched here.

A point in this PR's favour that is easy to miss: `upstream/main:tests/test_part_class.py:178` already asserts `by_ref['J7']['class'] == 'connector_affinity'` against an EMITTED intent. It is green on main only because its fixture (`wk/b2/tigard__swap/d0/perturbed.kicad_pcb`) is untracked and the test skips. Staged with that fixture present, main fails it with `KeyError: 'J7'` and this branch passes it. The emitter this PR adds is the one main's test was written for.

## Testing

Against upstream main at 0ba0226f, each run in the foreground with `python3 -X utf8`:

- `tests/test_defect_render_scale.py`: 38 passed, 0 failed.
- `tests/test_render_view_and_drc_panels.py`: ALL PASS (includes the new `--quiet` test).
- `tests/test_run23_pockets.py`: 6 tests OK.
- `tests/test_run23_connector_affinity.py`: 16 tests OK.
- `tests/test_run23_courtyard_channel.py`: 28 tests OK.
- `tests/test_504_ratsnest_metrics.py`: ALL PASS (unchanged; #683 listed it, and it still passes on its upstream version, so its one-line sys.path edit from #683 is not carried).
- `tests/test_run6_check_assembly.py`: OK.
- `tests/gui_parity/test_manifest_plan_parity.py` and `tests/gui_parity/test_cli_postpass_coverage.py`: 0 mismatches, 0 failures. `render_placement.py` stays in the converter's refused-tools map (it changes no board); `check_pockets.py` is a read-only checker like the other `check_*` tools, which the map does not list.

The tests use tracked boards only (`tests/fixtures/run23/`, `kicad_files/splitflap_driver.kicad_pcb`, `kicad_files/ulx3s.kicad_pcb`). No placement objective term changes, so `tests/test_placement_ab.py` is not affected: `edge_claims()` restores the pre-existing seeding and reconstruction inputs rather than altering them. That is asserted through the real call paths, not re-derived in the test: `TestSeederDrivesTheSplitForReal` calls `seeder.repair_placement` and `seeder.reseat_scope` with the full intent and with the same intent stripped of its `connector_affinity` entries and requires every output to match, `TestPlaceSeedLocksOnlyClaims` drives `place_seed.main()` and captures the `lock_refs` the polish quench is actually called with, and `TestInertToThePlacementEngines` calls `reconstruct.classify` on both intents and compares the tiers.

### DIFF
--- a/docs/floorplan-intent.md
+++ b/docs/floorplan-intent.md
@@ -129,7 +129,7 @@ for it, and the reason is printed:
 | `zone_side` | a member is on the other face | `legality.footprint_side` |
 | `zone_exclusive` | a non-member intrudes on a reserved zone | `rect_overlap_area` |
 | `keepout` | any part enters a keep-out, unless in `allow` | courtyard **and** through-hole rect |
-| `edge_connector` | overhang outside `[min,max]`, or the wrong edge | `BoardOutlineGate.rect_outside_amount` |
+| `edge_connector` | overhang outside `[min,max]`, or the wrong edge; a `connector_affinity` entry seated more than 3 mm from every edge fires at **warn** whatever the configured severity | `BoardOutlineGate.rect_outside_amount`, `edge_clearance` |
 | `decap_distance` | a decoupling cap is too far from its own IC | `groups.decap_tethers` |
 | `must_lock` | a declared-critical part is not locked in the file | `parser.extract_locked_refs` |
 | `legality` | overlap or off-board parts exceed a budget | `QuenchState.legality_metrics` |
@@ -324,3 +324,39 @@ This is the same spatial incoherence that makes sheet blocks useless for
 Parts already overhanging the outline are recorded as `edge_connectors` by
 observation, which is what stops `oob_count` reporting a card edge or USB shell
 as a defect forever.
+
+With `--declare-classes`, connector-family parts that claim no edge (headers,
+JST, terminal blocks; `part_class` calls them `connector_affinity`) are also
+recorded, with `"class": "connector_affinity"` and **no `edge`** (naming one
+would be an invention). The grade then flags such a part seated more than
+3 mm from every edge at `warn` only, because legitimately interior connectors
+exist; write `max_setback_mm` or `edge` on the entry to make it a real claim
+at the configured severity.
+
+These entries are **declarations, not seat claims**, and the placement engines
+do not act on them. `edge_connectors` therefore holds two populations, and
+`Intent.edge_claims()` is the split: it drops `connector_affinity` and is what
+`place_seed` (which LOCKS edge refs for its polish quench), `place_reconstruct`
+(banded off-outline allowance, exchange-stage exclusion) and
+`reconstruct.classify` (the anchor tier) read. The `edge_connector` **rule**
+reads the whole key, because flagging an interior pose is the entry's only
+purpose. Writing `max_setback_mm` or `edge` on an entry changes its class-based
+severity, not its membership -- to hand a part the edge-part treatment in the
+engines, declare it with an edge class.
+
+The `overlap_area` budget is **withheld** when the emitting board carries a
+blocking body pair, or an unwaived courtyard interpenetration past the
+blocking floors: baking the number would bless the board it was measured on.
+`context.budget_withheld` names each withheld key and why, so a reader can tell
+"withheld" from "forgot"; declare the budget by hand if the overlap is by
+design.
+
+A withheld key is **abstained at grade time, not passed**. `check_floorplan
+--intent` reads `context.budget_withheld`, reports every withheld key that the
+intent does not also declare under `N legality budget key(s) NOT DERIVABLE --
+not graded, not passed`, and carries them as `budget_abstained` in `--json` and
+as `budget_abstained` / `budget_abstained_keys` in `JSON_SUMMARY`. When the
+whole budget is empty the `legality` rule does not run at all, and its skip
+reason names the withholding rather than saying only "the intent declares no
+legality_budget". Declaring the key by hand overrides the note: a declared
+budget is graded.
--- a/py_placer/place_reconstruct.py
+++ b/py_placer/place_reconstruct.py
@@ -219,7 +219,11 @@ Examples:
     # set is part of the run's record.
     edge_bands = {}
     if intent is not None:
-        for c in intent.edge_connectors:
+        # edge_claims(): a connector_affinity entry declares a class, not a
+        # band. Reading the raw key gave every generic header the 2.0mm
+        # default allowance below, which is an off-outline licence nobody
+        # asked for.
+        for c in intent.edge_claims():
             if c['ref'] not in state.parts:
                 continue
             # EVERY banded entry keeps its allowance, suspect or not: with
@@ -317,7 +321,10 @@ Examples:
     # proxy would anchor them to their possibly-misplaced partners).
     edge_pref = {}
     if intent is not None:
-        for c in intent.edge_connectors:
+        # edge_claims(): the receptacle filter below already excluded the
+        # weak class, but every engine read goes through the one split so a
+        # future edit to this filter cannot re-open it.
+        for c in intent.edge_claims():
             # Receptacles ONLY (run-5): the edge metric is the right
             # objective for a part whose class says the mating face must
             # reach the edge. An edge-less ACTUATOR entry (a suspect
@@ -341,7 +348,7 @@ Examples:
     # under investigation, not a seat.
     excl_x = set(tiers.locked) | set(tiers.zero_net)
     if intent is not None:
-        for c in intent.edge_connectors:
+        for c in intent.edge_claims():   # seat claims only; see edge_claims
             ref = c['ref']
             if ref not in state.parts:
                 continue
--- a/py_placer/place_seed.py
+++ b/py_placer/place_seed.py
@@ -426,7 +426,9 @@ Examples:
         # polished by the objective the later steps rank with. Locks ride in
         # from the file (must_lock was just stamped); edge connectors are
         # locked per-call so the polish cannot walk them off their band.
-        edge_refs = [c['ref'] for c in intent.edge_connectors]
+        # edge_claims(), not edge_connectors: a connector_affinity entry
+        # makes no seat claim and must not be locked out of the polish.
+        edge_refs = [c['ref'] for c in intent.edge_claims()]
         placements = quench(
             pcb_seeded, pcb_file=args.output_file,
             max_displacement=args.max_displacement,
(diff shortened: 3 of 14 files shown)


edgehero:
One thing worth knowing while reviewing this: upstream/main already carries `tests/test_part_class.py` asserting `by_ref['J7']['class'] == 'connector_affinity'` on an emitted intent, which main's own `emit_intent` cannot produce (it skips the class). That test is green today only because its fixture is untracked, so it skips on a clean checkout. This PR is what makes it real, which is also why the entries stay in `edge_connectors` on the wire rather than moving to a separate key.

PR #692, opened 2026-08-22
skills: the routing doctrine from runs 20-24, bounded deterministically

edgehero:
# The budget doctrine, rebuilt on determinism

Replaces #685, rebuilt on current main (80e38e5e, on top of #686, #687, #689 and
#690). One commit, 9 files, +972 / -25. 259 of those lines are the routing skill;
623 are the two test files.

## Why it exists

#685 built its "how to run a long step" advice on a `--deadline` flag. That flag
was removed from every tool by #621, so the advice told readers to pass something
that is now an argparse error. The reason for removing it is the right one: the
routing has to be deterministic, and a clock breaks that across machines.

So the doctrine is rebuilt on that premise rather than cut out. Ground truth
first: `--deadline` appears in no argparse call anywhere on main. Every mention
of it in this diff is a statement that it is gone.

## What the doctrine says

In `.claude/skills/plan-pcb-routing/SKILL.md` and in `loop_driver.py`'s teammate
prompts:

- No main takes a wall-clock budget, and that is deliberate. The same board with
  the same arguments has to produce the same copper on a slow machine and a fast
  one.
- A shell `timeout` is not a substitute. It SIGTERMs the tool (on Windows
  `TerminateProcess`, which no handler, `atexit` or flush can catch), so you lose
  the partial board and read the shell's 124 or 143 instead of anything the run
  computed. There is no summary line left to parse.
- Bound the work by scope, with the caps the tools actually have, every one of
  them a count or a set: `route.py --nets`, `--max-ripup`, named refs on
  `place_seed.py --reseat`, `--evict-depth 0` (already the default), and
  `converge.py verdict --flat N` / `record --exhausted <half>`.
- Run the long step detached and read its log.

The teaching that motivated the old passage is kept, because it describes real
behaviour: `EXACT_FILL_TIMEOUT` (300 s) and `ORACLE_DRC_TIMEOUT` (240 s) both
exist, both degrade to a fallback with no failure signal and no effect on the
exit code, and the oracle one memoises the board so every later step skips it
too. Both numbers are now re-derived from their modules by a test.

## How to read a run, which nothing documented

#686 added a one-line `JSON_SUMMARY_MIN` verdict and nothing said how to read it.
The skill now does, with the four things that are easy to get wrong:

- **It says nothing about whether the DRC floors held.** The `.kicad_pro`
  writeback runs after this line prints and reports on its own, so a run whose
  copper is below its own declared floor still prints a clean MIN line.
  `route_summary.py` says so at the function that builds it. Given the fab-floor
  ratchet, "read this one line" without that caveat would be the defect this PR
  exists to prevent.
- `main_loop_time_s` counts the single-ended loop plus the reroute loop and
  nothing else; read as a run duration it is wrong by more than an order. Stated
  as a ratio, deliberately: an absolute second-count re-derives differently on
  every machine, which would refute the passage it sits in.
- `status` appears only on a run that legitimately did nothing, and takes exactly
  `no_valid_nets` and `already_connected`.
- `complete` means "a sub-run did not finish", never "a budget expired".
  `merge_summaries` forces it onto a merged summary when any part carries it and
  `place_route_loop` refuses one.

Plus the verification rules from runs 20 to 24: read both failure buckets from
every summary line (a call printed `routed_single: ["QSPI_SD1"], failed_single:
[]` and wrote a board with zero segments on that net, because the reconciliation
had reported it under `failed_multipoint`); read
`components.unrouted.placement_blocked` rather than working the on-board-pad
denominator out by hand; and the five-row table of checks that reported "the
guard held" when it had not.

## What it cuts

`plan-pcb-routing` has no loop, no ledger and no teammates, so three sections
that near-duplicated the combined skill are now one-line pointers into it (9.3a,
9.3b, 9.3d, 9.5), keeping only what the sibling does not carry: the box-in row,
the already-connected-net no-op, and the hint's `(current: ...)` tail. The
delegation doctrine moved out entirely, into `loop_driver.py`, which is the thing
that has teammates. The duplicated `complete`/`status`/`main_loop_time_s` notes
collapsed to one copy each.

Deliberately not included, and available separately: the L4 box-in guard and
everything built on `boxed_in[].at_floor` (a key no main emits and that was cut
from #686 in review, so a guard reading it can never fire); the L3/L4
defect-record re-entry machinery, the review gates, `--workdir` retargeting and
the `--status` view in `loop_driver.py`; `board_score.py`'s `not_applicable`
split and its identity-line scope suffix; and about 900 lines of placement
doctrine #685 had put into the routing skill.

`board_score.py` is untouched. An earlier revision of this branch added
`--fab-tier` / `--fab-overrides` to it and described that as a bug fix; it was
not one -- the flags never existed on `board_score` -- so it is dropped rather than
re-labelled. It should come back as its own change, with a test.

## Corrections to text already on main

- `--protect-nets` was removed in 53a5a16e; the combined skill still told the
  reader to pass it.
- `--track-width-floor` was removed in the same commit; the combined skill still
  told the reader to carry it on every retry.
- The `CMD:` list still named `route_disconnected_planes.py`, which is now
  `repair_planes.py`.
- "No width FLOOR exists on any tool ... no floor protection at all" was wrong on
  both halves: `_neck_pair_partner_grazes` floors necking at the fab minimum, and
  `route_diff.py` never carried `--track-width-floor`. The passage now says the
  spec width is protected down to the fab floor and no further.
- `docs/api-routing-config.md` still documented `track_width_floor` as a live
  config field with `--track-width-floor` as its flag. Neither exists in any
  python source. The gate lists that doc, and passed it only because the flag
  string survives in `krt_capabilities.py`; the harness PR removes that literal,
  which would have turned this gate red on a branch that had nothing to do with
  it. Fixed here, and verified by simulating the removal.
- `placement_driver.py`'s `REFS` pointed at
  `.claude/skills/plan-pcb-routing/references`, which does not exist, so every
  stage that cited a reference page cited a missing file.

## Tests

    tests/test_doc_flag_liveness.py           exit 0    OK   (11 checks)
    tests/test_run15_handback_contract.py     exit 0    Ran 19 tests, OK, 0 skipped

and the five skill/doc gates already on main, all exit 0:
`test_run8_skills_generic.py`, `test_431_skill_commands.py`,
`test_refusal_checks.py`, `test_run8_skill_drivers.py`,
`test_run8_skill_instruction_quality.py`.

`test_doc_flag_liveness.py` comes INTO this PR, because the audit above is only
worth what the gate behind it is worth, and that gate was on no branch. It holds
both skills and `docs/api-routing-config.md` to the engine's actual surface: any
`--flag` a doc tells you to pass must be registered somewhere in non-test source,
and any `x.py` a doc tells you to run must be a file the repo ships. It is
union-shaped on purpose (a per-tool gate drowns in false positives, and a gate
that cries wolf gets deleted), and a sentence saying a flag is REMOVED is
exempt -- otherwise the honest correction would fail it.

What the cover does and does not do, measured rather than asserted. It catches
DELETION and dead identifiers; it does not catch arbitrary REWRITING. Of the 259
lines this change adds to the routing skill: 19 passages are enumerated by name
and all 19 match; counting only lines that literally contain an anchor, 18 of 259
carry one and the longest run without one is 39; widening to every `assertIn`
literal in the test, 32 of 259 and a longest run of 22. Those are the honest
numbers -- an earlier revision of this body said 39 and 31, which do not
reproduce.

What makes the difference is not the string count but the five tests that
RE-DERIVE from source rather than matching prose, and they are the best thing
here:

- `EXACT_FILL_TIMEOUT` and `ORACLE_DRC_TIMEOUT` are read out of their modules and
  compared against the table row.
- The `JSON_SUMMARY_MIN` key list is compared against what
  `route_summary.summary_min` actually returns -- both directions, so a key the
  doc omits and a key the doc invents both fail.
- `finalize_excluded_nets`, `merged['complete'] = False` and the `{RED} Plane
  finalize` line are checked against `route.py` and `route_summary.py`.
- `--evict-depth`'s default is re-derived from `place_seed.py`'s argparse, so
  "(the default)" cannot be inverted.
- `placement_blocked` is checked against `board_score.py`, and the sentence that
  QUOTES 9.3d is checked against 9.3d.

An earlier revision of this branch reported eight passing mutations. That was
mutation-testing the test against itself: the eight were exactly the eight
strings it hard-coded. Two rounds of review then showed that deleting five whole
passages, falsifying either timeout constant, and renaming `repair_planes.py`
back to the deleted `route_disconnected_planes.py` all left it green -- and after
that was fixed, that five REWRITES still did. Twenty-four mutations now, all
caught:

    RW-1   rename two real JSON_SUMMARY_MIN keys        CAUGHT (handback)
    RW-2   falsify the 147/149 measurement              CAUGHT (handback)
    RW-3   invert "(the default)" on --evict-depth      CAUGHT (handback)
    RW-4   swap .kicad_dru for .kicad_pro               CAUGHT (handback)
    RW-5   invert "the writeback runs AFTER"            CAUGHT (handback)
    REV-1  delete the retry section                     CAUGHT (handback)
    REV-2  delete the stop conditions                   CAUGHT (handback)
    REV-3  delete the instruments table                 CAUGHT (handback)
    REV-4  delete the on-board-pads bullet              CAUGHT (handback)
    REV-5  delete the QSPI_SD1 bullet                   CAUGHT (handback)
    REV-6  delete the writeback caveat                  CAUGHT (handback)
    REV-7  falsify EXACT_FILL_TIMEOUT 300 -> 900        CAUGHT (handback)
    REV-8  falsify ORACLE_DRC_TIMEOUT 240 -> 640        CAUGHT (handback)
    REV-9  rename repair_planes.py back                 CAUGHT (both)
    REV-9b the same rename in the sibling skill only    CAUGHT (liveness only)
    NEW-1  delete the protected_nets passage            CAUGHT (handback)
    MINE-1 delete the determinism rationale             CAUGHT (handback)
    MINE-2 reword why a shell timeout does not help     CAUGHT (handback)
    MINE-3 drop --evict-depth 0 from the caps table     CAUGHT (handback)
    MINE-4 put at_floor back into the box-in row        CAUGHT (handback)
    MINE-5 describe `complete` as a budget signal       CAUGHT (handback)
    MINE-6 put --deadline back into loop_driver         CAUGHT (handback)
    MINE-7 break the hand-back contract                 CAUGHT (handback)
    MINE-8 rename summary['finalize_excluded_nets']     CAUGHT (handback)

REV-9b is why the gate had to come in: nothing else in the repo would have seen
it. It is also the one row the handback test does not catch, and that is correct
-- a stale script name in the sibling skill is the liveness gate's job.

## Two things I did not change, for you to rule on

The liveness gate's `EXTERNAL` allowlist -- the flags it treats as foreign
because worked examples quote `git`, `kicad-cli` and `pytest` -- shadows 19 names
that are ALSO real repo flags: `--dry-run`, `--output`, `--json`, `--quiet`,
`--force`, `--all`, `--verbose`, `--from-source`, `--version`, `--help` and
others. So if one of those is ever removed from a tool, the docs go unchecked for
it. Narrowing it properly means making the gate tool-aware, which is the design
it deliberately rejected (a per-tool gate drowns in false positives and gets
deleted). I would keep the allowlist as is and treat the narrowing as a
follow-up, because the class of flag it shadows -- generic verbs shared with
shell tools -- is not the class that gets removed. Say the word if you want it
narrowed now.

`tests/run_doc_examples.py` fails 4 blocks at this head, in
`docs/api-net-analysis.md` and `docs/api-impedance.md`. Both files are outside
this diff and both fail identically without it (a missing
`kicad_files/test_diffpair_ram.kicad_pcb` fixture), so it is pre-existing, not a
regression from the `api-routing-config.md` edit.

## Audit

Every `--flag`, env var, JSON key, exit code and file path in the added lines
checked against main, twice -- globally, and scoped to the tool named on each
emitted command line. Both clean. No corpus board is named that the repo does not
carry. Nothing refers to `.kicad_pro`-declared floors, check_drc annular /
poly-index / provenance, `PCBData.duplicate_references`, or anything wall-clock.
The added hunks introduce no double-blank-line runs: the file has the same one it
had before, and no triples.


edgehero:
Ordering note for the two open PRs: #691 removes a dead `--track-width-floor` string from `krt_capabilities.py`, and this PR's flag-liveness gate treats a flag as live if it survives as a string anywhere in non-test Python. I simulated that removal against this branch's gate and it still passes, because both docs here carry removal notes rather than prescriptions. So the two merge cleanly in either order. They touch no files in common.

drandyhaas:
Merging this. I pushed one commit to your branch first (`e9dc1250`) — the merge resolution plus a few review fixes. Here's what changed and why.

## Main moved under the branch, and had independently fixed two of the same things

`main` is now at `abbcd122` and `c320f38b` landed while this was open. It rewrote the same 9.3b width-clause passage and deleted the `track_width_floor` row from `docs/api-routing-config.md` — so both of your edits to those spots conflicted. **I resolved both in main's favour, not because yours were wrong but because main's is more informative:** it names the actual lever (`--track-width`, or the board's Default netclass) and the re-neck formula `min(nominal, fab_track, netclass)`, where the branch said only that the flag was gone. Main also fixed the same stale `krt_capabilities.py` usage example.

What survives from that half of the PR is the `--protect-nets` correction (still stale on main) and the `route_disconnected_planes.py` → `repair_planes.py` rename. I audited the merge result to confirm your other hunks all survived and nothing duplicated.

## The liveness gate was blind to two of the three flags it was written for

This is the one substantive fix. `live_flags()` counted any `--flag`-shaped string in non-test source, and the premise in the comment — that a removed flag "disappears from the source entirely" — doesn't hold: a removed flag survives in docstrings, usage examples and legacy tuples. Measured on your commit:

```
--track-width-floor -> LIVE      (krt_capabilities.py's own usage example)
--protect-nets      -> LIVE      (a tuple in tests/stress/manifest_to_plan.py)
--net-layers        -> dead
```

So the gate could not have caught `--track-width-floor`, the flag that cost run 22 the lap. It passed only because the branch had already removed the doc mentions by hand — re-adding one would have stayed green.

The scan now reads flag literals out of `add_argument(...)` calls only, quote-aware (help strings routinely carry unbalanced parens). **Both objections your comment raised still hold:** sub-package registrars stay live because the scan is a union over every non-test source file, not a per-CLI parser walk (`--suggest-locks`, `--allow-routed` verified live), and generated boolean pairs stay live via the `--no-` derivation. One correction there — the comment cited `--no-ratsnest` as the motivating pair, but there's no such flag; `render_placement.py` registers `--ratsnest-nets` and `--ratsnest-all`. I swapped in a real `BooleanOptionalAction` case (`--refs` → `--no-refs`, `py_router/route_render.py`).

Verified both directions, which the original only did in one: clean tree exits 0, and a reintroduced `--protect-nets` in the skill now exits 1 — an injection the old scan passed silently.

I also documented how this gate divides labour with `tests/test_431_skill_commands.py`, which is *already* a flag-liveness gate over the same skill files with a stronger method (the named tool's real `--help`). They're complementary and neither should be deleted as a duplicate: #431 reads **command lines** per-tool, this reads **prose** as a union. The stale `--track-width-floor` was a sentence, never a command line, which is why #431 couldn't have caught it and isn't at fault.

## Dropped: the `test_run15_handback_contract.py` additions

The 19 `ADDED_SECTIONS` anchors pinned literal prose sentences at exactly-one occurrence each, so any rewording of the doctrine fails CI until someone edits the test. That's too rigid for prose we'll keep tuning. The constants half was a good idea — re-deriving `EXACT_FILL_TIMEOUT`/`ORACLE_DRC_TIMEOUT` from the modules — and I'd take that back as a standalone test if you want to resubmit it.

## Split by audience: "your own checks are instruments too"

Kept in the routing skill: the framing, "test both directions", "suspect the check first", and the two table rows shaped like something a routing agent actually does (the probe calling a nonexistent function, the `awk` splitter matching the first of two identical headers).

Moved to `CLAUDE.md`'s **Testing & Verification**: the two rules pointing at `tests/run_utils.py` (`check(argv, refuse=...)`, `evidence(path)`). `run_utils` is a tests-only module — nothing outside `tests/` imports it, and a board-routing agent never writes a `run_all` test. That advice is for repo development, and it's discoverable where the rest of that genre already lives.

## Your two open questions

- **`EXTERNAL` allowlist** — left as-is, per your recommendation. Narrowing it doesn't get easier under the new scan, and the class it shadows (generic verbs shared with shell tools) still isn't the class that gets removed.
- **`run_doc_examples.py`** — confirmed. 4 failures, and I re-ran on a clean `main` checkout: identical there, same missing `kicad_files/test_diffpair_ram.kicad_pcb` fixture. Pre-existing, not from this diff.

## Verification

`test_doc_flag_liveness`, `test_krt_capabilities`, `test_431_skill_commands`, `test_431_board_gates`, `test_411_placement_siblings`, `test_457_startup_checks`, `test_run15_handback_contract`, `test_protect_nets_flag` — all exit 0.

One small extra: the `krt_capabilities.py` example named `board_score`, which isn't in its 22-module inventory, so the line printed as documentation exited 3. Kept main's flag choice and fixed the module token — it exits 0 as printed now.

Thanks for the audit discipline in the writeup — grounding each claim against main is what made this reviewable, and every mechanism I spot-checked (`boxed_in[].geometry`, the `rip_blocker_nets`-only PROTECTED log line, the two timeout constants, `_neck_pair_partner_grazes`) was accurate.

PR #715, opened 2026-08-22
capacity options and board_brief: say what a board cannot hold, with the number

edgehero:
Two modules that were written during the recovery runs and never proposed. They ride together because the second reads the first.

## `placement/options.py` + `check_capacity.py`

The prohibition against resizing a board's outline is written in six places, and every one of them says the same sentence: the honest response to a board that is genuinely too small is to say so with the measured number and stop. The number was never computed. `out_of_board_area` is the nearest quantity and `floorplan.py` explicitly disqualifies it as a budget key, because a part sitting inside a cutout scores 0.0 area.

So a run that hit a genuinely too-small board could only stop, and the person reading it had to go and measure by hand what to change.

Five options, each reporting the house shape (`ran` / `reason` / `measured` / `expected` / `action`), where `ran: false` means NOT MEASURED and never "fine":

- `grow_board` sums per-side courtyard area against the usable outline, and proposes a square side
- `add_layers` re-runs the escape ledger at another layer count
- `move_blocker` converts a face deficit into the millimetres a named part must move
- `relax_clearance` and `smaller_footprint` do the same for their levers

It reports; it never refuses and never acts. That follows `lock_advisor`, which prints a paste-ready `--lock` and locks nothing, because a wrong auto-action fails invisibly. Nothing here writes Edge.Cuts or a stackup.

## `board_brief.py`

One artifact describing what a board looks like to something that has to place it. Seven sections: board, state, parts, mechanical, structure, measurements, and an opt-in `fit`.

The design rule is that it **assembles and does not compute**: every number comes from the function that already owns it, and a `sources` map names that function per section, so there is one source of truth per fact and the brief cannot drift away from the graders.

Its `measurements` section reads `placement/options.py`. Without it the section records the ImportError in `skipped` rather than reporting a number, which is the correct degradation but not a useful one, and it is why these are one PR rather than two.

`tests/stress/stage_unaided.py` (352 lines) is included because `test_board_brief`'s staged arm imports it. It stages a board for an unaided run, netlist in and nothing else, which `--kind pile` does not do: pile hands over every free part's original rotation, measured at 65 of 65 on splitflap_driver.

## Tests

185 assertions, run against this branch on current main: `test_capacity_options.py` 52 passed, `test_board_brief.py` 133 passed.

## Two things I would rather flag than quietly fix

**`CROWDED_UTILISATION = 0.55` fires on shipping boards.** Measured across the 22 in-repo boards with outlines: median utilisation 0.495, but 8 of 22 are at or above 0.55, and the three highest are human-designed and routed: orangecrab_ext_pll 0.946, ulx3s 0.913, glasgow_revC 0.821. On those the advisory prints "Consider a larger outline or a denser layer stack", which is the one action the toolchain forbids everywhere else.

I have not changed the constant in this PR. Picking a new one needs a corpus sweep deduplicated by distinct design, several in-repo entries being the same board at different chain stages, and doing that unreviewed inside the PR that introduces the tool seemed worse than saying so here. The band that would not fire on any clean-routing in-repo board is nearer 0.91 than 0.55.

**`usable_area_mm2` is a bounding box.** `(W - 2*edge) * (H - 2*edge)` from `board_bounds`, so cutouts, slots and L-shapes are invisible to the denominator and utilisation is optimistic on exactly the boards with mechanical constraints. Measured on watchy: bbox 1553.06 mm2, cutouts 118.97, true copper 1434.09, so the reported 0.4687 should be about 0.509. `board_brief` already computes the true shoelace area and deliberately refuses to reuse the name, which is the seam if you want it fixed.

Happy to do either or both in this PR if you would rather they land together.

### DIFF
--- a/py_placer/placement/options.py
+++ b/py_placer/placement/options.py
@@ -1,0 +1,686 @@
+"""When a board cannot hold its parts, say so WITH THE NUMBER and the options.
+
+The prohibition this serves is written down in six places -- docs/floorplan-
+intent.md:29-40, floorplan.py:14-19, check_floorplan.py:19-24,
+rule_envelope:692, and both placement skills -- and every one of them says the
+same sentence:
+
+    the honest response to a board that is genuinely too small is to say so
+    with the measured number and stop
+
+The number was never computed. `out_of_board_area` is the nearest quantity and
+`floorplan.py:347-360` explicitly DISQUALIFIES it as a budget key, because a
+part sitting inside a cutout scores 0.0 area. Nothing sums a required area,
+nothing re-runs a lane ledger at another layer count, and nothing converts a
+face deficit into the millimetres of span that would clear it. So a run that
+hits a genuinely too-small board can only stop, and the person reading it has
+to go and measure by hand what to change.
+
+THIS REPORTS; IT NEVER REFUSES AND NEVER ACTS. That is the house pattern from
+`placement_driver._guard_congestion:779-823` -- "ok is False ONLY when the
+evidence is missing; the numbers themselves never refuse... The executor
+decides" -- and from `lock_advisor`, which prints a paste-ready `--lock` and
+locks nothing, because a wrong auto-action fails invisibly.
+
+Nothing here writes Edge.Cuts or a stackup. The only two things in the tree
+that write an outline stay `kicad_writer.strip_zero_length_edge_cuts` and the
+corpus tool `fix_outline_gaps.py`. This proposes with a number, exactly as
+`recommend-stackup/SKILL.md:44` already does: "propose the nearest workable
+option... Do not modify the board file directly -- stackup is a fab-facing
+decision the user must own."
+
+Every option carries the house shape:
+
+    ran        bool -- False means NOT MEASURED, never "fine"
+    reason     why it did not run
+    measured   what this board is
+    expected   what it would have to be
+    action     the sentence or argv a human would act on, never run here
+"""
+from __future__ import annotations
+
+import math
+from typing import Dict, List, Optional, Sequence
+
+# A board packed tighter than this by courtyard area is worth remarking on even
+# when every part currently fits: the parts have to be ROUTED between, and the
+# corpus's own comfortable boards sit far below it. Deliberately advisory --
+# see the module docstring on why nothing here refuses.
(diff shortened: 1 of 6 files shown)


drandyhaas:
Accepting all of this as-is. Verified against current main (`e5b973f5`): merges clean, `test_capacity_options` 52 passed and `test_board_brief` 133 passed on the merged content, and both front-ends run on a real board (`check_capacity.py` and `board_brief.py` on `watchy.kicad_pcb`, exit 0).

On your two flags — I re-measured both independently rather than taking them on trust, and both reproduce:

**`CROWDED_UTILISATION = 0.55`.** My sweep over the in-repo boards with outlines: median utilisation 0.492 (you had 0.495), and the top three are the same human-designed, human-routed boards you named — orangecrab_ext_pll 0.919, ulx3s 0.902, glasgow_revC 0.804. I count 6 of 22 at or above the line where you said 8, which is parameter sensitivity in how we each measured, not a disagreement about the finding. You were right not to pick a new constant inside the PR that introduces the tool; a corpus sweep deduplicated by distinct design is the right way to set it, and "nearer 0.91 than 0.55" matches what I see.

**`usable_area_mm2` is a bounding box.** Watchy's bbox comes out at 1553.06 mm² here, matching your number exactly, so cutouts are invisible to the denominator and utilisation reads optimistic on precisely the boards with mechanical constraints. Agreed that `board_brief` computing the true shoelace area and deliberately refusing to reuse the name is the right seam.

To answer "happy to do either or both in this PR": no — land it as it stands. Both are pre-existing-by-construction rather than regressions, and folding a corpus-derived constant change into the introducing PR is exactly the coupling you avoided. They are better as separate, separately-reviewable work.

One thing worth stating plainly for whoever picks those up: nothing outside these tests calls any of this yet. No skill names `check_capacity.py` or `board_brief.py` and no chain step invokes them, so at the moment this is capability rather than behaviour — which is also why shipping `0.55` today costs nothing in practice. It starts to matter the moment a skill or driver reads the advisory, so the constant should be settled before that wiring lands, not after.

Also note `tests/test_457_fresh_clone_fixtures.py` was red on this branch before merge — that was #716's untracked-fixture bug, not anything here. It is fixed on main now and passes on the merged content.

Issue #716, opened 2026-08-22
run_all: three reds and a 600s timeout from three unrelated causes (two share a missing untracked fixture)


drandyhaas:
Follow-ups measured while reviewing and merging #691 (merged as `abbcd122`). None of these were caused by that PR — they are what a main-vs-branch `run_all` baseline exposed. Filed together because they were found in one pass; they are independent and can be closed independently.

Baseline, this machine, 2026-08-22, both on the same hardware back to back:

| | passed | failed | timed out | wall |
| --- | --- | --- | --- | --- |
| `main` @ 4f296ea1 | 388 | 2 | 2 | 1448 s |
| #691 branch merged w/ main | 395 | 3 | 1 | 1069 s |

The bucket difference is **not** a regression: +7 passed is #691's seven new test files, and `test_obstacle_map_balance` flipped timeout → failure only because the branch run was less contended and got past 600 s to reach a check main never reached.

---

## 1. `kit-out-plane-connected.kicad_pcb` is absent and untracked — this is TWO of the three reds

```
$ ls kicad_files/kit-out-plane-connected.kicad_pcb
No such file or directory
$ git ls-files kicad_files/kit-out-plane-connected.kicad_pcb   # 0 lines
$ grep kit-out-plane-connected tests/fixture_boards.py         # no recipe
```

`tests/test_654_reconcile_status.py:29` points `BOARD` straight at it. Both failures follow:

- **`test_654_reconcile_status.py` — 8/11 checks pass.** The three that need the board degrade to an empty result rather than skipping or saying why:
  ```
  FAIL a run prints exactly one status line -- 0 found: []
  FAIL the fixture really has zones (else this tests nothing) -- []
  FAIL KICAD_NO_GRADE_RECONCILE is disclosed, not silent -- []
  ```
  Note the second one is the test's own vacuity guard, which is doing its job — it is telling us the other checks are testing nothing.
- **`test_457_fresh_clone_fixtures.py`** correctly flags it:
  ```
  AssertionError: test(s) reading an UNTRACKED board straight out of kicad_files/:
    test_654_reconcile_status.py: kit-out-plane-connected.kicad_pcb
  On a fresh clone that board does not exist. Add a recipe to tests/fixture_boards.py
  and call ensure(), or use a tracked board.
  ```

[...]

Issue #696, opened 2026-08-22
Placement: narrow_pad_joint is missing from check_weird's CATEGORIES, so it blocks DONE while printing nothing


edgehero:
`check_weird` emits a `narrow_pad_joint` finding at `py_router/check_weird.py:609`, but that category is not in the list the reporter iterates.

`py_router/check_weird.py:66-68`:

```python
CATEGORIES = ['dangling-end', 'soft-joint', 'redundant-cycle',
              'removable-segment', 'stacked-copper', 'unsupported-via',
              'dangling-via', 'orphan-island']
```

Eight entries, and `narrow_pad_joint` is not among them. `print_report` iterates `CATEGORIES` (`:698`).

So a board whose only finding is a `narrow_pad_joint` prints `FOUND 1 WEIRD THINGS` with every listed category showing 0, exits 1, and blocks DONE through `check_complete` (`check_complete.py:414-415`), with nothing on screen naming what to fix.

It is also unfilterable: the finding is emitted with `size=None`, so the `--tolerance` filter at `:685-687` never applies to it.

Suggested fix: add `'narrow_pad_joint'` to `CATEGORIES`. If the tolerance filter should apply, give the finding a size at emission.

Issue #597, opened 2026-08-10
Nets with leading-dash names (-3V3) break route.py/repair_planes argparse — eit_mux unreplayable


drandyhaas:
Boards with a net named `-3V3` (eit_mux: `-3V3 +1V1 VBUS` power set) cannot pass it through the CLI: argparse classifies any non-numeric leading-dash token as an option, so `--power-nets -3V3 +1V1 VBUS` dies with `error: unrecognized arguments: -3V3` in route.py AND repair_planes.py. The recorded manifest replays the same argv, so the board's chain breaks at every board-producing step — observed as eit_mux producing a broken chain in ALL 25 arms of the #584 Modal sweep (equal across arms, so comparisons pair-drop it, but the board is effectively lost to the corpus).

Sibling of the #312 net-name-escaping family. Candidate fixes:
- accept `--power-nets=-3V3` style by pre-splitting a single `=`-joined, comma-separated value (`--power-nets=-3V3,+1V1,VBUS`) — recorder emits that form when any member starts with `-`;
- or a tools-wide argv pre-pass that rewrites tokens matching a known net of the input board before argparse sees them (fragile);
- or an explicit escape (`--power-nets @-3V3` stripped after parse) mirroring how net glob patterns already handle `!`.

Recorder + all four fronts + manifest_to_plan FLAG_PARAMS need the same treatment (CLI/GUI parity rules apply).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0162EFgykSqFfJgcQxYmR3KB

drandyhaas:
Fixed in 55632d9: cli_nets.pin_dash_digit_values pins argparse's negative-number matcher to ^-\d in all ten net-accepting CLIs, so dash-digit net names (-3V3, -12V, -5V) always parse as values regardless of Python version. Root cause was version-dependent argparse behavior: 3.14's lenient prefix matcher let -3V3 through (why local replays worked), 3.12/3.13's strict pattern rejected it (why every container task failed). Proven under 3.13 both ways; eit_mux's banked broken sweep rows get a top-up at this commit after the current sweep completes.

PR #719, opened 2026-08-22
tests: clear the four run_all reds (#718)

### LINKED ISSUE #718: run_all: four reds -- two #522 reorg stragglers, a #597 merge conflict, and a test pinning tessellation reverted in 6166a98b
Four independent causes, none of them a routing regression. Filed together because they came out of one `run_all` pass; they are independent and can be closed independently.

Baseline, this machine (Windows, Python 3.13.14), `main` @ `d1aa5338`, after bringing the local Rust router in line with `Cargo.toml` (0.20.0 on disk vs 0.21.1 declared was masking everything behind `startup_checks`):

| | passed | failed | timed out | wall |
| --- | --- | --- | --- | --- |
| `run_all.py --fast` (282 unit) | 282 | 0 | 0 | 559 s |
| `run_all.py` (all 403) | 397 | 4 | 2 | 2916 s |

All four reproduce on a pristine `main` checkout, and all four are **regressions**: each shows `PASS` in recorded runs (`wk/run15/_tests_final.txt`, `wk/pr_prep/run_all_output.txt`). None appear in #716's baseline, which is the interesting part — see item 5.

---

## 1. `test_run6_backlog` and `test_run5_exchange` — the `#522` reorg the `ed779096` sweep missed

Both spawn a CLI at the repo root that no longer lives there:

```
$ python3 -X utf8 ./qfn_fanout.py ...
python3.exe: can't open file '...\KiCadRoutingTools\qfn_fanout.py': [Errno 2] No such file or directory
EXIT=2
```

| test | spawns | actually at |
| --- | --- | --- |
| `test_run6_backlog.py:28,41` | `ROOT/qfn_fanout.py`, `ROOT/bga_fanout.py` | `py_router/` (`ee860796`) |
| `test_run5_exchange.py:134,138` | `ROOT/check_floorplan.py`, `ROOT/place_reconstruct.py` | `py_tools/`, `py_placer/` (`a241b5ac`) |

The symptom is misleading in exactly the way `run_utils.tool`'s docstring predicts: the interpreter writes to **stderr** and stdout is `''`, so `assertIn('Auto-detected QFN/QFP component: U3', r.stdout)` fails as a bare assertion that reads like a product bug, and `test_run5_exchange`'s `assertEqual(r0.returncode, 0)` reads CPython's missing-file exit 2 as a tool exit code.

Neither test was in `ed779096`'s file list ("25 of 26 red tests were not FAILING, they were not RUNNING"), and `git merge-base --is-ancestor ee860796 61540b75` is false — the tests were correct on their branches and went stale on merge.

**Consequence worth noting:** `test_run5_exchange`'s `TestLadderEndToEnd` body has therefore **never executed** since `a241b5ac`. (Good news — it passes once the paths resolve.)

`tests/run_utils.py` already ships `tool()` / `tool_env()` for exactly this, and four sibling tests already use them. These files predate that adoption.

## 2. `test_run14_dash_net_hint` — a semantic merge conflict with #597

`test_route_py_emits_the_hint_and_still_exits_2` asserts `route.py --nets +12V -12V` exits 2 with `unrecognized arguments`. Since #597 (`55632d91`) it does not: `cli_nets.pin_dash_digit_values` replaces argparse's negative-number matcher with `^-\d` at `route.py:5721`, so `-12V` is read as a **value** and the run succeeds.

```
plain matcher  ^-\d+$|^-\d*\.\d+$   ->  SystemExit 2 ("unrecognized arguments: -12V")
pinned matcher ^-\d                 ->  Namespace(nets=['+12V', '-12V'])
```

`git merge-base --is-ancestor 55632d91 4b4098a0` → **false**: the #597 fix and this test were developed in parallel and each passed alone.

The test has a guard for precisely this — `_argparse_takes_dash_values()` (`:38-51`) — but it probes a **bare** `ArgumentParser`, one `route.py` no longer builds, so it returns `False` and the arm never skips. The guard measures the wrong parser.

Two side effects:
- The arm passes `ROOT/_no_such_out.kicad_pcb` on the assumption route.py dies in argparse first. It now really routes and writes, leaving untracked `_no_such_out.kicad_pcb` (~350 KB) and `_no_such_out.kicad_pro` in the repo root after every suite run.
- The hint machinery itself is fine — its four unit tests pass. It is simply inert *through the shipped CLIs*, which is #597's stated goal.

## 3. `test_run4_custom_pad_circle` — pins the tessellation `6166a98b` deliberately reverted

`TestRun3ArchivedBoard.test_the_shipped_graze_is_now_visible` expects `check_drc -c 0.09 --clearance-margin 0` to exit 1 on `wk/run3/final2.kicad_pcb`. It exits 0.

Not a checker bug — `6166a98b` ("revert six placement-only routing changes that lose to main on the corpus") put `kicad_parser.py:873` back to a fixed inscribed 32-gon, on the stated grounds that the adaptive version "lives in the PARSER, so it moved obstacle geometry and DRC grading together on every board with circle-primitive custom pads". The graze is still physically there:

| quantity | value |
| --- | --- |
| true copper gap | 0.0884 mm (0.0016 short of the 0.09 floor) |
| 32-gon sagitta at r=0.5 | **0.002408 mm** — larger than the shortfall |
| modelled gap, 32-gon | 0.0908 mm → clean |
| `check_drc -c 0.095` | exit 1, `Pad:/SWDIO (JP2.2) <-> Seg:/ICE_CDONE` |

So the test pins reverted behaviour and nothing reconciled it. Also worth recording, because it is why this sat unnoticed: **the other three sub-tests in the file do not cover tessellation at all.**

- `test_sub_floor_gap_is_flagged` / `test_above_floor_gap_stays_clean` are insensitive by construction — `board_with_gap` puts the rect pad due east of the circle centre and polygon vertex `k=0` sits at angle 0, i.e. exactly the true rightmost point (measured `max_x` = 50.500000 at N=32). The measured gap equals the true gap for any N.
- `test_polygon_radial_error_budget` calls `_adaptive_circle_n` directly. That helper now has **1 definition and 0 call sites** in `kicad_parser.py` — dead code whose only caller in the repo is the test.

## 4. The two timeouts are contention, not defects

Both pass when run alone on this machine, so they need no fix — recording them so the next reader does not re-diagnose:

| test | in `run_all -j 4` | alone |
| --- | --- | --- |
| `test_obstacle_map_balance` | timed out at its own 1200 s budget | **PASS**, including `repair: repairs actually performed (recipe still has work)` |
| `test_459_group_routing` | timed out at the 600 s cap (declares no budget) | **PASS** |

Note the first one clears the check #716 item 2 flagged as possibly real. The second may deserve a `RUN_ALL_TIMEOUT` declaration.

## 5. Why none of this showed up in #716's baseline — untracked `wk/` boards

#716 measured `main` @ `4f296ea1` at 2 failed / 2 timed out, and none of these four were among them. The reason is that four of these tests read boards straight out of gitignored `wk/`:

| board | read by |
| --- | --- |
| `wk/run5/s1_pour.kicad_pcb` | `test_run6_backlog` |
| `wk/b2/tigard__swap/d0/perturbed.kicad_pcb` | `test_run5_exchange`, `test_part_class` |
| `wk/run3/final2.kicad_pcb` | `test_run4_custom_pad_circle` |

Absent those, each `skipTest`s and the **file still exits 0** — green while covering nothing. This is the same class #716 item 1 fixed for `kit-out-plane-connected.kicad_pcb`, and `test_457_fresh_clone_fixtures` is the guard for it — but its regexes (`:120-121`) only scan `kicad_files/`, so `wk/` is invisible to it.

That makes item 3 above a *latent* finding on any machine but this one, and it means the suite's green/red state is machine-dependent in a way nothing reports.

---

## Suggested fixes

1. Port the two stale-path helpers onto `run_utils.tool()` / `tool_env()` (raises loudly instead of dying into an empty stdout), plus a guard that fails when any test spawns a root-level script that is not there.
2. Re-pin the #597 arm on what `route.py` now does, rather than skipping it, and route its output into a temp dir.
3. Re-pin the run-3 arm on the shipped geometry — masked at 0.09, visible at 0.095 — with the 32-gon vertex count and inscribed-ness asserted directly, and `_adaptive_circle_n`'s zero call sites asserted so restoring it is a decision rather than a drift.
4. Widen `test_457_fresh_clone_fixtures` past `kicad_files/` so a test depending on an untracked `wk/` board is named rather than silently skipped.

Happy to send a PR for all four; say if you would rather take any of them differently — item 3 in particular is a judgement call about what the file should pin now.

edgehero:
Closes #718

Four independent causes, none of them a routing regression. Each is separable — take any one and leave the rest if you disagree with it. Item 3 is the judgement call.

## 1. Two `#522` reorg stragglers the `ed779096` sweep missed

`test_run6_backlog` spawned `ROOT/qfn_fanout.py` and `ROOT/bga_fanout.py`; `test_run5_exchange` spawned `ROOT/check_floorplan.py` and `ROOT/place_reconstruct.py`. Python writes `can't open file ...: [Errno 2]` to **stderr** and exits 2, so stdout is `''` and the assertions fail with bare messages that read like product bugs — precisely the class `run_utils.tool()`'s docstring records for eighteen tests.

Both now use `run_utils.tool()` / `tool_env()`, matching `test_part_class.py:25,114-118`. `tool()` raises naming what it looked for; `tool_env()` fixes the sibling defect both files also had (a child launched with `PYTHONPATH=ROOT` alone can no longer `import kicad_parser`).

**`tests/stress/corpus_noop_sweep.py` carried three more of the same shape** (lines 85/136/146). `run_all` does not discover that file, so those were broken silently — found by the new guard, not by hand.

**`test_run5_exchange`'s `TestLadderEndToEnd` had therefore never executed since `a241b5ac`.** It passes.

## 2. `test_run14_dash_net_hint` — a semantic merge conflict with #597

The arm asserted `route.py --nets +12V -12V` exits 2 with `unrecognized arguments`. Since #597 (`55632d91`) `cli_nets.pin_dash_digit_values` swaps argparse's negative-number matcher for `^-\d`, so `-12V` is read as a **value** and the run succeeds.

```
plain matcher  ^-\d+$|^-\d*\.\d+$   ->  SystemExit 2 ("unrecognized arguments: -12V")
pinned matcher ^-\d                 ->  Namespace(nets=['+12V', '-12V'])
```

`git merge-base --is-ancestor 55632d91 4b4098a0` is **false** — the two were built on parallel branches and each passed alone. The test's own guard, `_argparse_takes_dash_values()`, probes a **bare** `ArgumentParser`, one `route.py` no longer builds, so it never skipped.

**Inverted rather than skipped**, so it still fails if the pin is ever dropped. The four unit tests of the hint machinery are untouched and still pass — the hint is inert *through the shipped CLIs*, which is #597's stated goal, and the module docstring now says so.

Its output also goes to a tempdir now. It had been passing `ROOT/_no_such_out.kicad_pcb` on the assumption route.py dies in argparse first; since #597 it really routes and writes, leaving `_no_such_out.kicad_pcb` (~350 KB) and `.kicad_pro` in the repo root after every suite run.

## 3. `test_run4_custom_pad_circle` — the judgement call

It pins the adaptive tessellation `6166a98b` deliberately reverted to a fixed 32-gon. **I did not restore it** — that was a measured corpus decision (real DRC 71 → 49 across 73 boards) and reversing it silently is not mine to make. The graze is still physically in the copper:

| quantity | value |
| --- | --- |
| true gap | 0.0884 mm — 0.0016 short of the 0.09 floor |
| 32-gon sagitta at r=0.5 | **0.002408 mm** — *exceeds* the shortfall |
| modelled gap | 0.0908 mm → grades clean |
| at `-c 0.095` | exit 1, `Pad:/SWDIO (JP2.2) <-> Seg:/ICE_CDONE` |

So the arm is re-pinned on what ships: **masked at 0.09, the same pair visible at 0.095**, with the sagitta-vs-shortfall arithmetic asserted so the claim cannot rot silently. If you would rather restore the adaptive path, this is the one to send back.

While there — the file's other three checks did not cover tessellation at all, which is why this sat unnoticed:

- `test_sub_floor_gap_is_flagged` / `test_above_floor_gap_stays_clean` are insensitive by construction: `board_with_gap` puts the rect due east of the circle centre and polygon vertex `k=0` sits at angle 0, so `max_x` is 50.500000 at N=32 and the measured gap equals the true gap for **any** N.
- `test_polygon_radial_error_budget` calls `_adaptive_circle_n`, which now has **1 definition and 0 call sites**.

Added direct checks of the shipped geometry — vertex count, inscribed-ness, and zero call sites — so restoring the adaptive path is a decision rather than a drift. The insensitivity is documented in place rather than papered over.

## Two guards, because cause 1 is now its third occurrence

**`tests/test_718_stale_tool_paths.py`** — every `[sys.executable, ...]` argv in `tests/` must name a script that exists, and the failure says where it moved. It scans **list literals**, not spawn call sites, because three of these were built as `argv = [...]` and passed by name.

It is explicit about what it cannot see: the two headline tests used `os.path.join(ROOT, script)` with `script` a **parameter**, which no static check resolves — that is exactly why they were moved onto `tool()` rather than repointed. Remaining variable call sites are reported, not failed (`test_431_skill_commands` feeds that shape from `discovered_tools()`, which only admits names that already resolve, so failing it would be wrong).

**`test_457_fresh_clone_fixtures`** gains the `wk/` half of its scan. Its regexes only matched `kicad_files/` (`:120-121`), and **15 test files read boards out of gitignored `wk/`**. On any machine without that work output they `skipTest` and the file still exits 0 — green while covering nothing. That is why cause 3 was invisible to #716's baseline and visible here.

Recorded as a **ratchet**, not an approval list: a new file fails the gate, and clearing one without deleting its entry *also* fails, so the census cannot go stale. Fixing the 15 is a separate job and I did not fold it in — say if you would rather have recipes than a baseline.

It also now registers the new test in `TESTS` and asserts every `test_*` in that file is registered — the new one silently did not run until I noticed, which is the same hand-maintained-registry bug as #696.

## The two timeouts are not defects

Both pass run alone on this machine, so nothing is changed for them:

| test | in `run_all -j 4` | alone |
| --- | --- | --- |
| `test_obstacle_map_balance` | timed out at its own 1200 s budget | **PASS**, including `repair: repairs actually performed` — the check #716 item 2 flagged as possibly real |
| `test_459_group_routing` | timed out at the 600 s cap | **PASS** (declares no `RUN_ALL_TIMEOUT`; may deserve one) |

## Verification

| run | before | after |
| --- | --- | --- |
| `run_all.py --fast` | 282 passed, 0 failed | **283 passed, 0 failed** (+1: the new static guard) |
| `run_all.py` (all 404) | 397 passed, **4 failed**, 2 timed out | **403 passed, 0 failed**, 1 timed out, 2771 s |

The remaining timeout is `test_obstacle_map_balance` at its own 1200 s budget under `-j 4`; it passes run alone (above). No stray `_no_such_out.*` in the working tree after the full run.

Both new guards were **mutation-tested** — they must fail on the pre-fix tree, not merely pass after it:

| mutation | result |
| --- | --- |
| put one `corpus_noop_sweep` path back | `test_718` FAILS, naming `corpus_noop_sweep.py:136 -> check_floorplan.py (it moved to py_tools/check_floorplan.py)` |
| a new test file reads a `wk/` board | `test_457` FAILS: `new test(s) reading a board out of gitignored wk/` |
| a baselined file stops reading `wk/` | `test_457` FAILS: `delete them from _WK_BOARD_CONSUMERS so the census stays true` |
| a `test_*` defined but not in `TESTS` | `test_457` FAILS: `absent from TESTS, so they never run` |

One note on the new guard file: it deliberately avoids the literal word for a spawned child process in its prose, because `run_all` classifies a test as slow "integration" by substring-matching its **source** for it. This check spawns nothing and runs in well under a second, so it belongs in the `--fast` loop where a stale path is caught immediately. That is stated in the file.

### DIFF
--- a/tests/stress/corpus_noop_sweep.py
+++ b/tests/stress/corpus_noop_sweep.py
@@ -82,7 +82,8 @@ def sweep_reconstruct(board, workdir):
     """place_reconstruct --dry-run: proposals and would-move must both be empty."""
     out = os.path.join(workdir, 'out.kicad_pcb')
     argv = [sys.executable, '-X', 'utf8',
-            os.path.join(ROOT, 'place_reconstruct.py'), board, out, '--dry-run']
+            os.path.join(ROOT, 'py_placer', 'place_reconstruct.py'),
+            board, out, '--dry-run']
     proc = subprocess.run(argv, capture_output=True, text=True, encoding='utf-8',
                           errors='replace', cwd=ROOT, timeout=TIMEOUT_S)
     rep = _summary(proc.stdout or '')
@@ -133,7 +134,8 @@ def sweep_repair(board, workdir):
     """
     intent = os.path.join(workdir, 'intent.json')
     emit = subprocess.run(
-        [sys.executable, '-X', 'utf8', os.path.join(ROOT, 'check_floorplan.py'),
+        [sys.executable, '-X', 'utf8',
+         os.path.join(ROOT, 'py_tools', 'check_floorplan.py'),
          board, '--emit-intent', intent],
         capture_output=True, text=True, encoding='utf-8', errors='replace',
         cwd=ROOT, timeout=TIMEOUT_S)
@@ -143,7 +145,8 @@ def sweep_repair(board, workdir):
             None, f'emit-intent rc={emit.returncode} '
                   + (emit.stderr or '')[-140:])
     out = os.path.join(workdir, 'rep.kicad_pcb')
-    argv = [sys.executable, '-X', 'utf8', os.path.join(ROOT, 'place_seed.py'),
+    argv = [sys.executable, '-X', 'utf8',
+            os.path.join(ROOT, 'py_placer', 'place_seed.py'),
             board, out, '--intent', intent, '--repair', '--dry-run']
     proc = subprocess.run(argv, capture_output=True, text=True, encoding='utf-8',
                           errors='replace', cwd=ROOT, timeout=TIMEOUT_S)
--- a/tests/test_457_fresh_clone_fixtures.py
+++ b/tests/test_457_fresh_clone_fixtures.py
@@ -19,6 +19,8 @@ Two invariants, both of which were violated:
    suite left modified tracked files in the working tree.
 """
 
+import ast
+import glob
 import os
 import re
 import subprocess
@@ -140,6 +142,100 @@ def test_no_test_reads_an_untracked_board_directly():
     print("  PASS: no test reads an untracked kicad_files/ board directly")
 
 
+#: Test FILES that read a board out of gitignored `wk/` (issue #718).
+#:
+#: This is a RATCHET, not an approval list. `wk/` is work output -- every board
+#: under it is untracked, so on any machine but the one that produced it these
+#: tests `skipTest` and their file still exits 0: green while covering nothing.
+#: That is how #718 item 3 hid. It was found only because this machine happened
+#: to still have `wk/run3/final2.kicad_pcb`; #716's baseline, on a machine that
+#: did not, reported the file as passing.
+#:
+#: The list may only SHRINK. Adding a new file fails the gate; clearing one and
+#: forgetting to delete its entry fails it too, so the census cannot go stale.
+#: Clearing an entry means a `tests/fixture_boards.py` recipe rooted in a
+#: tracked board, or repointing at a tracked board outright.
+_WK_BOARD_CONSUMERS = {
+    'test_outline_prefilter.py',
+    'test_part_class.py',
+    'test_place_reconstruct.py',
+    'test_placement_pad_legality.py',
+    'test_run4_custom_pad_circle.py',
+    'test_run4_reconstruct.py',
+    'test_run5_emit_guard.py',
+    'test_run5_exchange.py',
+    'test_run6_backlog.py',
+    'test_run6_body_overlap.py',
+    'test_run6_check_assembly.py',
+    'test_run7_vectors.py',
+    'test_run8_airwire_refuted.py',
+    'test_run8_oob_outline.py',
+    'test_run8_rigid_consistency.py',
+}
+
+
+def _reads_a_wk_board(tree):
+    """Line numbers of `os.path.join(..., 'wk', ..., '*.kicad_pcb')` calls.
+
+    AST rather than regex: these paths are built as multi-segment joins
+    (`join(ROOT, 'wk', 'run5', 's1_pour.kicad_pcb')`, and one via
+    `join(dirname(__file__), '..', 'wk', ...)`), which no single literal
+    pattern catches. Anchored on the 'wk' segment plus a .kicad_pcb segment, so
+    a test writing its own board into a tempdir is not swept in.
+    """
+    hits = []
+    for node in ast.walk(tree):
+        if not (isinstance(node, ast.Call)
+                and isinstance(node.func, ast.Attribute)
+                and node.func.attr == 'join'):
+            continue
+        lits = [a.value for a in node.args
+                if isinstance(a, ast.Constant) and isinstance(a.value, str)]
+        if 'wk' in lits and any(l.endswith('.kicad_pcb') for l in lits):
+            hits.append(node.lineno)
+    return hits
+
+
+def test_no_new_test_depends_on_an_untracked_wk_board():
+    """The cousin of the kicad_files/ scan above, for `wk/`.
+
+    That scan (test_no_test_reads_an_untracked_board_directly) matches only
+    'kicad_files/x.kicad_pcb', so a board read out of `wk/` was invisible to
+    it -- and `wk/` is gitignored in its entirety, so EVERY such read has the
+    fresh-clone problem the file exists to prevent.
+    """
+    here = os.path.dirname(os.path.abspath(__file__))
+    found = {}
+    for path in sorted(glob.glob(os.path.join(here, 'test_*.py'))):
+        try:
+            with open(path, encoding='utf-8', errors='replace') as fh:
+                tree = ast.parse(fh.read())
+        except SyntaxError:
+            continue
+        hits = _reads_a_wk_board(tree)
+        if hits:
+            found[os.path.basename(path)] = hits
+
+    added = sorted(set(found) - _WK_BOARD_CONSUMERS)
+    assert not added, (
+        "new test(s) reading a board out of gitignored wk/:\n  "
+        + "\n  ".join('%s (line %s)' % (f, ', '.join(map(str, found[f])))
+                      for f in added)
+        + "\nOn any machine without that work output the test skips and the "
+          "file still exits 0 -- green while covering nothing. Add a recipe to "
+          "tests/fixture_boards.py rooted in a TRACKED board and call ensure(), "
+          "or use a tracked board.")
+
+    cleared = sorted(_WK_BOARD_CONSUMERS - set(found))
+    assert not cleared, (
+        "these no longer read a wk/ board -- delete them from "
+        "_WK_BOARD_CONSUMERS so the census stays true:\n  "
+        + "\n  ".join(cleared))
+
+    print("  PASS: %d known wk/-dependent test file(s), none new"
+          % len(found))
+
+
 def test_fixtures_build_from_a_clean_state():
     """End to end: every fixture is producible right now."""
     for name in _RECIPES:
@@ -167,9 +263,30 @@ TESTS = [
     test_no_tracked_project_file_without_its_board,
     test_consumers_do_not_silently_skip_a_missing_fixture,
     test_no_test_reads_an_untracked_board_directly,
+    test_no_new_test_depends_on_an_untracked_wk_board,
     test_fixtures_build_from_a_clean_state,
     test_building_a_fixture_leaves_no_tracked_file_modified,
 ]
+
+
+def test_every_test_in_this_file_is_registered():
+    """TESTS is hand-maintained, so a new test here runs only if someone
+    remembers to list it -- and a test that never runs is indistinguishable
+    from one that passes. (Added after writing
+    test_no_new_test_depends_on_an_untracked_wk_board and watching it not run.)
+    """
+    listed = ({t.__name__ for t in TESTS}
+              | {'test_every_test_in_this_file_is_registered'})
+    defined = {k for k, v in sorted(globals().items())
+               if k.startswith('test_') and callable(v)}
+    missing = sorted(defined - listed)
+    assert not missing, (
+        "test(s) defined in this file but absent from TESTS, so they "
+        "never run:\n  " + "\n  ".join(missing))
+    print(f"  PASS: all {len(defined)} test functions are registered")
+
+
+TESTS.append(test_every_test_in_this_file_is_registered)
 
 
 if __name__ == '__main__':
--- a/tests/test_457_fresh_clone_fixtures.py
+++ b/tests/test_457_fresh_clone_fixtures.py
@@ -205,8 +205,11 @@ def test_no_new_test_depends_on_an_untracked_wk_board():
     fresh-clone problem the file exists to prevent.
     """
     here = os.path.dirname(os.path.abspath(__file__))
+    # Recursive: tests/stress/ and tests/gui_parity/ can grow one too. All of
+    # today's are in tests/ itself, so the baseline below is keyed on basename.
     found = {}
-    for path in sorted(glob.glob(os.path.join(here, 'test_*.py'))):
+    for path in sorted(glob.glob(os.path.join(here, '**', 'test_*.py'),
+                                 recursive=True)):
         try:
             with open(path, encoding='utf-8', errors='replace') as fh:
                 tree = ast.parse(fh.read())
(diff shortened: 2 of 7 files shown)


edgehero:
both fixed the same bug at the same time :D

drandyhaas:
I'd already pushed a fix for #718 (fccac7e2) before this landed — sorry for pre-empting the PR you offered. I read yours line by line afterwards rather than closing it unread, and **four things in here are better than what I shipped**, so they are now ported in with credit to you. Two others I checked and did not take, with reasons, because I think they are actually vacuous as written.

## Ported from this PR

**1. `tests/stress/corpus_noop_sweep.py` — a real bug I missed entirely.** Three stale root spawns, still broken on `main` until now. I had scanned only `tests/test_*.py` and never looked in `tests/stress/`. Your observation that `run_all` does not even discover it is the important half: `discover()` globs `tests/test_*.py` non-recursively, so the sweep had been dying into "can't open file" on every board with nothing to report it. Fixed via `run_utils.tool()` rather than repointing the strings, per your own note that a literal that resolves today can go stale tomorrow.

**2. The vacuity self-check.** "A scan that matches nothing passes for free" is the check I should have written and didn't — and it is not hypothetical: my first draft of the spawn gate matched only the LITERAL `os.path.join(ROOT, 'x.py')` form and **printed PASS against the pre-fix source**, because both headline offenders build the path from a variable. Exactly the point your docstring makes about what a static check cannot see. I only caught it by reverting the two files and watching the gate stay green. Your `seen < 20` floor is now in, generalised to three floors (files, argv lists, literal joins).

**3. Scope: recursive, and not limited to `test_*.py`.** This is what makes (1) catchable at all — a gate scoped the same way as the runner is blind to precisely the files the runner never runs, which are the ones most likely to rot. Mine had the runner's blind spot baked in.

**4. The `--fast` classification insight.** I had not registered that `is_integration()` substring-matches source text, so my two static gates sat inside `test_457_fresh_clone_fixtures` (which shells out to `git ls-files`) and were skipped by `--fast` — a millisecond AST scan hiding behind a 900 s classification. They now live in `tests/test_718_static_test_hygiene.py`, which avoids the marker word and runs in the fast loop. Amusingly the first version of that file still classified as integration, because I wrote the marker word in the docstring explaining why the marker word must not appear.

Also taken: `os.devnull` → tempdir in `test_run6_backlog` (a board write wants a real path), and your argument for keeping `_adaptive_circle_n`'s arithmetic under test — I had deleted that coverage with the stale arm, and "correct, and deliberately unused" is the better pair of assertions. `test_every_test_in_this_file_is_registered` is in too.

## Two I did not take, because I think they are vacuous

**`test_route_py_takes_the_bare_dash_net_since_597` does not do what its docstring claims.** It says it "still fails if the pin at route.py's parse_args is ever dropped." On Python 3.14 it does not. I removed the pin and ran your three assertions:

```
assertNotIn 'unrecognized arguments' : PASS
assertNotIn 'begins with'            : PASS
assertIn '-12V'                      : PASS
```

Bare argparse on 3.14 already accepts dash-prefixed values, so with the pin gone the run still parses and your arm cannot tell. It is a real regression detector on your 3.13 but not on mine — the same interpreter-dependence that produced the original conflict. What I shipped instead asserts the mechanism directly: `route.py`'s source routes through `pin_dash_digit_values`, **and** the pinned matcher changes a bare parser's behaviour. That fails on every interpreter when the pin goes.

**`_reads_a_wk_board` misses directory-shaped dependencies.** Requiring a `.kicad_pcb` literal drops two: `test_run20_run_watch.py` (`wk/run20`) and `test_run8_starved_face_gate.py` (`wk/run7/glasgow_revC`), both of which join a work DIRECTORY. Measured against the same tree: your filter finds 15 files, mine 17, and the difference is one-way. My census also records the board list per file (not just the filename) and prints which entries are inert on the current machine, which is the fact that actually explains a machine-dependent green.

One place I kept mine that is a genuine judgement call rather than a defect in yours: for the VARIABLE-form spawn you print a note and pass, on the grounds that `discovered_tools()` only admits names that already resolve. Mine fails on the shape and requires a declared exemption (that one call site, with your reasoning as its stated reason). Yours cannot false-positive; mine catches a *new* stale variable spawn, which yours would only note. I preferred the ratchet, but I don't think you're wrong.

## One thing worth your attention

For item 3 I went further than the archived-board re-pin, because `wk/run3/final2.kicad_pcb` is exactly the machine-dependence your item 5 is about — re-pinning it leaves the arm skipping everywhere but your machine. The defect reproduces synthetically: two `gr_circle`-primitive pads at the same 0.0884 true gap, one pair along a polygon-**vertex** ray and one along the ray exactly **between** vertices (θ = 180/N = 5.625°). Each 32-gon under-reaches by a full sagitta on the second, so `0.0884 + 2 × 0.002408 = 0.0932 > 0.09` and `check_drc` flags one and misses the other. No artifact, runs everywhere, and it fails if the tessellation changes in either direction.

Related, and it explains why the two pre-existing gap arms pinned nothing: vertex `k=0` is built at global angle 0 and the rect pad sits due east, so measured `max_x` is exactly 50.500000. Footprint rotation does **not** shift the phase — the polygon is constructed in global frame — so those arms cannot be made sensitive by rotating the fixture.

Thanks for this. The diagnosis in #718 was exact — every causal claim in it held when I checked it independently — and this PR caught a bug and three gate weaknesses I had missed.

Issue #697, opened 2026-08-22
Placement: the legality and repair census ignores pad.local_clearance, reporting 0 conflicts where check_drc grades 4


edgehero:
`pad.local_clearance` is documented in CLAUDE.md as the resolved per-pad clearance override, with the instruction that clearance consumers read that field and never re-derive footprint inheritance. The placement side never reads it.

```
$ grep -rn "local_clearance" py_placer/ py_tools/ | wc -l
0
```

`check_drc` does, at `py_router/check_drc.py:1926` (`_pad_pair_cl`) and `:2497-2500`, taking the max over both pads of the pair.

The placement census prices every pad pair at one flat scalar. `py_placer/placement/legality.py:1758-1766`:

```python
g = rect_gap(...)
if g >= clearance - EPS:      # broad-phase reject at one flat scalar
    continue
...
hit, over, _pt = check_exact(pa, pb, clearance,   # same flat scalar
                             routing_layers, clearance_margin=0.0)
```

A pair separated by more than `clearance` but less than `max(lc_a, lc_b)` never reaches the exact test, so `pad_conflicts` (`:1776`) never counts it.

Observed on a real board: the repair census reported `Repair census: 0 conflict pair(s), all listed` and `"pad_conflicts_before": 0`, while `check_drc` on the same board reported `"by_type": {"pad-pad": 4}` with `"required_mm": 1.016` and the note `Required clearance: 1.0160mm (local/netclass override; global 0.1500mm)`. It was a fiducial keep-clear. `place_reconstruct --stages legalize` reported the same 0.

The consequence is that the fix tool reports nothing to fix on a board that cannot pass DRC, and every instrument reading the census repeats it: `py_tools/check_assembly.py:148`, `py_tools/render_placement.py:317-319`, `py_placer/place_optimize.py:217`, `py_placer/place_reconstruct.py:727`.

The same scalar also flattens per-net netclasses. It is resolved once from the Default netclass alone (`py_router/list_nets.py:345-348`), while `check_drc` grades at `max(netclass(a), netclass(b), lc_a, lc_b, dru layer rule)`.

[...]

Issue #617, opened 2026-08-12
plane_region_connector, pcb_modification and fanout_clearance still price NPTH at the flat 0.20 fab floor


edgehero:
`obstacle_map.resolve_hole_clearance` (#616) reads the board's declared
`min_hole_clearance` for `add_drill_hole_obstacles` and
`plane_obstacle_builder`, and its docstring names the engines it does NOT
cover: `plane_region_connector` (taps / region joins / reconnects),
`pcb_modification`, and `placement/fanout_clearance` still hardcode
`NPTH_TO_TRACK_CLEARANCE`. On a board declaring above 0.20 those paths route
into a band `check_drc` now flags (measured on neo6502: 0.2263 mm against a
declared 0.25). The helper exists and is cached; each site is a one-expression
change, best made engine-by-engine with a pinning test per engine.

edgehero:
This looks resolved on `main`, flagging it rather than closing since it is your call.

`2cfaef64` ("read the board's min_hole_clearance where the engine CHOOSES where copper goes", merged via #641) is an ancestor of `50a26760`, and all three modules named in the issue now resolve the board's value:

- `py_router/plane_region_connector.py` calls `resolve_hole_clearance` at 5 sites
- `py_router/pcb_modification.py` at 6
- `py_placer/placement/fanout_clearance.py:311-313`, as `max(defaults.NPTH_TO_TRACK_CLEARANCE, resolve_hole_clearance(pcb_data, None))`

with four pinning tests: `tests/test_617_plane_connector_hole_clearance.py`, `test_617_pcb_modification_hole_clearance.py`, `test_617_placement_fanout_hole_clearance.py`, `test_617_origin_inheritance.py`.

One thing worth carrying into the closing note if you do close it: `obstacle_map.py:1553-1583` documents the sites that still price flat (`close_soft_joints`, `nudge_grazing_octolinear`, `nudge_vias_for_unresolved`, `legality.PartPads`) as a deliberate measured scope decision rather than leftovers, on the grounds that they are all-or-nothing repairs where raising the gate abandons the fix without removing the violation. That reads as intentional, but it is the kind of thing that gets re-reported later if the reasoning is only in a code comment.

edgehero:
Closing: verified fixed on `main`.

`2cfaef64` is an ancestor of `50a26760`, and all three modules named in the title now resolve the board's own value rather than the flat 0.20 floor:

- `py_router/plane_region_connector.py`, 5 call sites
[...]

PR #729, opened 2026-08-23
#725: the fanout-clearance repair pass prices every pair at its REAL requirement

### LINKED ISSUE #725: Placement: the fanout-clearance repair pass prices every pad pair at one flat scalar (#697 in a second channel)
`placement/fanout_clearance.py` is a second, independent different-net pad-pair
clearance channel, and it has the defect #697 just fixed in
`placement/legality.py`: it prices every pair at one flat scalar.

```
$ grep -c "local_clearance" py_placer/placement/fanout_clearance.py
0
```

`_pad_pair_shortfall` at `py_placer/placement/fanout_clearance.py:77-89`:

```python
def _pad_pair_shortfall(pads_a, pads_b, clearance):
    ...
            gap = _rect_gap((ax0, ay0, ax1, ay1), (bx0, by0, bx1, by1))
            if gap < clearance - EPS:          # one flat scalar
                pen += (clearance - gap)
```

It carries its own broad phase, its own seed baseline and its own candidate
gate, none of which the #697 model reaches — the pass builds `_Cap` 5-tuples
rather than `PartPads`, so nothing carries a per-pad floor:

| site | code |
|---|---|
| `:707` | `if gap < self.clearance - EPS:` (per-foreign-net component-pad shortfall) |
| `:775` | `if _rect_gap(cand_bbox, other.pad_bbox()) >= self.clearance:` (broad-phase reject) |
| `:1307` | `... < vr + clearance - EPS` (`nudge_vias_for_unresolved`) |

Consequence is the same one #697 was filed for: a pad pair separated by more
than `clearance` but less than `max(lc_a, lc_b)` — a fiducial keep-clear, a net
class, a `.kicad_dru` layer rule — is never charged, so the repair pass reports
nothing to fix on geometry `check_drc` flags. Reached from
`place_fanout_clearance.py:114` (`repair_fanout_clearance`), i.e. the
decoupling-cap repair step in the BGA-fanout pipeline.

**Suggested fix**: `placement/legality.PadClearanceModel` already resolves the
requirement exactly as `check_drc` does (`max(clearance, netclass a,
netclass b)` → `.kicad_dru` layer rules over the shared copper layers, which
REPLACE → `max(…, lc_a, lc_b)`), and is strictly inert on a board that declares
none of them. The `_Cap` pad tuples would need a per-pad floor the way
`PartPads.pad_floors` carries one, and the three comparison sites above would
take `model.pair(fa, fb)` instead of the scalar.

One caution carried over from #697: `:775` and `:707` are hot, so the reject
radius should be bounded by the two parts' own maxima (`PartPads.max_floor`'s
analogue), never the board-wide maximum.

Noted while fixing #697; deliberately left out of that PR to keep it to one
failure mode.

edgehero:
Closes #725.

`placement/fanout_clearance.py` — the decoupling-cap repair pass in the BGA fanout
pipeline — priced **every** different-net clearance requirement at one flat scalar. It
read `local_clearance` nowhere and imported neither `list_nets` nor `kicad_dru`, so it
never saw a netclass, a pad `(clearance …)` override, or a `.kicad_dru` layer rule.
This is #697 in a second channel; #697 fixed it in `placement/legality.py` and
deliberately left this file out to keep that PR to one failure mode (CLAUDE.md names
the gap).

## The defect, measured

`kicad_files/orangecrab_ext_pll.kicad_pcb`, staged with a sibling `.kicad_pro`
declaring a Default class of **0.4**, run at `--clearance 0.1`, graded with
`check_drc.py --clearance 0.4`. "Mover-scoped" = a `pad-pad`/`pad-via` item naming a
movable near-BGA cap.

| | |
|---|---|
| `check_drc`, at the board's own 0.4 | **79** mover-scoped `pad-pad`+`pad-via` across **32** caps |
| the pass | **17** violators, moved 48, printed **`0 unresolved`** |
| of the 32 flagged caps, scoring `graze_penalty` **exactly 0.0** | **16** |

Control: the same board with no `.kicad_pro` gives 26 violations across 13 caps and
**0** caps at zero — the blind spot appears only when the board declares something.

## A/B, same input, same grading terms

| graded at the board's own 0.4 | input | **before** | **after** |
|---|---|---|---|
| mover-scoped `pad-pad`+`pad-via` | 79 / 32 caps | **85 / 37 caps** | **8 / 9 caps** |
| the pass reported | — | 17 violators, `0 unresolved` | 26 violators, 1 unresolved (C20) |
| `segment-segment` (a cap pass must not touch tracks) | 448 | 448 | 448 |

The old pass made the board **worse** (79 → 85): it optimised against a cost function
blind to the real requirement, so it walked caps into violations it could not see.

## What changed

Reuses `legality.PadClearanceModel` unchanged (netclass → `.kicad_dru` REPLACE over the
shared layers → pad override), so this grades at exactly what `check_drc` grades at.
Converted: pad↔pad (`_pad_pair_shortfall`, `_pad_shortfalls`, the `_blocked_geom`
prescreen and its gate, the `base_cap_pad` seed baseline), via↔pad (`via_penalty`,
`_via_shortfalls`, the locked-part warning), track↔pad (`_seg_shortfalls` incl. its
cheap bbox reject), the whole of `nudge_vias_for_unresolved`, and the four prune radii
— which are documented as EXACT and become lossy the moment a requirement exceeds the
flat scalar.

Three things this had to get right:

* **The baselines move with the candidates.** `base_cap_pad` priced flat while
  candidates price at the requirement makes `_worsens_any_net` compare two different
  units, so the accept gate reads a pose as worse-than-seed on a pair that did not
  change. The test asserts the *currency*, not a placement count — reverting the line
  moves placements in a class-dependent direction and not by much.
* **The nudger and the grader share one helper.** `nudge_vias_for_unresolved`'s offender
  test and `via_penalty`'s grazing test both resolve through `_Repair.via_required`.
  Diverged, the nudger returns an empty offender list for a cap the grader flagged, and
  the cap stays unresolved forever with nothing printed.
* **A via's `PadFloor` is scoped to ALL copper, never `None`** — `pair()` does
  `fb.layers or ()`, and an empty shared set discards every dru rule. A track's is
  scoped to its OWN layer, reproducing `check_drc`'s single-layer REPLACE for
  pad-segment.

**No tuple shape widened.** `st.vias` (4), `cap.pad_rects()` (5), `cap_foreign_pads`
(6) and `cap_segs` (7) are all pinned positionally by existing tests and by
`animate_fanout_clearance.py`, and tests assign `st.vias` wholesale — so floors ride in
index-aligned parallel lists (the `PartPads` precedent) and identity-keyed maps, and the
per-pair requirement is memoised per (cap, neighbour) so `model.pair()` is resolved once
rather than per candidate pose. `_Repair.__init__` keeps its 10-positional signature.
Duck-typed `_FakeSt`/`_FakeCap` grade flat.

**Inert when the board declares nothing.** The model is dropped to `None` and every
converted site branches on that first, with the original body verbatim underneath.
`rp2350_fpga_eensy_prePlane` (model inactive): 2 placements, 1 resolved, 7 unresolved
`{C16,C17,C18,C19,C22,C23,C24}`, 0 via_moves — **identical to main**, and a counter
proves `PadClearanceModel.pair` is never called. Raw orangecrab (model *active* via 6
tracked fiducial overrides, none of which bind): 17 violators, 48 moved, 17/17, 0
unresolved — **also identical**.

**Disclosure.** The result dict gains `required` (rows `[cap, '<kind> <net>', mm,
source]`, the shape `legality.format_required_clause` renders) and `clearance_notes`,
read *before* the active-drop because a failed read of an unreadable sibling is exactly
what makes a model look inert. Printed from the **engine**, so the CLI and the plugin
both inherit it; neither front-end needed a change, since both already pass `pcb_file`.
It reports the union of seed-pose and final-pose charges — the pass succeeds by leaving
nothing charged, so a final-only report was empty exactly when the raised requirement
had done its work.

**Deliberately NOT converted**, each with the reason in a comment: the board-edge margin
(a different KiCad rule), courtyard overlap and its prune (mechanical extents, not
copper), hole-to-hole and `npth_clr` (net-independent fab floors — #617 *measured* that
raising `npth_clr` turns 1 move/1 connector into 0/0), the NPTH keep-out inflation, and
#549 track-scoped dru rules.

Also: registers `py_placer/place_fanout_clearance.py` in the gui-parity gate's
`CLI_MAINS` — it was never scanned, which matters now that this engine reads the sibling
`.kicad_pro` for its own arithmetic. And drops a duplicated driver call in
`tests/test_fanout_clearance.py`.

## Verification

`tests/test_725_fanout_clearance_pad_floors.py`, 35 tests, ~24 s, in-process. Every
assertion carries a `# MUTATION:` comment naming the line whose reversion kills it, and
every test that talks about a band asserts it is IN that band first — two of mine passed
without reaching the code they named until those spies went in.

* **27 mutations, all killed.** One row per converted site; a site with no row is a site
  with no test. The first run killed 13 and **six survived** — all holes in my own
  tests, not the code.
* `run_all.py --fast`: **288 passed, 0 failed**. (The new file was initially bucketed as
  integration and skipped, because its own docstring contained one of the three marker
  words `is_integration` greps for. Fixed with the documented opt-out.)
* Regression set green: `test_fanout_clearance`, `test_617`, `test_370`, `test_628`,
  `test_431`, `test_697`, `test_placement_pad_legality`.
* Parity gates green: `test_cli_postpass_coverage` (with the new registration),
  `test_engine_kwarg_parity`, `test_manifest_plan_parity`, `test_cli_defaults_parity`.
  `validate_pcb_data.py` on the override-carrying board: `PASS: pcbnew data matches file
  parse`, so the GUI's pcbnew path feeds the model the same overrides.
* Perf: no meaningful change, and — the claim that matters — **results identical to
  main** on both boards across 5 reruns, every result key.

An independent parity check compared this module's resolvers against `check_drc`'s over
**6.59M pair combinations, 5,953 real violation records and 139,902 eff rows: 0
mismatches**, on 9 board configurations (raising and relaxing dru rules, multi-netclass,
full-REPLACE, `*.Cu` and `F&B.Cu` wildcards).

## Review rounds

Three adversarial rounds plus a fact-check, each of which found something real:

1. The branch charged a **copper-less NPTH pad's** keep-clear override — moving caps to
   clear copper that does not exist, and contradicting the model's own inertness rule,
   which refuses to *activate* for such an override. Also fixed the locked-part
   warning's raw pad rect, the only pad-geometry site in the file without
   `rect_rotation` (0.14 mm under-block on a 45° locked pad).
2. The netclass term is layer-blind, so it **amplified a pre-existing off-layer phantom**
   3× — `cap_segs` is pruned on an F/B side collapse that files an `In1.Cu` track under
   `'F'`. R17/R18/R5's entire graze is that phantom: 0.0821 mm each flat, 0.6991/0.6991/
   0.5686 raised, 0.0821 again after the fix. Removing the phantom itself means fixing
   the side collapse — filed, not folded in.
3. A **relocated** via left the radius map and was graded at the prune over-reach; and
   the copper-less-pad fix was incomplete — it zeroed that pad's own override while the
   *partner's* netclass still flowed through.
4. A fact-check of my own commit messages: the code came through clean, but two claims
   were false and five overstated. `a602687e` records exactly what was wrong and fixes
   the three comments that ship.

## Known gaps, filed rather than folded in

* `check_drc` honours an NPTH hole pad's own `local_clearance` on the copper-to-hole rule
  (#505); this pass does not. Measured on `ulx3s` AUDIO1: 1.050 mm modelled vs 1.250 mm
  required, a 0.200 mm under-block. It is the hole rule, net-independent, and interacts
  with the #617 balance.
* `self.segments` records an F/B side rather than a layer, so inner-layer tracks are
  compared against F-side cap pads at all. Now capped at the flat scalar; removing it is
  a separate change.
* Pre-existing and unchanged: `__init__` uses `default_via_size` for a size-0 via while
  the nudger uses `(v.size or 0.5)`, so on such a board the grader and the nudger already
  resolve different keep-outs. This PR unified the *clearance* term through one helper
  and left the radius alone.

### DIFF
--- a/py_placer/placement/fanout_clearance.py
+++ b/py_placer/placement/fanout_clearance.py
@@ -1339,9 +1339,10 @@ def repair_fanout_clearance(pcb_data: PCBData, pcb_file: str,
     print(f"BGAs: {', '.join(st.bga_refs) or '(none)'}  "
           f"fanout vias: {len(st.vias)}  "
           f"movable near-BGA caps: {len(st.caps)}")
-    # #725: the early returns carry the same key set as the full one, so a
-    # caller reading result['required'] / ['clearance_notes'] does not have to
-    # special-case a no-op board.
+    # #725: the early returns carry 'required' / 'clearance_notes' too, so a
+    # caller does not have to special-case a no-op board for them. (They still
+    # omit 'via_moves' / 'new_segments', as they always have -- every caller
+    # reads the dict with .get.)
     if not st.vias:
         print("No vias on the board - run this AFTER bga_fanout.py.")
         return {'placements': [], 'resolved': [], 'unresolved': [],
--- a/py_placer/placement/fanout_clearance.py
+++ b/py_placer/placement/fanout_clearance.py
@@ -43,8 +43,9 @@ import routing_defaults as defaults
 from bga_fanout.grid import analyze_bga_grid
 from placement.parser import extract_courtyard_bboxes, extract_locked_refs
 from placement.utility import compute_footprint_bbox_local, snap_to_grid
-from placement.legality import (BoardOutlineGate, point_to_seg_dist, rect_gap,
-                                ring_is_rect, rotate_local_bounds)
+from placement.legality import (BoardOutlineGate, PadClearanceModel, PadFloor,
+                                format_required_clause, point_to_seg_dist,
+                                rect_gap, ring_is_rect, rotate_local_bounds)
 
 ROTATIONS = [0.0, 90.0, 180.0, 270.0]
 EPS = 1e-6
@@ -74,18 +75,26 @@ def _point_to_rect_dist(px, py, rect):
     return math.hypot(dx, dy)
 
 
-def _pad_pair_shortfall(pads_a, pads_b, clearance):
+def _pad_pair_shortfall(pads_a, pads_b, clearance, effs=None):
     """Sum of different-net pad clearance shortfalls between two movable parts'
     pad rects (#275) -- the mover-vs-mover analogue of pad_penalty. Same-net
-    pad pairs are fine (shared rail copper may touch)."""
+    pad pairs are fine (shared rail copper may touch).
+
+    `effs` (#725) is the per-pair REQUIRED clearance, `effs[ia][ib]` aligned
+    with pads_a x pads_b -- a pad override / netclass / .kicad_dru layer rule
+    can put a pair's requirement above (or, for a relaxing rule, below) the
+    flat scalar. None keeps the flat path, which is what a board declaring
+    none of them takes."""
     pen = 0.0
-    for (ax0, ay0, ax1, ay1, anet) in pads_a:
-        for (bx0, by0, bx1, by1, bnet) in pads_b:
+    for ia, (ax0, ay0, ax1, ay1, anet) in enumerate(pads_a):
+        row = None if effs is None else effs[ia]
+        for ib, (bx0, by0, bx1, by1, bnet) in enumerate(pads_b):
             if anet == bnet:
                 continue
             gap = _rect_gap((ax0, ay0, ax1, ay1), (bx0, by0, bx1, by1))
-            if gap < clearance - EPS:
-                pen += (clearance - gap)
+            clr = clearance if row is None else row[ib]
+            if gap < clr - EPS:
+                pen += (clr - gap)
     return pen
 
 
@@ -134,7 +143,7 @@ class _Cap:
     for decoupling caps).
     """
 
-    def __init__(self, fp, courtyard_local):
+    def __init__(self, fp, courtyard_local, model=None):
         self.ref = fp.reference
         self.side = 'B' if (fp.layer or '').startswith('B') else 'F'
         self.seed_x, self.seed_y = fp.x, fp.y
@@ -145,6 +154,21 @@ class _Cap:
         # separate paste-only "pads" (e.g. gkl_misc C_0201_0603Metric), which are
         # not copper and must not enter the clearance/attraction geometry.
         self.pads = []
+        # #725, mirroring PartPads (legality.py): the per-pad clearance FLOOR
+        # rides in a parallel list index-aligned with self.pads, never widened
+        # into the pad tuple -- pad_rects()' 5-tuple is unpacked positionally by
+        # tests and by animate_fanout_clearance.py. `max_floor` is the
+        # over-reach a broad phase must use (a .kicad_dru rule REPLACES, so it
+        # can also LOWER a pair; a prune must never under-reach).
+        # NOTE the copper filter below stays _Cap's own `endswith('.Cu')` test
+        # and is deliberately NOT unified with legality._pad_carries_copper,
+        # which additionally rejects np_thru_hole. Unifying would change which
+        # pads enter pad_rects/pad_bbox/attraction and desynchronise the
+        # n_copper cap-detection test in _Repair.__init__ -- a placement change
+        # dressed as a clearance fix. Building the floors against THIS rule
+        # makes the alignment true by construction.
+        self.pad_floors = []
+        self.max_floor = 0.0
         for p in fp.pads:
             if not any(str(l).endswith('.Cu') for l in p.layers):
                 continue  # paste/mask-only aperture, not copper
@@ -156,6 +180,12 @@ class _Cap:
             half_x = hx * c + hy * s
             half_y = hx * s + hy * c
             self.pads.append((off_x, off_y, half_x, half_y, p.net_id))
+            if model is not None:
+                fl = model.pad_floor(p)
+                self.pad_floors.append(fl)
+                mf = model.max_floor(fl)
+                if mf > self.max_floor:
+                    self.max_floor = mf
         self.local_bounds = courtyard_local
         # Rotation-only geometry is reused across every candidate position
         # (millions of times on a dense board), so memoize it per angle and
@@ -298,6 +328,29 @@ class _Repair:
         self._edge_active = self.edge_gate.active
         self._max_disp_cap = max_displacement_cap
         self.clearance = clearance
+        # #725: `clearance` is ONE flat scalar, but KiCad's different-net
+        # requirement for a pair is max(clearance, netclass a, netclass b),
+        # then REPLACED by any .kicad_dru rule on the layers the two share,
+        # then raised by either item's own pad `(clearance ...)` override.
+        # This pass is the second channel of #697's defect: priced flat, a pair
+        # separated by more than `clearance` but less than its real requirement
+        # is never charged, so the repair reports nothing to fix on geometry
+        # check_drc flags. PadClearanceModel resolves it exactly as check_drc
+        # does and is STRICTLY INERT on a board declaring none of the three.
+        # `notes` must be read BEFORE the active-drop: a FAILED read (an
+        # unreadable sibling) is exactly what makes the model look inert.
+        _model = PadClearanceModel.for_board(pcb_data, clearance, pcb_file)
+        self.clearance_notes = list(_model.notes)
+        self._floors = _model if _model.active else None
+        # Pad-layer scope for the non-pad items below. A via spans copper, so
+        # it is scoped to ALL copper: check_drc's pad-via path passes
+        # layers_b=None meaning "scope to the pad's own layers", and
+        # intersecting with all-copper reproduces that exactly (blind vias
+        # included). Never None here -- PadClearanceModel.pair does
+        # `fb.layers or ()`, and an EMPTY shared set discards every dru rule.
+        self._all_cu = frozenset(pcb_data.board_info.copper_layers or ())
+        self._via_floor: Dict[int, PadFloor] = {}
+        self._seg_floor: Dict[Tuple[int, str], PadFloor] = {}
         # #617: KiCad's copper-to-hole rule is the BOARD's `min_hole_clearance`
         # whenever it declares one above the 0.20 NPTH fab floor -- the same
         # value check_drc grades at. The NPTH keep-out rects below were grown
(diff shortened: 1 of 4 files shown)


drandyhaas:
Thanks — I went through this carefully and reproduced your measurements independently. First, what holds up, because the engineering here is genuinely careful:

- `tests/test_725_fanout_clearance_pad_floors.py`: 39 tests, OK.
- `run_all.py --fast`: **288 passed, 0 failed** — exactly as you report.
- **Inertness confirmed on the real CLI.** `orangecrab_ext_pll` and `rp2350_fpga_eensy_prePlane` with no sibling `.kicad_pro` produce boards **byte-identical to main modulo UUID**. That's the claim I cared most about and it's solid.
- GUI parity is fine: `fanout_gui.py` passes a real `pcb_file` and reads the result with `.get`, and you print from the engine, so both fronts inherit the disclosure.
- Your headline A/B reproduces exactly: staged at Default 0.4, run `--clearance 0.1`, graded at 0.4 — mover-scoped `pad-pad`+`pad-via` goes input **79**/32 caps → main **85**/38 → this branch **8**/9, with `segment-segment` 448 on all three. Main really does make that board worse.

The reuse of `PadClearanceModel` rather than mirroring it, the index-aligned parallel lists to avoid widening pinned tuples, the identity-keyed floor maps with the tuple-lifetime hazard called out, one shared `via_required` so the nudger and grader can't diverge — all of that is right, and the known-gaps section is the kind of disclosure I want to see.

## The problem: `--clearance` is a ceiling here, and the model treats it as a floor

This step's writeback clamps the netclass **down** to `--clearance`. Staged at 0.4 and run at `--clearance 0.1`, the output project declares **0.1** — I checked, on both arms:

```
input  project: Default 0.4
main   output:  Default 0.1   min_clearance 0.1
branch output:  Default 0.1   min_clearance 0.1
```

So `--clearance` behaves as the ceiling documented for the routing steps (`min(class, --clearance)`), while `PadClearanceModel.pair` takes `max(base, ncl)`. The branch moves caps to satisfy 0.4 and then ships a project saying 0.1.

That inverts the A/B. Graded at **each board's own declared floor** (`check_drc` with no `--clearance`, so each board is graded at what it ships):

| | main | this branch |
|---|---|---|
| orangecrab | **1** violation | **1** violation |
| rp2350 | 1 | 2336 |

orangecrab is a **tie**. The 79 → 8 improvement only appears when grading at 0.4, a class the output has already clamped away — which is the phantom-violation case CLAUDE.md warns about.

And the clincher: **every charged row on both boards is `(netclass)`** — 74/74 on orangecrab, 43/43 on rp2350. Not one row from a pad `local_clearance` override or a `.kicad_dru` rule. So the entire demonstrated effect comes from the one source whose semantics say *clamp*, and none from the two sources where I think you're clearly right.

## Second board: the pass freezes

`rp2350_fpga_eensy_prePlane`, staged identically (Default 0.4, `--clearance 0.1`), mover-scoped at 0.4: input 15 → main **11** → branch **15**. The branch moves zero caps.

Dose response on the branch, varying only the declared class:

```
class 0.10 -> 2 moves     0.25 -> 0 moves
class 0.15 -> 1 move      0.30 -> 0 moves
class 0.20 -> 0 moves     0.40 -> 0 moves
```

Monotone. Monkeypatching `_Repair._worsens_any_net` to return `False` restores **7 moves**, so the cause is the #441/#445 per-net "don't worsen any net vs seed" veto binding much harder against an inflated halo: at a large requirement nearly every candidate worsens *something*, so everything is rejected. I read that as a symptom of the floor/ceiling inversion rather than an independent bug.

(The 2336 in the table above is downstream of the freeze: zero moves takes the "unchanged copy" path at `place_fanout_clearance.py:134`, which skips `fix_project_for_output`, so it ships the 0.4 spec unmet. That gate is pre-existing and arm-independent — it only bites because the pass froze.)

## Questions

1. **Should the netclass term be capped at `--clearance`, with pad overrides and `.kicad_dru` rules still outranking it?** That's the asymmetry CLAUDE.md already describes, and it would make the split clean: dru and pad `local_clearance` genuinely outrank `--clearance` and survive the writeback, so pricing at them is strictly more correct than main and fixes a real blind spot (#697's 1.016 mm fiducial keep-clear reporting 0 pairs to fix). Netclass is the odd one out. The catch is that this makes the branch numerically inert on both boards you measured — so it'd need a board where a pad override or a dru rule actually binds near a BGA cap to demonstrate value. You note orangecrab's 6 fiducial overrides don't bind; is there a corpus board where one does?

2. **Should a cap-*placement* step be clamping netclasses at all?** It lays no copper. If the answer is no, then the writeback is the bug, your pricing is right as-is, and the freeze becomes the thing to fix on its own terms. I genuinely don't know which way this should go, and it decides which fix is correct — so I'd rather ask than pick.

Separately, on evidence: both boards here are synthetic (a staged 0.4 class), and per the placement A/B rules in CLAUDE.md I'd want ≥3 distinct boards, paired and directional, before this changes behaviour on declaring boards.

Minor, non-blocking:
- `from check_drc import pad_copper_layers` sits inside the per-pad loop in `_Cap.__init__` — worth hoisting.
- Cross-module import of the private `legality._pad_carries_copper`.
- `required_rows`'s `best()` closes over the loop variable `cap` — correct, but fragile to a later edit.

None of this is a criticism of the mechanism, which I think is right and which I'd like to land. It's the `--clearance` semantics I want to settle first.

edgehero:
Thanks, I reproduced all of it. Process note first: I'm deliberately keeping these to one defect per PR rather than one big branch, which is why there's a row stacked on this one. The cost is that your review landed with ten follow ups already on top, so some of it has moved. I re-ran every measurement at the tip of that row.

## Three of your findings are already fixed, by PR #739 (issue #731)

Commit `c3598400`, the next commit after the branch you reviewed. Every number that moved traces to it alone; the other nine PRs moved none.

| your finding | at #729 | at the row tip |
|---|---|---|
| rp2350 freeze, 0 caps moved | 15 to 11 to **15**, dose 2/1/**0/0/0/0** | gone: 12/7, dose 1/2/2/2/1/1 |
| rp2350 graded at its own floor | main 1, branch **2336** | main 1, tip **0** |
| rp2350 mover scoped | worse than main | better than main |

The off layer phantom #731 had filed rather than folded in. Your diagnosis survives though: disabling `_worsens_any_net` still takes 1 move to 5, and forced to a declared 0.80 where the tip does freeze, it still ships the unmet spec and grades 5754. That gate wants its own issue.

## Two stand

The headline A/B reproduces digit for digit at the tip (79 to 85 to 8, `segment-segment` 448 on all arms), and the writeback inversion stands on all three arms, input 0.4 output 0.1.

## Your corpus question: no

Worse than "doesn't bind", and for a more specific reason than I first wrote (see the edit note at the bottom).

* `required` is empty on all 22 tracked boards at 0.1/0.2/0.25. No override, rule or netclass row.
* orangecrab prices 8 pairs at `pad override` 0.375 and charges none, tightest 1.727 mm clear.
* Toggling `PadClearanceModel.active` gives identical output on every board at every clearance, so it is inert on the whole corpus, not just your two boards. Your prediction was right.
* No board carries a `.kicad_dru` in git, and the two stress corpus boards that do ship one (`mez_rx`, `storm_tracker`, fetched by `assemble_corpus.py`, which carries `.kicad_dru` deliberately for #498) both resolve to `layer_rules = {}`. Every clearance rule in them is conditioned (`A.Net != B.Net`, netclass, `memberOfGroup`) and `parse_dru_layer_clearances` skips all of them with a note.

So the `layer rule` source has never fired on a real board, and not for want of importing dru files: the parser drops the rules the ones we import actually contain. That looks like its own issue against #498, and I'd rather raise it separately than fold it in here.

Also worth saying: #697's own witness doesn't overlap this channel. esp_prog's 1.016 mm fiducial is real and tracked, but esp_prog has no BGA, so the pass never runs there.

## A third option

CLAUDE.md already specifies it: `--clearance` given means `min(class, --clearance)` and the writeback clamps down; omitted means the class is honoured and the writeback preserves. `pair_with_source` does `max(base, ncl)`, the omitted branch, while the writeback does the given branch, in the same run. So it contradicts the documented rule rather than being a judgement call.

It cannot currently do either, because `place_fanout_clearance.py` declares `--clearance` with a non None default and so cannot tell given from omitted. It is also the only placement CLI that doesn't resolve board first (`place_optimize`, `place_seed`, `place_reconstruct`, `place_portfolio`, `converge` all use `list_nets.board_floor_knobs`) and the only one that writes back a DRC floor at all, which answers your second question empirically. Give it `default=None` and `board_floor_knobs` like its five siblings and both branches hold without picking either.

One correction to the Q2 premise: this step does lay copper. The #313 via nudge emits connector segments and moves barrels. Rare, zero via moves on all 22 boards at file poses, but it is why a writeback exists here and not in the siblings.

## Proposed follow ups

You merged the stack while I was measuring, so these are new PRs off current main rather than changes to this one:

1. `default=None` plus `board_floor_knobs` in `place_fanout_clearance`, netclass capped when the flag is given. Self contained, and it is the semantics fix.
2. The zero moves writeback gate, which still ships an unmet spec.
3. `parse_dru_layer_clearances` skipping every conditioned rule, which is why #498 has never fired on a real board.

Agreed on three distinct boards paired and directional as the bar. I can't meet it on real boards for the pad override channel, because it has no real witness in the corpus, and I'd rather say that than stage three synthetic boards and call it a corpus. For the dru channel a staged sibling would do it: `(rule outer_wide (layer "F.Cu") (constraint clearance (min 0.25mm)))` beside orangecrab makes the arms differ, 2 `layer rule` rows, resolved 18 vs 17.

Minor items: `pad_copper_layers` is already hoisted (incidentally, by #739); the private `_pad_carries_copper` import and the `best()` closure over `cap` are still there and I'll pick them up in the first follow up.

---

*Edited: my first version of this said there are zero `.kicad_dru` files in the repo and that `layer rule` is therefore unreachable by construction. That was measured off `git ls-files` and is misleading. The stress corpus does import dru files, deliberately and for #498. The finding is narrower and I've restated it above: the two boards we import both have conditioned rules that the parser skips. Also corrected: I originally proposed rebasing this PR, which no longer applies now that it is merged.*

Issue #703, opened 2026-08-22
Placement: no pre-route predictor has been correlated with routed blocking; the quoted r-values are distance-to-truth


edgehero:
The placement instruments are asserted to predict routability, and none of them has been correlated with the routed outcome.

Every correlation number quoted in the skills and docs is measured against distance-to-the-original-placement or against the gap a human left, not against `blocking`. The crossings figure of r = +0.78 and the overlap-area figure of r = +0.723 are both distance-to-truth. The corridor law's r = +0.41 to +0.90 on 8 of 8 boards is against the human's gap. CLAUDE.md's own "What a placement run is FOR" section says distance-to-truth is the wrong headline and `blocking`, meaning unrouted plus broken, is the right one.

The one study with routed evidence is two boards from June 2026, and its conclusion (`docs/placement-optimization.md:430-435`) is:

> Proxy-routability correlation is weak ... the variant with the best crossing reduction failed 2 nets. Any production version should rank candidate placements by an actual trial route.

`tests/test_placement_ab.py` never routes; it grades placement metrics with independently re-derived placement metrics, which is a good anti-self-deception design and a non-answer to this question. `tests/test_placement_probe.py` routes but is a two-arm A/B, not a correlation study.

Meanwhile the data for a first pass already exists in-repo. Recorded runs carry both a pre-route `render_placement --json-out` document and a post-route `board_score` result. Joining six of them:

| board | crossings | hpwl | blocking | vias |
|---|---|---|---|---|
| castor | 453 | 3069.0 | 0 | 545 |
| neo6502 | 172 | 1129.7 | 79 | 427 |
| piantor | 84 | 1990.0 | 0 | 162 |
| urchin | 191 | 3231.4 | 3 | 175 |
| tigard | 670 | 1988.3 | 3 | 377 |
| esp_prog | 36 | 221.9 | 0 | 28 |

Tie-corrected Spearman against `blocking`: crossings +0.339, hpwl -0.062. Leave-one-out span for crossings is +0.053 to +0.632, so at n=6 it is not a measurement. The one strong signal is crossings against `vias`, +0.714, meaning the shipped metric predicts a quality tie-break rather than the headline. And on the one recorded multi-candidate slate, crossings against vias is -0.400 within a single board, the opposite sign, so pooling across boards has been measuring board size.

[...]

PR #784, opened 2026-08-28
#703: what predicts routed blocking, measured within board

edgehero:
Refs #703. **This lands the measurement apparatus and the first measurement; it
does NOT close the issue.** The declared table is 6 boards and 4 were routed, so
every N and every p-value below is over 4 — never over the planned count. The
rows themselves are not committed (see *What a reviewer can check* below).

## What #703 asked, and what was actually there

Every correlation this repo quoted was measured against
*distance-to-the-correct-placement* or against *the gap a human left*, never
against routed `blocking`. Two things turned up while building the instrument
that are worth stating before any result:

**The recorded corpus cannot answer the question at all.** Both documents have
been written side by side in every recorded run for months and nobody had joined
them. Joining them properly gives 7 rows and 5 refusals — **one placement per
board**, which is n=1 per board and below the arithmetic floor. That is a
stronger answer than the ρ = +0.339 the issue quotes, because it does not depend
on which score file someone reaches for.

**The issue's own six-row table is joined by filename, and a filename is not a
verdict.** `RUN_STATE.board_sha` is copied from the *last* ledger row whether or
not that lap was accepted; on `wk/run23/tigard` it names the output of a rejected
lap whose own lever text ends *"Step back to routed.kicad_pcb"*, while the same
file's `quality.vias` reports the accepted board's 377. The ledger's last
`final: true` row is the authority. Applying it moves three of the harvested
rows (urchin's blocking flips 3 → 0) and un-refuses two more.

*(I got this wrong first, in the other direction, and said so in the commit that
fixed it: an earlier commit message on this branch claims the published 377 was
wrong. It was not. The ledger is the authority and the published value was
right.)*

## The result

80 placements, each routed once with an argv frozen before any variant existed,
each graded by `board_score`. Spearman **within each board**, aggregated by a
sign test across boards. Full write-up: `docs/placement-predictors.md`.

**The predictors that rank routed `blocking` are the LEGALITY counts. The
classical placement proxies do not.**

| predictor | right / wrong | median ρ | verdict |
|---|---|---|---|
| `pad_*_pairs`, `body_overlap_pairs`, `pad_shortfall` | 4 / 0 | +0.724 | **passes** |
| `courtyard_blocking_pairs` | 4 / 0 | +0.652 | **passes** |
| `pad_copper` | 4 / 0 | +0.615 | **passes** |
| `hole_shortfall`, `hole_conflicts` | 3 / 0 | +0.42 | **passes** |
| `halo` | 3 / 1 | +0.409 | fails |
| **`crossings`** | **3 / 1** | **+0.396** | **fails** |
| `overlap_area` | 3 / 1 | +0.372 | fails |
| **`hpwl`** | **2 / 2** | **+0.158** | **fails** |

- **`pad_copper` — the one pre-route number that already refuses anything
  (`loop_driver`'s L2 gate) — is validated**, 4 of 4 boards. That gate was right.
- **`hpwl`, which the drivers gate on, is the worst of the classical trio**
  (−0.525 esp_prog, +0.739 splitflap, −0.045 tigard, +0.360 watchy; p = 1.0).
  Its justification is the distance-to-truth argument, which is untouched — but
  nobody should call it a routability gate. **This PR does not remove it.**
- **`crossings` fails and its sign flips** (−0.179 on esp_prog).

**The circularity control changed the answer for `crossings`.** The
realistic-end sampler is `portfolio.generate`, whose quench minimises crossings.
Excluding those rows `crossings` goes 4/0 and passes; including them 3/1 and
fails. So neither arm is reported as the answer for it. The legality predictors
pass in **both**.

**How much to believe a PASSES: 9.0–18.0%, median 11.5%.**
`--shuffle-control 200` permutes truth within each board and measures the rule's
own false-positive rate at N=4 over the 21 predictors defined on all four
boards. It also found that `cross_side_stacks`, defined on one board, passed
**100%** of shuffles — `MIN_SIGN_BOARDS = 3` exists because of that measurement,
not because it seemed wise. (An adversarial review caught this stated as
"13–17%" in an earlier revision of the doc, which the documented command does
not produce; both the doc and this body carry the measured range now, and the
control was also corrected to permute over the same deduplicated sample the ρ
values use.)

## What a reviewer can check, and what it costs

*(Updated after review — the rows file has been withdrawn, see below.)*

| cost | what it proves |
|---|---|
| 0 s | `python3 tests/test_703_rank_stats.py` — 63 kernel assertions + 82 external checks |
| ~1 s | `python3 tests/test_703_predictor_claims.py` — every r-value in the repo names its dependent variable |
| ~2 s | `python3 tests/test_703_predictor_harvest.py` — 17 recorded-run shapes, synthesized in tempfile |
| ~6 s | `predictor_study.py --verify-row esp_prog:authored` — regenerate one placement, re-measure its predictors, diff against the recorded literals |
| ~51 s | `python3 tests/test_703_predictor_regen.py` — the same for four declared variants across two boards |
| ~60 s | `python3 tests/mutate_703.py` — 29 rows, 27 killed, 2 expected survivors, 0 broken |
| **8.8 h** | the 80 routes (~2.5 h wall clock at `-j 4`), after which `--from-rows` gives every statistic for free |

Every study board is git-tracked and `portfolio.generate(..., only=i)` is
byte-identical by contract, so every row is regenerable from this repo.

**The rows are no longer committed.** On review drandyhaas asked for
`tests/stress/predictor_rows.json` to go, having verified the premise rather
than trusting it (`--task esp_prog:authored` rebuilt a row in 4.4 s, matching).
The cost is stated in the doc: no number in `docs/placement-predictors.md` has
an automated change detector any more, and re-deriving them is 8.8 h. What is
guarded instead is the **rig** — `test_703_predictor_regen.py` — so a change to
the placement engine, the predictor extraction or the route argv cannot silently
move what the study measured.

The same review found `--verify-row` was a **dead flag**: declared, documented,
advertised in this table, and read zero times, so passing it silently started
the whole study. It is implemented now. And the withdrawn rows pinned **CRLF**
board hashes, so their gate was green only on a Windows checkout; the
replacement pins `poses_sha256`, computed from parsed poses and identical on
either checkout.

## What this PR does NOT claim

- Not 6 boards. `sonde_u` and `kit-dev-coldfire-xilinx_5213` are declared and
  were not routed; the doc names them.
- `perturb-pile` timed out at 2400 s on 3 of 4 boards, systematically excluding
  the most-damaged sample. watchy lost 6 of 20 slots to duplicate placements
  (effective K = 13). Both disclosed per board.
- The six predictors sharing a median of +0.724 are one quantity through six
  counters, not six findings.
- **No rank key and no gate is changed.** `rank_key`'s crossings slot,
  `rule1_check`'s crossings clause and `loop_driver`'s prohibition are all
  untouched and now DISCLOSED at both code sites, with the exit criterion
  pre-registered in the doc.

## Doc corrections

18 sites annotated, not deleted — the r-values are real measurements of a real
quantity and only the label was wrong. One deletion: the "Routability signal at
PCB scale" cheat-sheet cell in `docs/placement-optimization.md`, which asserted
crossings is the "best single PCB proxy" with no measurement behind it.
`tests/test_703_predictor_claims.py` makes the sweep a standing gate, registered
in both directions so it cannot rot.

## Things that went wrong, and what caught them

Recorded because each is a change detector for the next person:

- The sampler produced **nine rows carrying four distinct placements** —
  `perturb` clips a dose to the feasible travel and three doses clipped to one
  value. Killed the run at nine rows rather than spend four hours on it. Doses
  are now fractions of the *feasible* travel, `wrong_side`/`swap`/`pile` are
  undosed kinds contributing one variant each, and a duplicate guard collapses
  any survivor and names it.
- `poses` portfolio candidates were **barren on every even index**, silently
  costing a quarter of K. Now jitter-only.
- The kernel's anti-pooling guard was **documented and not executed**:
  `board_rho` on the six recorded runs returned exactly the pooled +0.339 the
  docstring calls the trap, and the introspection check whitelisted the two
  functions that violated it.
- The first mutation battery reported 16 kills, **all of them the kernel's own
  self-test raising and aborting the file** before any external check ran. The
  table now carries a measured external-only column.
- The doc-derivation gate's regex **did not match rows written in bold** — the
  rows the doc emphasises — so flipping `crossings` from fails to passes left it
  green.

Full `tests/run_all.py`: **436 passed, 1 failed, 1 timed out**. The failure is
`test_763_kicad_locate.py`, which fails identically on `upstream/main` (a
Windows path-separator assertion) and is untouched here. The timeout was
`test_obstacle_map_balance.py` at its own 1200 s budget while the study was
still running; re-run alone it exits 0, per the runner's own instruction not to
record a timeout as a failure without that.

https://claude.ai/code/session_01TRf6KQc2F9Jwtv5DMFZaS9


drandyhaas:
Reviewed. The study itself holds up — I re-derived the headline from `predictor_rows.json` with my own Spearman and the primary arm reproduces exactly (`pad_copper` +0.615 PASSES, `crossings` +0.396 fails with esp_prog −0.179, `hpwl` +0.158 fails with esp_prog −0.525 / watchy +0.360). My first pass disagreed on watchy until I applied the dedup and `perturb-pile` exclusions, both of which you disclose — my error, not yours.

I also confirmed the no-behaviour-change claim rather than taking it: AST-comparing each engine file against main with docstrings stripped, `portfolio.py` / `reconstruct.py` / `routability.py` are identical. `placement_driver.py` flagged, but with every string constant blanked its AST is identical too — 2 of its 644 string constants changed, both prose. Nothing in `py_router/`.

**Andy's decision: the committed rows file should not stay in the repo.** The results are regenerable, and I verified that rather than relying on the contract: `--task esp_prog:authored` rebuilt the row in 4.4 s and all 28 predictors, `truth.blocking`, `truth.quality`, `argv_sha` and `poses_sha256` matched the committed row exactly (`routed_board_sha` differs, which is the expected per-run UUID behaviour).

So: please drop `tests/stress/predictor_rows.json` and rewire what depends on it. Three things worth knowing before you do.

**1. `--verify-row` is a dead flag.** It is declared at `predictor_study.py:793` and named in the usage text at line 74, and `args.verify_row` is read **zero** times in the file (every other flag is read: `task` 2, `from_rows` 3, `append` 37, ...). So the row in your reviewer table —

| ~30 s | `predictor_study.py --verify-row esp_prog:portfolio-3` — regenerate one placement, re-measure its predictors |

— does not do that. Passing it falls through to a full run: I ran `--verify-row esp_prog:authored` and it printed `111 task(s) to run`. That is worse than an error, because the advertised cheap check silently starts the whole study. `--task BOARD:VARIANT` is the working single-row path.

**2. The rows pin CRLF hashes, so `test_703_predictor_rows.py` fails on any LF checkout.** It hashes the tracked board's raw bytes with no line-ending normalisation and compares to `provenance.input_board_sha`. On macOS it fails for 3 of 4 boards; I checked the hypothesis directly and **all four** committed shas are the CRLF hash of the tracked file:

```
kicad_files/esp_prog.kicad_pcb          rows say 165302e6a4  -> CRLF MATCHES
kicad_files/splitflap_driver.kicad_pcb  rows say a9fd165133  -> CRLF MATCHES
kicad_files/tigard.kicad_pcb            rows say fa6c4cc6f9  -> CRLF MATCHES
kicad_files/watchy.kicad_pcb            rows say bcc1b4d3a2  -> CRLF MATCHES
```

The gate is therefore green only on a Windows checkout. Same family as the pre-existing `test_763_kicad_locate` failure, pointing the other way. Worth fixing even if the file goes, since the harvester will keep stamping platform-dependent shas.

**3. What actually depends on the file:** 3 of the 4 suites (`rank_stats`, `claims`, `harvest`) pass without it — I removed it and re-ran. Only `test_703_predictor_rows.py` needs it, and `--from-rows` needs it. So dropping the file costs the doc's numbers their cheap change-detector, which is the real trade and worth saying out loud in the doc: with the rows gone, re-deriving the table is an **8.8 h** job, not the ~5 h the PR quotes (summing `provenance.total_seconds` over the 87 rows gives 8.77 h, median 217 s/row, max 2012 s on `tigard:perturb-wrong_side`).

Suggestion, but your call on shape: keep the doc's table as the finding, keep the three suites that stand alone, and replace the fourth with something that regenerates a *small declared subset* on demand (a couple of `--task` rows, seconds each) rather than shipping all 87. That keeps a real change detector without the artifact.

One correction to my own earlier read, since it bears on how you weigh this: I first reported all four suites passing. They do not — I had written `$?` after a `$(basename ...)` substitution, which resets it, so I was reading `basename`'s status. `test_703_predictor_rows.py` exits 1 here.

edgehero:
All three findings reproduced here before I touched anything, and the artifact is gone. Pushed as `acad9c0c`.

**`--verify-row` did nothing, and I advertised it.** Confirmed: `args.verify_row` read zero times, so it fell through to a full run. That is the worst version of the bug — the PR's own reviewer table sold it as the cheap check and it would have started an 8.8-hour study. It is implemented now, against the same declared subset the new test uses:

```
$ python3 -X utf8 tests/stress/predictor_study.py --verify-row esp_prog:authored
  esp_prog:authored: regenerated and matched every recorded field (5.6s)
```

**The shas were CRLF.** Confirmed all four, both ways:

```
kicad_files/esp_prog.kicad_pcb          crlf=MATCH  lf=no
kicad_files/splitflap_driver.kicad_pcb  crlf=MATCH  lf=no
kicad_files/tigard.kicad_pcb            crlf=MATCH  lf=no
kicad_files/watchy.kicad_pcb            crlf=MATCH  lf=no
```

The replacement pins `poses_sha256` instead, which is computed from *parsed* footprint poses and is therefore identical on either checkout, and it never compares `routed_board_sha` — CLAUDE.md is explicit that outputs carry per-run UUIDs and must not be hashed to judge determinism. Thank you for chasing the hypothesis rather than reporting three failures; that saved the fix from being platform-specific in the other direction.

**8.8 h, not ~5 h.** Your figure is right and mine was wrong — I had summed route seconds from a partial view. `sum(provenance.total_seconds)` over the 87 rows is 8.77 h, median 288 s, max 2012 s on `tigard:perturb-wrong_side`. It is in the doc now, with the ~2.5 h wall-clock at `-j 4` beside it.

**On the artifact: dropped, and I took your suggested shape.** `tests/stress/predictor_rows.json` and `tests/test_703_predictor_rows.py` are removed. `tests/test_703_predictor_regen.py` regenerates four cheap declared variants — `esp_prog:authored`, `esp_prog:perturb-scatter-d1`, `esp_prog:portfolio-1`, `splitflap_driver:authored` — and diffs predictors and routed truth against literals. Measured at 51 s for all four:

```
ok  esp_prog:authored regenerates identically (5.4s vs 8.2s recorded)
ok  esp_prog:perturb-scatter-d1 regenerates identically (8.2s vs 11.2s recorded)
ok  esp_prog:portfolio-1 regenerates identically (15.2s vs 23.8s recorded)
ok  splitflap_driver:authored regenerates identically (22.4s vs 32.6s recorded)
```

Two boards deliberately, so a change that only moves one board's numbers cannot pass, and their two `argv_sha` values differ — the frozen-argv contract visible in the data rather than asserted. `esp_prog:portfolio-1` is in the subset because it is the study's headline in miniature: 23 crossings against the authored board's 53, the best on the slate, routing to blocking 3 where the authored board routes to 0.

**The cost, stated in the doc rather than left implicit.** I widened it past what you asked, because it is not only the ρ table: *no* number in `docs/placement-predictors.md` has an automated change detector any more — not the per-board tables, not the shuffle-control rates. It is the recorded finding, and re-deriving it costs 8.8 h. What is guarded now is the rig, not the artifact.

Two other things from your review worth confirming since they bear on trust in the rest:

- Your AST comparison matches what I intended — the engine files carry docstring changes only. `placement_driver.py`'s two changed string constants are both prose annotations from the r-value sweep.
- Your correction about `$?` after a `$(basename ...)` substitution is noted and it changes nothing about your conclusions; `test_703_predictor_rows.py` did fail on a non-Windows checkout, for exactly the CRLF reason, and it no longer exists.

Full suite on the branch: **436 passed, 1 failed, 1 timed out**. The failure is `test_763_kicad_locate.py`, which fails identically on `upstream/main` (a Windows path-separator assertion) and which this branch does not touch. The timeout was `test_obstacle_map_balance.py` at its own 1200 s budget while the study was still running; re-run alone on an idle machine it exits 0, per the runner's own instruction not to record a timeout as a failure without that.

PR #785, opened 2026-08-29
#703: the complete table -- 6 boards, 120 placements

### LINKED ISSUE #703: Placement: no pre-route predictor has been correlated with routed blocking; the quoted r-values are distance-to-truth

edgehero:
Closes #703. Follow-up to #784, which landed the apparatus and the first 4 boards.

**The declared table is 6 boards × K=20 = 120 placements. All 120 are now
routed; 119 produced a routed truth.** #784 shipped 4 boards and disclosed two
gaps — two unrouted boards, and `perturb-pile` timing out on 3 of 4 boards.
Both are closed here.

## The finding did not move as the sample grew

That is the point of having run the whole table rather than stopping at four:

| | 4 boards (#784) | 5 boards | 6 boards |
|---|---|---|---|
| legality family | 4/0 +0.724 | 5/0 +0.830 | **6/0 +0.785 passes** |
| `courtyard_blocking_pairs` | 4/0 +0.652 | 5/0 +0.826 | **6/0 +0.684 passes** |
| `pad_copper` | 4/0 +0.615 | 5/0 +0.568 | **6/0 +0.524 passes** |
| `crossings` | 3/1 +0.396 | 4/1 +0.394 | 5/1 +0.515 fails |
| `hpwl` | 2/2 +0.158 | 3/2 +0.040 | 4/2 +0.059 fails |

Ten predictors pass at a two-sided p of **0.031**, the floor for a 6-board sign
test.

- **`pad_copper` — the one pre-route number that already refuses anything
  (`loop_driver`'s L2 gate) — is positive on 6 of 6.** That gate was right.
- **`hpwl`, which the drivers gate on, is NEGATIVE on two of six** (−0.525
  esp_prog, −0.185 tigard; median +0.059, p = 0.69). Its distance-to-truth
  justification is untouched and **this PR does not remove the gate** — but
  nobody should call it a routability gate.
- **`crossings` has the highest median of any failing predictor** (+0.515) and
  is negative on esp_prog. Not useless; inconsistent, and the rule is about
  consistency.

## The sixth board paid for itself in the shuffle control

`--shuffle-control 200` permutes truth within each board and measures the
acceptance rule's own false-positive rate:

| N | range | median |
|---|---|---|
| 4 | 9.0–18.0% | 11.5% |
| 5 | 4.5–8.0% | 6.5% |
| 6 | **1.5–5.5%** | **2.5%** |

Each added board roughly halved it. And `cross_side_stacks`, defined on only
three boards, still shuffles at **30%** — the clearest available statement of
why the board count is the whole game, and why `MIN_SIGN_BOARDS` exists.

## Every exclusion #784 disclosed is gone

#784 reported `perturb-pile` timing out on 3 of 4 boards at a 2400 s budget.
Measured: those were **near-misses, not impossibilities** — splitflap needed
2798 s. All six boards now carry a pile row (blocking 466 / 877 / 5424 / 8521 /
13078 and one still-missing).

Raising a *timeout* is not a protocol change for rows that already finished: a
route that completed in 300 s completes identically at any larger budget, so it
strictly adds samples. Only the killed rows were re-run.

One variant of 120 remains missing: **kit-dev-coldfire's `perturb-pile`**, which
did not finish in a four-hour budget. It is the hardest route in the study by
construction — 160 parts collapsed onto one coordinate on four layers — and the
pile route time across the five boards that did finish scales 183 s → 9877 s
with board size. Its effective K is 19, far above the floor, and the doc says so.

## kit-dev-coldfire, for the record

It is the most expensive board in the set by a wide margin: its **authored**
placement routes in **5013 s** against tigard's 603 s, and its twenty variants
took about fourteen hours at 4-way. A first attempt at the 2400 s budget
produced four rows with `blocking = None` including the authored board, which is
what prompted measuring its real cost rather than assuming the board was
unmeasurable. It is measurable, just slow, and its data is good: `blocking`
spans 1 … 671.

## One bug fixed

`--from-rows` accepted only a `{rows: [...]}` document, while the study writes
`<out>/rows.jsonl` one row per line — and both the usage text and the doc tell
the reader to feed exactly that file back in. Following the documented command
died with `JSONDecodeError: Extra data: line 2 column 1`. A tool refusing the
file it had just written. It reads either form now. This surfaced only because
withdrawing the committed rows file in #784 made the JSONL the sole path anyone
would use.

## What a reviewer can check

Unchanged from #784 and all still cheap:

| cost | what it proves |
|---|---|
| 0 s | `tests/test_703_rank_stats.py` — 63 kernel assertions + 82 external checks |
| ~1 s | `tests/test_703_predictor_claims.py` — every r-value names its dependent variable |
| ~2 s | `tests/test_703_predictor_harvest.py` — 17 recorded-run shapes, synthesized |
| ~51 s | `tests/test_703_predictor_regen.py` — four declared variants regenerated and diffed |
| ~60 s | `tests/mutate_703.py` — 29 rows, 27 killed, 2 expected survivors, 0 broken |
| ~24 h | the 120 routes; `--from-rows` is free thereafter |

No engine file changes in this PR — it is the doc's measured table plus the
`--from-rows` fix.

## Suite

`tests/run_all.py`: **436 passed, 1 failed, 1 timed out** (2873 s). The failure is
`test_763_kicad_locate.py`, which fails identically on `upstream/main` — a
Windows path-separator assertion — and is untouched here. The timeout is
`test_obstacle_map_balance.py` at its own 1200 s budget; run alone it exits 0,
and `run_all` executes four tests in parallel, so it is contention rather than a
defect. Both were present before this branch.

## Two issues filed from this work

Profiling the study's own routes turned up something worth its own tracking, and
it contradicts the obvious assumption:

- **#786** — the Rust A\* main loop is **2.5% of wall time** across the 119
  routes (1.02 h of 40.7 h; 1.2% on kit-dev-coldfire). Post-route cleanup is
  **71%** of a route, bottoming out in 2.3M `_pt_seg` and 1.7M `union` calls.
- **#787** — `bga_fanout` carries a **second, pure-Python A\*** and 2.05M `abs` /
  297k `point_to_segment_dist` calls, in the same three predicates.

Both recommend numpy before Rust and argue GPU is the wrong tool for most of it,
with the numbers for why.

https://claude.ai/code/session_01TRf6KQc2F9Jwtv5DMFZaS9

### DIFF
--- a/docs/placement-predictors.md
+++ b/docs/placement-predictors.md
@@ -1,11 +1,10 @@
 # What predicts routed `blocking` — the first measurement (#703)
 
-**Status: MEASURED, on 5 boards of a declared 6, with NO excluded variants.
-The sixth (`kit-dev-coldfire-xilinx_5213`) is routing as of this revision.**
-No predictor here is "validated" in a strong sense; the acceptance rule's own
-false-positive rate at N=5 is **4.5%-8.0%** (median 6.5%) over the 21 predictors
-defined on all five boards, measured, and that number belongs beside every
-PASSES below.
+**Status: COMPLETE. The declared table is 6 boards x K=20 = 120 placements, and
+all 120 were routed.** No predictor here is "validated" in a strong sense; the
+acceptance rule's own false-positive rate at N=6 is **1.5%-5.5%** (median 2.5%)
+over the 21 predictors defined on all six boards, measured, and that number
+belongs beside every PASSES below.
 
 Every correlation number this repo quoted before #703 was measured against
 *distance-to-the-correct-placement* or against *the gap a human left*.
@@ -14,7 +13,7 @@ the corridor law's `r = +0.41 … +0.90` is against the human's gap. CLAUDE.md's
 *"What a placement run is FOR"* says the headline is routed `blocking`, and no
 predictor had ever been correlated with it.
 
-This is that measurement. Nothing in it is a proxy for a proxy: 100 placements
+This is that measurement. Nothing in it is a proxy for a proxy: 120 placements
 were generated, each routed once with an argv frozen before any variant existed,
 and each graded by `board_score`.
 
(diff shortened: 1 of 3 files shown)


drandyhaas:
Reviewed and merging. The measurement work holds up — the finding not moving as the sample grew from 4 to 6 boards is the part that matters, and the shuffle-control halving with each added board (11.5% → 6.5% → 2.5%) is a good argument for `MIN_SIGN_BOARDS` that didn't exist before.

**I pushed one commit to this branch** (`0d66263f`), because the `--from-rows` fix left a quieter version of the same defect behind.

A one-row `rows.jsonl` parses as a lone JSON object, so `d.get('rows') or []` reads it as a document declaring no rows and returns nothing. `main()` then prints `carries no study rows (only 0 harvest row(s))` — blaming the file's content for what the loader dropped — and returns **0**. The crash it replaced at least said "I cannot read this"; this says "there was nothing in it", which is false, and exits successfully. An interrupted run leaves exactly a one-row file.

Measured across every shape before the fix:

```
jsonl 3 rows         -> 3 rows
jsonl 1 row          -> 0 rows   <-- silent
document (indent2/compact/indent0) -> 2 rows
bare list            -> 2 rows
```

The rule that fixes it: only a *document* carries the `rows` key, so anything else that parses as an object is one row. That replaces the `startswith('{') and '\n{' not in ...` pre-check, which turned out to be redundant — the general `try/except` already handled every document form (compact, `indent=0`, `indent=2`), and it produced the same silent-empty answer on a one-row file. Removing it (and the `io` import) makes the loader shorter, not longer.

Added `tests/test_703_from_rows.py` — twelve checks, temp-dir fixtures, no `wk/` tree. Verified in **both** directions: against this branch as submitted it fails on exactly the one-row case; against `main` it dies with the original `JSONDecodeError: Extra data: line 2 column 1`. It also pins the two things a loosened loader could break — "no rows" must not become a phantom row (empty document, empty object, empty file all stay 0), and an unreadable line must still be refused **by name** with its `file:line`.

Worth noting for next time: the PR fixed a bug in a file-reading function and added no test for it. That's the gap that let the second defect through — the same fix, tested, would have caught it immediately.

Two smaller things, neither blocking:

- The doc headline says *"all 120 were routed"* where 119 produced a truth. The disclosure at lines 216–218 and 238 is precise and the table carries `K=19`, so this is phrasing rather than misrepresentation — but the headline is the line people quote.
- `tests/test_703_rank_stats`, `test_703_predictor_claims` and `test_703_predictor_harvest` all pass on the branch, as does the new file.

PR #790, opened 2026-08-29
#701/#704: a declared keep-out is enforced, and a decap limit can be authored

### LINKED ISSUE #701: Placement: intent keep-outs are graded and enforceable, and no seat-search call site passes them
The floorplan intent has a `keepouts` rule, the grader enforces it, the seeder has a matching seat predicate, and no seat-search call site passes the list.

`py_placer/placement/seeder.py:87,106-126` implements `pose_ok(..., forbid=())` with rect entries, `sides` and `allow` globs. It defaults to the empty tuple.

```
$ grep -c "_try_place(" py_placer/placement/seeder.py
13
$ grep -c "forbid=" py_placer/placement/seeder.py
0
```

Thirteen call sites, none of them passing `forbid`. So `pose_ok` is always evaluated with no keep-outs, while `rule_keepout` (`py_placer/placement/floorplan.py:795-827`) grades them afterwards. A declared keep-out is a region the seeder walks parts into and the grader then complains about.

It cannot be authored either. `emit_intent` hardcodes `'keepouts': []` at `floorplan.py:1527`, no CLI flag creates one, and no stage text asks for one. So the only way a keep-out exists is a hand-written intent file, which then produces a guaranteed failing grade and no change in behaviour.

The other direction does not round-trip either. The only producer of keep-outs is the plan op `place_keepout` (`plan_resolve.py:1580-1598`), and `compile_intent` (`plan_resolve.py:1812-1918`) emits `blocks`, `edge_connectors` and `must_lock` with no `keepouts` key. So a plan keep-out is honoured during the plan and forgotten afterwards, and an intent keep-out is graded forever and honoured never.

This matters because keep-outs are how mechanical reality reaches a floorplan: enclosure ribs, standoffs, a battery, a display window, an antenna clearance, a connector's cable shadow. All of them reduce to "a rectangle nothing may enter", which is the one thing the engine already implements end to end.

Suggested fix, mostly plumbing since the geometry exists:

1. Thread `intent.keepouts` from `seed_from_intent`, `repair_placement` and `reseat_scope` into every `_try_place` and `count_legal_poses` call, converting each entry once via `_keepout_entry`.
2. Honour the `allow` glob so a spec-derived keep-out can exempt the part that owns it, which is the same lesson as interior milled rings needing an owner exemption.
3. Have `compile_intent` emit a `keepouts` entry per `place_keepout` op, so the plan's mechanical facts survive into the graded artifact.

`_keepout_entry` (`seeder.py:70-83`) also does not read the `circle` form that `rule_keepout` already grades.

One caution: a keep-out that strands a part should surface as "no legal pose" naming the keep-out, not as a silent failure. `count_legal_poses` already reports pose counts, so the reporting channel exists.

### LINKED ISSUE #704: Placement: emit_intent writes keepouts and decaps empty, so three of nine floorplan rules can never fire
`emit_intent` is the only automatic producer of a floorplan intent, and it writes two of the rule inputs as constants.

`py_placer/placement/floorplan.py:1527-1529`:

```python
'keepouts': [],
'edge_connectors': conns,
'decaps': {},
```

`rule_decap_distance` returns immediately when `max_distance_mm` is None (`:918-920`), and `rule_keepout` has nothing to iterate. `must_lock` is emitted empty too, deliberately and with a comment explaining why. So three of the nine entries in `RULES` (`:993-1003`) cannot fire on an auto-emitted intent by construction.

Observed on a real closed run: the emitted and the final intent both carried `"decaps": {}` and `"keepouts": []`, `rules_run` was `[envelope, edge_connector, legality]`, and the verdict was `pass: true`. Three rules of nine ran, and the document reported a clean grade.

The module already has the principle this violates, at `floorplan.py:32-34`: "0 violations" and "0 rules ran" must not look the same to a machine. `rules_skipped` records the distinction honestly, but the top-level `pass` does not, and consumers key on `pass`.

This is the authoring half of the same gap as the enforcement half filed separately: keep-outs are graded, the seeder has a seat predicate for them, and nothing can author one.

Suggested approach for `decaps`, following the pattern `legality_budget` already uses at `:1512-1525`: under `--declare-classes`, populate `max_distance_mm` from the board's own measured tether distribution, rounded up rather than to nearest, and withhold it into `budget_withheld` when the emitting board already violates it. That is the self-bless guard the neighbouring key already has. Tag the entry `source: "auto-tethers"` so it reads as an observation rather than a spec, matching how the connector entries already carry `source: "auto-class"`.

`keepouts` is different: a keep-out is a mechanical fact that cannot be read off a board, so the emitter should keep writing `[]`. What it should not do is leave the reader unable to tell "no keep-outs declared" from "keep-outs not considered". Emitting the key with a stated reason in `rules_skipped` is enough.

edgehero:
Closes #701. Closes #704.

**On #701 I asked first and got no answer.** I offered on the issue to open this
rather than "push a change to the seat predicate unannounced", and the offer went
unanswered for a week. Here it is as a PR instead, because a diff is easier to
judge than a question — and it is **strictly inert** on every board that declares
no keep-out, which is every board in the corpus. Measured two ways: 0 invocations
of either new primitive across seed / `--repair` / `--reseat` on three boards
(including runs that did real work), and **7/7 byte-identical** `place_seed`
outputs base-vs-branch. Close it if you don't want it.

## Why one PR for two issues

#704's own body says it: *"This is the authoring half of the same gap as the
enforcement half filed separately."* They share `floorplan.py`, the
`budget_withheld` → abstention channel, and the same emit/seed/grade round trip.

---

## THE RULE WAS ENFORCED BY NOTHING

`rule_keepout` grades a declared keep-out. `grep -c "forbid\|keepout"
py_placer/placement/seeder.py` returned **0**. A keep-out was a region
`place_seed` walked parts into and `check_floorplan` complained about, forever.

The issue's suggested fix — thread an existing `forbid` parameter through 13 call
sites — does not apply; the author corrected that on the issue. The real count is
**12** `_try_place(` call sites (the 13th grep hit is the `def`) plus one outside
the module, and threading would touch ~19 places **and still miss the pose
census** — the exact drift `pose_ok`'s own docstring exists to prevent.

So the keep-outs ride on the state and `pose_ok` gains one conjunct. That covers
all 12 sites, the census, and `_evict_trade`'s acceptance test. Two more paths
needed their own work:

- **`edge_seat_ok`** bypasses `pose_ok` deliberately (an edge connector overhangs
  by design), and is used by `_seat_edge` *and* by stage 1, which "runs no
  legality gate at all by design". Editing only `_seat_edge` leaves stage 1 open.
- **The polish quench has no keep-out term**, so it can walk a part back in and
  make `place_seed` fail its own grade. The self-check already had this exact
  repair for `zone_containment`; it now covers `keepout` too.

**Grader and seat share one implementation**, and that is tested behaviourally
rather than by grep: patch `floorplan.keepout_hit` and require the seat predicate
to observe the patch. A copy-pasted geometry would not, and would still pass any
agreement check written against equivalent-but-separate code. The agreement sweep
is 715 (part, keep-out) combinations — 198 hit, 517 miss, **0 disagreement** —
with both bounds asserted, because a sweep with no hits proves nothing.

A keep-out that strands a part now reports `keepout_blocks` with the keep-out
**named and the poses it frees counted**, instead of `no_movable_neighbour`, whose
prose ("the outline, the zone or its own size refuses it, not a neighbour") is
false in that case and sends the reader to three things that are not the problem.

---

## THE SELF-BLESS GUARD #704 ASKS FOR IS UNREACHABLE

The issue asks for the `overlap_area` guard: withhold when the emitting board
already violates the number. `ceil(max(observed))` is ≥ every observed value **by
construction**, so that branch can never execute — worse than absent, because it
reads like a guard.

The reachable analogue is **censoring**: `decap_tethers` drops any cap past 5 mm,
so a survivors' max can bless a board whose caps have left decoupling range.

| board | tethers ≤5 mm | beyond | censored | worst beyond |
|---|---|---|---|---|
| tigard | 25 | 1 | 0.04 | 6.17 |
| glasgow_revC | 87 | 5 | 0.05 | 10.34 |
| splitflap_driver | 11 | 1 | 0.08 | **19.30** |
| watchy | 24 | 2 | 0.08 | 11.16 |
| ulx3s | 53 | 7 | 0.12 | 12.24 |
| kit-dev-coldfire | 41 | 12 | 0.23 | — |
| interf_u_unrouted | 1 | 4 | **0.80** | 19.82 |
| sonde_u | **0** | 5 | **1.00** | 15.58 |

Withheld when there are no tethers, fewer than 3, or >25% censored. **A rejected
alternative, recorded so it is not re-proposed:** withhold when `max > K * p75`.
Built and measured; it does **not** separate — healthy splitflap scores 2.46 and
mid-repair `tigard_placed` 2.22.

The statistic is `ceil(max)` because an emitted intent needs a **fixed point**.
The median is refuted by measurement: glasgow_revC's is **0.0000** (58 of 87 caps
sit inside their IC's bbox and clamp to zero), so a median limit flags 29 caps on
a healthy human-routed board at the first emission.

---

## THIS KEY CHANGES PLACEMENT, NOT ONLY GRADING

`place_seed` reads `decaps.max_distance_mm` and pulls every 2-net-bearing-pad
`C*` out of zone packing into its per-supply-pin stage:

| board | seeder scope | graded tethers | in scope, never graded |
|---|---|---|---|
| ulx3s | 70 | 53 | **17** |
| glasgow_revC | 92 | 87 | 5 |

The two "is this a decap" predicates genuinely differ (grouper: *exactly two
distinct net ids*; seeder: *exactly two net-bearing pads*), as do the metrics.
So the derivation is **opt-in, default off, on its own `--declare-decaps`** rather
than folded into `--declare-classes` — three measured reasons for the separate
flag, including that `tests/test_run23_courtyard_channel.py` asserts
`budget_abstained == ['overlap_area']` *exactly* and would have broken. It is
green here. The coupling is disclosed in the census, the stdout block, the help
text and the docs, and pinned by a test that fails **with the numbers** if the
predicates are silently reconciled.

---

## What this PR does NOT claim

- **The run reported in #704 is not fixed.** On a board of that shape the
  withholding fires, so it still reports 3 rules run, still `pass: true`, still
  exit 0 — plus one line in `budget_abstained`.
- **`rules_run` does not go up on a withheld board.** This converts a silent skip
  into a loud abstention; it does not convert it into coverage.
- **`ceil(max)` is maximal self-bless**, and the censoring guard is only a proxy.
  A cap dragged from 1.2 to 4.9 mm stays inside the radius and sets the limit.
  Known, unfixed.
- **The 5 mm horizon is disclosed, not closed.** splitflap_driver emits 3.4618
  while a rail-sharing cap sits 19.30 mm out, invisible to emitter and rule alike.
- **#701's suggestion 3** (`compile_intent` emitting keep-outs per `place_keepout`
  op) is not implementable on `main` — `plan_resolve.py` lives only in an unmerged
  tree, as the issue author's own correction notes.

## Tests

Three new files, 105 assertions. Every headline is backed by a mutation applied to
the shipped code and reverted — **20 mutations, all killed.** Highlights:

- `pose_ok` conjunct removed → 14 seating failures
- `edge_seat_ok` conjunct removed → the edge arm alone goes red
- `_ceil4` → `round` → the decap round trip fails on 3 boards
- derive without the flag → the default arm **and** the A/B run
- `keepouts=intent.keepouts` → `keepouts=None` → the static AST gate

**Six defects this work caught in itself**, all recorded in the commits — including
deriving the limit from a *display-rounded* max (reintroducing the exact bug
`_ceil4` exists to prevent, on 3 of 9 boards), and **four of my own tests that
passed in both directions**: the `>`/`>=` boundary, the through-hole rect, a
grader-independence check reading a constructor kwarg instead of the effective
state, and a censoring guard asserted by *nothing* until I found the board that
isolates it.

An adversarial verifier then re-derived the whole mutation table, found **three
BLOCKING defects** (all fixed here), and **invented a 14th mutation that survived**
— the four-character `keepouts=None`. Its other two: two overlapping keep-outs
reported `{}` and fell back to the misleading verdict; and stage 1 traded a
`keepout` grade error for an `edge_connector` one by parking a connector inland
when 26 clear edge seats existed. Both fixed, both pinned.

## Verification

- **Full suite: 442 passed, 2 failed, 3 timed out.** Both failures
  (`test_763_kicad_locate`, `test_782_nondefault_netclass_clamp`) reproduce
  identically on `upstream/main` — verified by running each in a worktree at
  `2c84c012`. All three timeouts **pass when run alone**, which is what the
  runner's own note asks for.
- **`test_placement_ab.py` PASS**, 3/3 rows match the committed baseline — no
  re-record owed. The emitted-intent hashes are byte-identical to main on all
  three A/B boards, and a grep for `keepout` over the whole log finds nothing.
- **No `tests/gui_parity/` gate is owed** — the placement seeder is unreachable
  from the plugin (the Placement tab drives Claude Code headless; the only direct
  `placement.*` imports are `placement.labels` and `groups.parse_sources`). Both
  gates CONTRIBUTING.md names were run anyway and are green.
  `.gui-parity-checked` is **not** updated; that is a separate audit.
- One caveat on portability: `test_run5_emit_guard.py` and `test_part_class.py`
  self-skip without the gitignored `wk/` corpus, so they are not portable evidence.

Follow-ups worth filing separately, all measured here: `place_seed --reseat`
seats correctly and then `prune_assignment` reverts it (a gate with no keep-out
term); `place_portfolio` generates violating candidates and gates them after;
`allow` globs are not validated against the board; and the two decap predicates
disagree.

### DIFF
--- a/.claude/skills/plan-pcb-placement/SKILL.md
+++ b/.claude/skills/plan-pcb-placement/SKILL.md
@@ -81,6 +81,7 @@ this part* and *everything near it is locked*.
 
 | verdict | what it means | what to do |
 | --- | --- | --- |
+| `keepout_blocks` | a **declared keep-out** refuses it — measured, by recounting the poses with that keep-out lifted (#701) | move the keep-out, or add the part to its `allow` list if it owns it |
 | `no_movable_neighbour` | nothing seated is near it | eviction cannot help — the pocket does not exist; re-check the outline or the intent |
 | `immovable_given_frozen` | everything near it is frozen (it names each neighbour **and the decision that froze it**) | unlock one of them, or accept it |
 | `blocker_available` | a useful blocker exists, the rung is disarmed | raise `--evict-depth` |
--- a/docs/floorplan-intent.md
+++ b/docs/floorplan-intent.md
@@ -20,8 +20,9 @@ python3 py_tools/check_floorplan.py board.kicad_pcb --intent floorplan.json --js
 
 The intent is also a GENERATOR input, not only a grader's: `place_seed.py`
 turns a declared intent into an initial placement for an unplaced board
-(zones, edge bands, locks and decap rules become placement constraints, and
-the emitted seed is graded against the same intent it was built from), and
+(zones, edge bands, keep-outs, locks and decap rules become placement
+constraints, and the emitted seed is graded against the same intent it was
+built from), and
 `place_portfolio.py` uses the intent as the hard gate plus the `health`
 signals when ranking K perturbed placement candidates. See
 `placement/README.md` for both.
@@ -221,7 +222,7 @@ for it, and the reason is printed:
 | `zone_containment` | a member's courtyard leaves its block's zone | `GradedPart.rect` |
 | `zone_side` | a member is on the other face | `legality.footprint_side` |
 | `zone_exclusive` | a non-member intrudes on a reserved zone | `rect_overlap_area` |
-| `keepout` | any part enters a keep-out, unless in `allow` | courtyard **and** through-hole rect |
+| `keepout` | any part enters a keep-out, unless in `allow` | courtyard **and** through-hole rect. **Enforced, not only graded, since [#701](https://github.com/drandyhaas/KiCadRoutingTools/issues/701)** — the seat search refuses such a pose through the same `keepout_hit` this rule calls |
 | `edge_connector` | overhang outside `[min,max]`, or the wrong edge; a `connector_affinity` entry seated more than 3 mm from every edge fires at **warn** whatever the configured severity | `BoardOutlineGate.rect_outside_amount`, `edge_clearance` |
 | `decap_distance` | a decoupling cap is too far from its own IC | `groups.decap_tethers` |
 | `must_lock` | a declared-critical part is not locked in the file | `parser.extract_locked_refs` |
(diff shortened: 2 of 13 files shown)


drandyhaas:
Reviewed and merging. This is the better-evidenced of the two, and the inertness claim is the one that mattered — so I checked it independently rather than taking the PR's word for it.

**Verified inert on two boards**, both doing real work (the first attempt at this was vacuous: `place_seed` refuses an already-placed board with exit 3 and writes no output, so comparing the outputs compared two files that didn't exist — it needs `--force`):

| board | parts placed | poses base vs branch | summary |
|---|---|---|---|
| splitflap_driver | 65 (128 pose lines changed vs input) | **byte-identical** | identical (crossings 691, hpwl 3269.926) |
| glasgow_revC | 264 | **byte-identical** (3776 lines) | exit 0 both arms |

Also checked the thing that isn't obviously inert: `pose_ok` now calls `part.rects()` instead of `part.rect()` **before** the keep-out guard, so every board pays for the extra through-hole rect on the hottest predicate in the seeder. Measured on two runs per arm: 5.6 s vs 5.6 s. No cost.

And the claim that the quench objective never reads keep-outs holds — `quench.py` only stores and resolves them; the consumers are `seeder.pose_ok` / `edge_seat_ok` and the grader.

All four gates pass here: `test_701_keepout_predicate`, `test_701_keepout_seating` (45), `test_704_decap_emission`, `test_549_floorplan_schema`.

**One claim in the description needs correcting** — the emitted intent is *not* byte-identical to main:

```
ONLY-IN-PR790 /context/decap_census
ONLY-IN-PR790 /context/keepouts_note
```

Both land on **every** emitted intent, flag or no flag. That's deliberate and commented in the code, and it's the right call — but it means anyone diffing or hashing emitted intents will see a change, and "byte-identical" is the kind of phrasing that later gets cited as a guarantee it can't carry. The *constraint* set genuinely is unchanged, which is what matters: `decaps = {}` on both arms, and `max_distance_mm` appears only with `--declare-decaps` (`{'max_distance_mm': 3.4618}` with the flag).

On #701 specifically: you asked on the issue first and got no answer for a week. That was the right thing to do and the silence was mine, not a verdict — opening it as a diff was the right call too. The design choice that sold it is putting the keep-outs on the state and giving `pose_ok` one conjunct, rather than threading `forbid` through ~19 call sites and still missing the pose census.

The follow-ups you name at the end — `--reseat` seated-then-reverted by `prune_assignment`, `place_portfolio` gating after generating, unvalidated `allow` globs, and the two disagreeing decap predicates — are worth separate issues; file them and I'll look.

Issue #694, opened 2026-08-22
Placement: the corridor A/B row's recorded evidence has inverted, and the gate still reports PASS


edgehero:
`tests/test_placement_ab.py` rejects the corridor term on a recorded measurement that is no longer true.

The `why` string at `tests/test_placement_ab.py:85-92` reads:

> MEASURED: the term moves the placement (crossings 2417 -> 2390) but does NOT improve what it claims -- the re-derived bus_foreign_crossings went 62 -> 63 and hpwl 7512 -> 7634.

Re-running that row today with `python3 tests/test_placement_ab.py --row corridor-ulx3s`:

```
crossings                      2477 ->      2407
hpwl                        7437.99 ->   7416.23
health_bus_foreign_crossings     62 ->        55
intent errors 14 -> 17
corridor-ulx3s   REGRESS (as measured)
pinned:   1/1 rows match their measured verdict
PASS
```

Every number in the recorded justification is now wrong, and the sign of the signal the term exists to improve has reversed. `bus_foreign_crossings` improves by 11 percent, and both guards improve as well.

The REGRESS verdict now rests entirely on `intent errors 14 -> 17`, which the `why` never mentions. Grading the two written boards directly: OFF is `errors=14 {block_unresolved: 10, zone_containment: 4}`, ON is `errors=17 {block_unresolved: 10, zone_containment: 7}`. So the verdict is 3 extra `zone_containment` violations, while 10 of the 14 baseline errors are `block_unresolved`, a broken-intent condition unrelated to placement quality.

The gate still reports PASS because `_verdict` (`tests/test_placement_ab.py:274-298`) compares only the sign of the outcome, and that sign happened to be preserved by an unrelated criterion.

The same stale numbers are published in `docs/placement-optimization.md:210-216` and in `py_placer/place_portfolio.py:98-111`'s `--help`.

Two structural causes:

1. Row evidence lives in prose `why` strings. Nothing is machine-checked except a three-valued `expect`, so a row can invert silently.
2. The ">= 3 boards, paired and directional" rule stated in CLAUDE.md is documentation only. The code is `ok = ok and not t_reg and len(t_imp) >= max(1, judged - 1)` at `:398`, where `judged` counts trial rows rather than distinct boards, so one improving row on one board passes.

[...]

Issue #625, opened 2026-08-12
core64_logic: a 4-minute board no longer terminates -- route.py burns the full 3h cap in an unbounded rescue loop


drandyhaas:
## Summary

`core64_logic` (set19) replays in **4.0 minutes** historically and now **does not
terminate**: it is killed by the 3-hour per-command cap with the chain
incomplete. It failed this way in **every arm of every cloud run** on
2026-08-12 — a dozen arms, three hours each — and in a 7-arm full-corpus sweep it
was the one board whose chain broke in all seven.

Because one non-terminating board holds every arm at N-1, it also blocks
corpus-scale A/B work until excluded.

## Evidence

Recorded cost table entry (replay-measured, not an LLM estimate):

    core64_logic: {'seconds': 240.018, 'peak_mb': 426.0, 'src': 'replay'}

Observed (cloud replay at `bd0df44`, harvested row + persisted replay log):

    chain_complete: False      final: None
    total_seconds: 10804.4     peak 2061 MB RSS
    [9] route.py  10800.38s  TIMED OUT after 10800s (killed)
    [2] route_planes.py  4.02s
    Replayed 2 of 9 command(s) (pruned to the dependency chain)

So one `route.py` invocation consumes the entire budget: 4 min -> >3 h, and
peak RSS grows 426 MB -> 2061 MB.

## What it is doing

Not a deadlock — an unbounded search/rescue loop. From the replay log:

* **1477** separate `No route found after N iterations` searches
* single searches reaching **4,967,820** iterations (forward+backward)
* 14 `net rescue` lines, including a fab-floor escalation:

      WARNING: net rescue net_17: escalated standard->advanced fab floor
      (0.25/0.15 via etc., more costly to fab)

      Probe: forward=5000, backward=5000 iters, trying forward with full iterations...
      No route found after 115809 iterations (forward), trying backward...
      No route found after 1091413 iterations (both directions)
      Probe: forward=5000, backward=5000 iters, trying forward with full iterations...

i.e. the rescue rungs keep re-probing nets that cannot route, each attempt
allowed to run to a very large iteration budget, with no global cap on the
aggregate.

## Repro

    python3 tests/stress/redo_stress_test.py \
        ~/Documents/kicad_stress_test/runs_set19/core64_logic/redo_commands.sh

The hanging step is the manifest's `route.py`:

    route.py step1_pour.kicad_pcb step2_route.kicad_pcb --nets '*' \
[...]

Issue #529, opened 2026-07-30
max_iterations could dynamically expand while the search is still making progress toward the target


drandyhaas:
Today `max_iterations` is a hard guillotine: the Rust A* stops dead at the cap (`router.rs:488-504`, `GridSearch::is_exhausted` / the per-iteration check) regardless of whether the search was actively closing in on the target or hopelessly flooding. A long net on a big board can be 90% of the way there and get cut off, then graded as a routing failure — while a genuinely walled-in net wastes its whole budget flooding a pocket. One static number can't serve both.

**Proposal:** let the cap expand dynamically, up to **50× the configured value**, as long as the search is making *actual progress toward the target*:

- Track the minimum remaining-distance estimate to the target set (the A* heuristic of the best node popped so far) and the iteration at which it last improved.
- When the base cap is hit, continue in increments (e.g. +1× base each grant) **iff** the min-distance improved within a recent window (say the last 25–50% of the base budget). Stop granting the moment progress stalls, or at the 50× ceiling.
- Report the extension in the result/log ("extended to 3.2x: closest approach 1.8mm -> 0.4mm") so replays and A/Bs can see it, and expose the closest-approach metric on FAILED searches too — it would also sharpen the existing `blocked_cells` frontier diagnostics that rip-up blame uses.

**Where:** the clean implementation is Rust-side (the search loop already owns the heuristic; a min-tracker + grant check is a few lines) — noting that means a router version bump + rebuild + binary redistribution per the usual policy. A Python-side fallback (on iteration-exhausted failure with improving closest-approach, retry with a doubled cap) avoids touching Rust but re-pays the whole flood on every retry, so it is strictly worse for the boards this most helps.

**Interactions to keep in mind:**
- Determinism: extension decisions depend only on search-internal state, so replays stay deterministic.
- Wall-clock: 50× worst case only occurs while monotone progress holds; a stalled search still dies at 1× + one stalled window. The per-command timeouts in the stress harness stay the backstop.
- The chains mostly pass `--max-iterations 1000000` precisely to paper over the static cap; with dynamic extension the default (`GridRouteConfig.max_iterations = 200000`) may be able to come back down, which is a memory/latency win on the boards that never need the headroom.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013UMMzuYzs9AMptGGX5EZdX

drandyhaas:
[...]

PR #827, opened 2026-09-01
#713: the wall clock stops deciding which placement wins, and an instrument that could not answer stops reading as a clean verdict

### LINKED ISSUE #713: Placement: wall-clock decision sites that survived #621 still change placement and plane outputs
Following #621, which deleted the wall-clock budget facility repo-wide on the principle that nothing should depend on timing for the results, a handful of wall-clock sites remain that still change an output rather than only reporting one.

I went looking for the survivors. Classifying every `time.time` / `monotonic` / `perf_counter` / `timeout` site as (a) reporting only, (b) decision-changing, or (c) infrastructure, the (a) and (c) sets are in good shape: roughly 45 sites across the routing loops accumulate `total_time` or print, and none gates a branch; the test-runner timeout and the harness wrappers are legitimate. These are the (b) cases.

**1. `--plane-score-budget` decides which placement candidate wins.** `py_placer/place_portfolio.py:174` defaults it to 300 s; `:466-471` sets `incomplete = True` and breaks on overrun, and `:497-500` then strips `plane_islands` and `plane_neck` from `rank_key` for **all** candidates. The same board on a slower machine ranks on a different key set and can promote a different winner. The all-or-nothing strip is correct reasoning applied to the wrong variable: the work is already bounded by `--candidates N`, so scoring is N+1 refills regardless. The legitimate reason to strip is `plane_fragility_score` returning None, which is a capability fact rather than a timing one.

**2. A timed-out probe route is recorded as a missing verdict.** `place_portfolio.py:268-270` and `compare_seeds.py:168-171` wrap the probe in `subprocess.run(timeout=...)` and return `{'failures': None, 'note': 'timeout after Ns'}`. `None` is the absence of a verdict, produced by machine speed. `route.py` already exposes the deterministic knob for this, `--max-iterations`, and exhausting it yields a real reproducible result with a real summary. Both tools exist to compare candidates, and a comparison where one arm's verdict was erased by a clock is not one.

**3. One slow DRC disables the KiCad oracle for the rest of the run.** `py_router/kicad_oracle.py:245-252` catches `TimeoutExpired`, returns None, and also memoises the board's fill-cost signature into a module-global, so every later plane-repair step short-circuits to `available: False` at `:1424`. The run summary then reads `oracle_check: skipped` with no recorded cause. Since the oracle is the authoritative zone-aware check, a run where it silently went away has an unbacked clean verdict. Suggested split: keep the subprocess timeout as a hang detector, make the memo opt-in and default off so the result is a property of the board rather than of run history, and record the event with a reason so the enclosing summary is not read as fully checked.

**4. `EXACT_FILL_TIMEOUT` makes plane geometry depend on CPU speed, and reports it as "unavailable".** `py_router/kicad_exact_fill.py:476` runs the pcbnew refill under a 300 s timeout, and `:484`'s bare `except Exception` swallows `TimeoutExpired` into the same `return None` used for "pcbnew is not installed". `plane_fragility.py:424-427` then takes the outline-only fallback, so the copper the router builds against differs. This was the root cause of a lost campaign on a 217-part 4-layer board: all 9 zones ended with `filled_polys = 0`, the outline fallback re-carved copper on every refresh, and the plane repair never terminated at 2400, 1500 or 1200 seconds. The two causes need different responses and are currently indistinguishable, so at minimum `TimeoutExpired` should be caught separately from `Exception`. A deterministic pre-flight predicate on board size would make the path taken a function of the board alone.

**5. A truncation must not be readable as success.** Not wall-clock itself, but it is the contract any counter budget would emit through, so it is worth settling first. `placement/floorplan.py` computes `budget_abstained` for legality channels that were never graded and prints "NOT DERIVABLE -- not graded, not passed" at `:1659`, while `passed` is `not self.errors` (`:1113-1115`) and an abstention produces no Violation. So the machine channel `pass` stays true and the human line still reads PASS. `route_summary.py:82-92` already has the pattern that works, a sticky `complete: false` that `place_route_loop.py:178-186` refuses on. The placement side should reuse it rather than invent a second channel.

Worth noting for scope: none of this touches the GUI. `grep -rn "deadline" kicad_routing_plugin/*.py` returns nothing, so every one of these is CLI and engine side.

On the other half of the principle, that things should be fast enough not to need budgets, the honest answer is board-dependent and measurable. A 20-part board's whole recorded chain is 253 s across 153 invocations with a 24 s worst case, and no budget was ever wanted. A 92-part board's is 3962 s, of which routing is 2600 s, with one reachability call taking 306 s for a single reference. Those are performance defects rather than budget problems. I have placement-side profiling that quantifies the largest one and can file it separately if useful.

edgehero:
Closes #713

Refs #621 (deleted the wall-clock budget facility repo-wide; this finishes the job), #420, #625

#621 deleted `--deadline` from ten mains on one rule: *nothing may depend on timing for the results*. Five sites survived it, in four files, for two years. They survived because each one **looks** like reporting — its wrong answer is silent, and the thing it silently decides is which board ships.

This closes all five, and adds the change detector that would have caught them.

## What the issue got wrong

I checked every claim before building on it. Six were stale or backwards, and each changes the work, so the corrections are part of the PR rather than a footnote.

| # | The issue says | Actually |
|---|---|---|
| all | line numbers throughout | **every one is stale**; the real ones are in the commits |
| 2 | "`route.py` already exposes the deterministic knob for this, `--max-iterations`" | **materially wrong.** It is a per-A\*-**search** cap (`docs/configuration.md`: "per route"), extended at runtime by #529 to a 10M ceiling. Exhausting it fails *one net* while the run completes normally and exits 0. Lowering it changes the routing **result**, not the duration. |
| 3 | one slow DRC "disables the KiCad oracle for the rest of the run" via a memo | **inverted.** The memo's only *read* sat behind `env_knobs.LEGACY_ORACLE` — `KICAD_LEGACY_ORACLE`, default off, set by **nothing** in the repo. It was written and never consulted. The live defect was the opposite: the 240 s was **re-paid every round**. |
| 3 | "the run summary then reads `oracle_check: skipped`" | a timeout maps to `'failed'`; an ordinary run reports `'disabled'` (the block is opt-in). And the `oracle_reconnect` leg that returns `available` had **no summary key at all** — that is the real hole. |
| 4 | "at minimum `TimeoutExpired` should be caught separately" | **already landed** in `f019ed91`. What survived is quieter: both arms still returned a bare `None`, and 4 of 6 call sites never pass `verbose`, so a 300 s timeout produced *no output at all*. |
| scope | "none of this touches the GUI" (`grep -rn "deadline" kicad_routing_plugin/*.py`) | **false for items 3 and 4** — the grep is for the wrong word. `gui_utils.py` calls `oracle_reconnect`; `kicad_parser._live_fill` (the GUI's `exact_fill_provider`) calls `refill_islands`. pcbnew is a **long-lived process**, so module-globals there persist across every board in a session. |

## The one contract

> **An output must be a function of the board and the flags, never of machine speed. And where an instrument genuinely could not answer, that must be a named, ranked, non-passing verdict — never a bare `None`, never a row dropped from a ranking, never a PASS.**

The first half is #621's rule. The second is the half it never wrote down, and it is *why* the survivors survived.

## Item 1 — a clock picked the winning placement

`--plane-score-budget` (300 s over all candidates) broke the scoring loop on overrun and stripped `plane_islands` / `plane_neck` from **every** candidate's rank key. The blast radius is larger than a printed order: the static ranking feeds `select_diverse` → the kept set → `probe_cands` → `_affected_nets` → the routed ranking → `select_best`, so the clock also decided **which boards got probe-routed**.

Deleted, hard, no shim — the repo's four precedents (`--clamp-netclasses`, `--no-clamp-netclasses`, `--protect-nets`, `--deadline`) are all hard removals. The work was never unbounded: `contenders` ≤ `--candidates`, one refill each.

**The strip itself stays, and has to.** `rank_key` reads the terms as `m.get('plane_islands', 0)`, and **0 is the optimum** — so there is no neutral sentinel and partial coverage cannot be ranked at all. That is asserted, not described.

What changes is which causes may trigger it, split on `PlaneScoreStatus.uniform`:

- **uniform** (no bounds, no named net, no pcbnew) — identical for every candidate, so nothing was scored, the strip is fair, and the run continues ranking as an un-flagged one would. It says so.
- **non-uniform** (a timeout, a failed refill, an empty pour) — can strike candidate 7 and spare candidate 1. `--plane-score` is a **demand**: the run refuses with exit 3, naming the cause, the candidate and the coverage, and emits `complete: false`.

Also fixed: `contenders` filtered on `c.board`, not `c.viable`, so a candidate that failed its hard gates paid a full pcbnew refill for a rank slot it can never occupy — while the flag's own help text has always said "every viable candidate". Measured on splitflap_driver at `--candidates 3`: 2 refills, not 3.

## Item 2 — a clock erased a probe's verdict

Two probe sites wrapped `route.py` in `subprocess.run(timeout=...)` — 900 s and 1800 s, two clocks, two defaults, neither covered by a test nor named in any doc — and recorded the expiry as `failures: None`. **Both tools then filter `failures is not None` out of their rankings**, so a clock did not rank a candidate worse; it stopped it being a contender.

Both gone. The bound is scope, which both already pass. Precedent for calling `route.py` unbounded is in the repo already (`place_route_loop._run_route_cmd`, `test_placement_probe.py`).

**One probe helper.** There were two, whose no-verdict rows dropped *different* subsets of the success row's keys — which is why `tests/test_compare_seeds.py:96` was a latent `KeyError` rather than the assertion it looks like. `converge.probe_route` returns one shape whatever happens, and `failures: None` stops being an overloaded channel: `crashed` / `no_summary` / `screened` are distinct statuses.

Two real bugs found alongside and fixed:

- **`compare_seeds` returned 0 with `best_seed: null`** when every probe failed — a no-verdict row is a *truthy* dict, so `any(r['probe'])` was true — contradicting `docs/utilities.md`'s own contract. Now 4 in every no-winner path.
- **An absent verdict is now named** on stderr, on both the window and the full-board leg, saying the winner was decided without it.

## Item 3 — the oracle memo, corrected in the opposite direction

The module-global `_ORACLE_TIMED_OUT` is replaced by **call-scoped** state. The issue asks for "opt-in and default off", which would deliver the determinism half and leave the re-pay — and the re-pay is what actually killed the campaign the issue cites. Call-scoping delivers both: a timeout in round 1 stops rounds 2 and 3 re-paying, and because the set dies with the call, no verdict crosses a call boundary. That matters most in pcbnew, where one slow board would otherwise poison every board opened afterwards in the session.

`ORACLE_DRC_TIMEOUT` stays at 240 s — a hang detector on an external tool is the one legitimate use of a clock under #621's rule. What was wrong was the verdict it produced.

**A verdict that did not happen now says why.** `available: False` was a bare bool, and its consumers guessed: `repair_planes` blamed "kicad-cli not found" for *every* cause; `gui_utils` never inspected it at all, so an oracle that could not run was byte-identical to one that ran and found everything connected; `route.py`'s plane-finalize leg had no summary key, so it never reached a `JSON_SUMMARY` consumer; and a fourth site printed "0 link(s) welded, -1 remaining". All four fixed.

## Item 4 — five causes, one `None`

`refill_islands` had five ways to return the same bare `None`. Exactly one — the timeout — is a fact about the machine. `plane_fragility` filled the gap by re-deriving the reason itself, and reported a timeout as **"the KiCad refill failed"**: a property of the board, sending the reader to the wrong repair.

`refill_islands_ex` returns `(islands, RefillStatus)` over a closed six-reason vocabulary. `refill_islands` keeps its exact signature and return type, so no call site *needs* an edit; three were edited because they decide on the absence. A sixth outcome — an empty dict, which several callers test with truthiness rather than `is None` — gets its own clause, and it survives the memo.

**Deliberately not done:** the board-size pre-flight predicate the issue suggests. No measured threshold exists, and inventing one swaps a silent path-selector for a silent path-selector. Filed separately; the building blocks are named in the code.

## Item 5 — a truncation read as success

`GradeResult.passed` was `not self.errors`, and an abstention produces no Violation. So a run printed `PASS: 5 rule(s) ran, no violations` **twenty-six lines above** `1 declared value(s) NOT DERIVABLE -- not graded, not passed`, reported `pass: true`, and exited 0.

Measured on the tracked corpus in the **default** emit-then-grade round trip: **5 of 22 boards** — `ulx3s`, `tigard`, `watchy`, `glasgow_revC`, `orangecrab_ext_pll`, every one abstaining on `overlap_area`. Committed as `tests/713_abstention_census.json`.

`complete` and `not_graded` reuse `route_summary.py`'s existing name rather than inventing a second channel, as the issue asks. `not_graded` names *which* channel — an aggregate verdict cannot say which of its inputs moved (#694).

**The distinction that makes it 5 and not 21.** My first census counted `rules_skipped` wholesale and reported 21 of 22. That dict holds two opposite things: `_SKIP_REASON` entries all mean *"the intent declares no X"* — **nobody asked**, honest, and firing on 22 of 22 — while `_ARM` entries mean *"the intent asked, this board cannot answer"*. Counting the first would make every board with a minimal intent permanently incomplete. `_is_not_asked` is the single home for it.

It lands on the **wire and the exit code**, not only on `passed`: no production consumer reads that property, so a fix there alone would have been invisible.

## The standing gate

There was **no** source-level gate over `time.time` / `monotonic` / `perf_counter` / subprocess timeouts anywhere in `tests/`. #621 deleted the facility and left nothing to stop it regrowing.

`tests/test_713_wallclock_census.py` requires every clock-carrying file to be registered with a **category** and a reason. The categories encode the real question — not *is there a clock* but *what does the clock decide*: `reporting` (must never reach a comparison — the shape all five survivors had), `throttle`, `record`, `time_claim`, `hang_detector`, `harness`, `file_mtime`, `unused_import`. Three failure modes, not one: an unregistered file, a `reporting` file whose clock reaches a comparison, and an entry naming a file whose clock is gone — so the list cannot rot into folklore.

**Writing it changed the registry twice**, which is the argument for having it: my first draft was written from memory and the gate refused it — six entries named files with no clock, twenty real files were missing, and `placement_gui.py` was registered `reporting` while comparing `QUIET_AFTER_S`.

## What review found, including in my own work

Three blind adversarial passes. They confirmed the load-bearing claims and found **thirteen real defects in my work** plus five false statements in my own commit messages. All fixed here, each with the measurement; the commits carry the detail. The ones that would have shipped:

- **A live ranking defect.** `plane_score` guarded `if not fills:` — the *whole* refill dict — while the rank is fed by `per_net`. A refill that pours copper for some **other** zone but none for the `--plane-score` net returned `status: ok, islands: 0` — the **optimum** of a term sorting before hpwl, with `neck_sum` priced over the wrong copper. That candidate outranks every candidate whose pour was real. This is the defect an earlier commit in this very PR claims to have fixed; it fixed only the total-emptiness case, and **29 boards here carry pre-existing zones**, so `fills` is non-empty regardless.
- **The census could not see three real files**, so its own headline claim was false. `\btime\.time` does not match `_time.time()`; nothing matched `_perf()` after `from time import perf_counter as _perf`; and `done.wait(timeout_s)` passes its timeout positionally. The last is `py_router/ui_thread.py` — a live 120 s hang detector whose expiry returns a bare `False`, the exact shape the category text forbids. My alias fix had gone into the AST visitor, which only runs on files **discovery already found**.
- **`_Decides` missed `AugAssign`** — `spent += time.time() - t0` is the literal definition of the `reporting` category, and was invisible.
- **`_is_not_asked` compared reason strings**, so the `'not requested'` fallback (in no rule's entry) would have marked the **whole corpus** incomplete the first time a rule was added without a skip reason.
- **An incomplete grade was silent whenever the board also had violations** — the reader fixes the errors, sees a clean run, and still has not measured what was never measured.
- **My own oracle guard re-introduced the cost I criticised**: `timed_out is not None` is always true, so it paid a full board parse on all three calls where upstream paid one.

Five false statements of mine, corrected in the commits rather than amended away — including claiming a uniform key set for `plane_score` on the wire when the ok arm carried four of six (a `KeyError` on the happy path, in a PR whose headline is uniform key sets).

## Gates

`tests/test_713_refill_status.py` (61 checks), `test_713_plane_score_ranking.py` (23), `test_713_probe_verdicts.py` (26), `test_713_oracle_scope.py` (19), `test_713_wallclock_census.py` (25). Two mutation batteries: `mutate_713_phase1.py` **17 rows, 17 killed**; `mutate_713_census.py` **8 rows, 7 killed, 1 recorded expected survivor with its reason**.

Every arm asserts the **reason**, never merely that a result was `None` — a check that asserts only `is None` passes for all five causes at once and would have stayed green through the entire defect. Timeout arms are driven by monkeypatching the specific failure, never by a slow board: a timeout test that waits 300 s is a machine-speed test of a fix whose purpose is to remove machine speed from the answer.

Six tests moved, **none loosened**. Each was pinning the defect and each gets the claim it actually meant (`not r.errors`), plus an explicit completeness assertion. `test_704_decap_emission` is *strengthened* from `accept=True` to a refusal check with its reason.

## What these gates do NOT cover

- **A budget added inside a `hang_detector` file is not caught.** Those files legitimately compare a clock, so the `reporting` comparison rule does not run on them. Recorded as an expected survivor in the battery, which fails if it is ever killed.
- **`ui_thread.save_board_on_ui_thread` still returns a bare `False` on expiry.** Registered and named; changing the GUI save path is its own change.
- **`converge.cmd_poses` still hand-builds a probe row of a different shape.** Pre-existing, byte-identical upstream, unmigrated.
- **`board_bounds` is pose-dependent** when the outline lives inside a footprint, so `no_bounds` being "uniform" is true because it can never fire, not because the staged outline is identical across candidates.
- **`_ARM` and `edge_seating_abstained` fire on zero tracked boards** under the default protocol, so their gates are synthetic. Said out loud rather than shipping two channels whose only evidence is that they compile.
- **The plane-scoring worst case rises** from 300 s to ~12×300 s before the refusal. Bounded by a count, not free.

## Verification

- **Full suite: 503 passed, 0 failed, 0 timed out, 0 skipped** (498 at `upstream/main` + the 5 new gates). The baseline is recorded in the first commit, so any red would be attributable.
- `run_doc_examples`: 4 failed, **all pre-existing** — `FileNotFoundError` on the untracked fixture `kicad_files/test_diffpair_ram.kicad_pcb`.
- wx-free parity gates pass: `test_manifest_plan_parity`, `test_cli_postpass_coverage`, `test_cli_defaults_parity`, `test_engine_kwarg_parity`. No new CLI-only post-pass.
- **KiCad python was checked, not assumed** — pcbnew 10.0.0 / wx 4.2.2 are present here.

Two inherited expectations did **not** reproduce and are not carried forward: the suite is fully green at `upstream/main`, and `test_portfolio_determinism` completes in 342 s rather than exceeding `run_all`'s 600 s default.


edgehero:
Follow-ups from the "what these gates do NOT cover" section are now filed, so they outlive this PR's prose:

- #828 — `ui_thread.save_board_on_ui_thread` reports a 120 s timeout and a save failure as the same bare `False`. Registered in the new census with the weakness named; the census could not previously *see* it, because the timeout is positional.
- #829 — `plane_score` assumes the staged outline is pose-invariant; a footprint-embedded outline makes it not. Reachable from a board shape the parser docstring names, but **0 of 22 tracked boards exercise it**, so it is unmeasured and filed as such.
- #830 — `route_verdict` scores a truthy-but-malformed summary as a perfect probe (`{'x': 1}` → `(0, 'clean')`). The layer below the two verdict channels this PR fixed.
- #831 — the deterministic pre-flight predicate #713 item 4 asks for, deferred here with its reasoning and the measurement it needs.

Not filed, deliberately: the ~12x plane-scoring worst case (bounded, and a direct consequence of the refusal being a demand), and the census's hang-detector blind spot (a recorded expected survivor in `mutate_713_census.py`, which fails if it is ever killed).

edgehero:
Correction to my earlier comment: the four follow-ups were independently validated and **three of the four bodies were wrong in ways that mattered**. All are now rewritten; the numbers below are re-measured rather than carried over.

- **#829 was not reachable as filed** and has been retitled and rewritten. I cited the zero-pad "outline footprint" the parser docstring names — but `portfolio.free_refs` and `quench` both filter on `if not fp.pads`, so such a footprint can never move. The reachable form is a **pad-bearing** footprint owning Edge.Cuts (USB-C receptacle, card slot, castellated module). Measured by injecting one `fp_line` into C1 of splitflap_driver: `C1 in free_refs: True`, bounds `223.52 -> 240.66`. Nothing in `py_placer/` inspects a footprint for board geometry before moving it, so the real fix is the free-set guard, not a `plane_score` assertion. I also withdrew a criticism of `PlaneScoreStatus.uniform`'s docstring — that was my misreading; `uniform` answers bounds *presence*, which is pose-invariant.
- **#830's reachability paragraph was wrong.** `route.py` writes `failed_single`/`open_single`/`multipoint_pads_total` unconditionally, so nothing it emits reaches the keyless-summary bug — that part is a schema-drift tripwire, not a live defect, and now says so. Validation surfaced a **better** one in its place: `scoped_route`'s `json.load` is unguarded, so a truncated `--json-out` raises `JSONDecodeError` straight through `probe_route` — which this PR documents as returning a row *whatever happens*. Removing the probe `timeout=` makes a half-written file more likely, not less, so that one is on me.
- **#831 kept its headline but lost three of five supporting claims.** `_fill_cost_key` *is* used as a gate (by this PR, `kicad_oracle.py:304`); the `W*H > 40M` guard is 35 lines away and measures grid cells, not board complexity; and "`_refill_memo_key` discards the stat sizes" is flatly false — `st_size` is in the key.
- **#828 survives, with two corrections.** It is not the sole caller: `swig_gui.py:3427` reaches it via the alias `gui_utils.save_board_via_ui_thread` and handles a `False` **silently**, which a fix scoped to `_live_fill` would miss. And each arm *does* print distinctly, so the defect is that a programmatic caller cannot branch, not that an operator cannot tell.

Issue #862, opened 2026-09-03
Placement: eight unnetted alignment marks make all 379 of a BGA's balls "interior", so it reports demand 0 on every face


edgehero:
**Found while:** fixing #850, which puts both lane ledgers on one face rule.
It is the reason that PR does *not* claim ulx3s as a win.
**Verified against:** `upstream/main` @ `e239e067` and the #850 branch — the
number is the same on both, because this predates either ledger.

## The defect

`escape.face_of` decides whether a pad can escape sideways by measuring its
copper edge against `CopperGeometry.copper`, the union of **every** pad's
copper. A pad further than `max(pad_pitch/2, INTERIOR_EPS)` from all four
edges is INTERIOR: it needs a via, not a lane.

That is a bounding box, and a bounding box cannot tell a few small marks
outside the pad field from a ring that encloses it.

## Measured — ulx3s U1

An LFE5U BGA, 389 pads, **379 netted balls**. Eight **unnetted**
0.127 × 0.508 mm oval alignment marks sit at the corners:

```
ball field   x [130.680, 146.280]   y [80.000,  95.600]
copper box   x [129.726, 147.234]   y [79.046,  96.554]   <- the 8 marks,
                                                             0.954mm out on
                                                             all four sides
pad pitch 0.800   tolerance 0.400
```

0.954 > 0.400, so **every one of the 379 balls is interior**, the whole outer
ring included:

```
interior 379 / 379      demand   north 0   east 0   south 0   west 0
```

A 379-ball BGA that asks nothing of any of its four channels. Only the eight
unnetted marks are assigned faces (2 per face).

The outer ring can obviously escape — there is 0.954 mm of clear space between
it and the marks. What the box models as "enclosed on all sides" is in fact
"inside the convex hull of eight corner dots".

## Why it matters now

`escape` has reported it this way since it was written, and it only reached
one instrument. #850 puts `routability.face_lane_ledger` on the same rule, so
the number is now **published by `check_channels`** as well: ulx3s goes from
68 deficit lanes at the finest grid to 0. The two ledgers agree, which is what
#850 is for, and they agree on a wrong number.

The same shape reaches `board_brief`'s `measurements.escape.*` keys and
`options.move_blocker`.

## Corpus scope

Refs whose `copper` box is set by a pad carrying **no net**, at clearance 0.2
over the 22 tracked boards — 6 of 97 fine-pitch refs:

| ref | netted pads read interior |
|---|---|
[...]

Issue #849, opened 2026-09-02
Placement: face_lane_ledger re-parses the board's courtyards once per REF, and it is 118% of the work the ledger itself does


edgehero:
**Found while:** fixing #841 (both lane ledgers on one obstruction rectangle).
**Verified against:** the #841 branch. Pre-existing -- #841 added ~0.01 s to it.

## The defect

`face_lane_ledger` grades ONE ref per call, and `check_channels` calls it once
per auto-detected fine-pitch ref (`check_channels.py:232`). Every call rebuilds
the whole board's per-part geometry from scratch:

```python
# routability.py:914
parts = build_part_pads(fps, clearance)
# routability.py:968
_graded = graded_parts_from_file(pcb_data, pcb_file)
```

`graded_parts_from_file` reaches `part_local_bounds`, which calls
`parser.extract_courtyard_sides(path)` -- a **full text parse of the
`.kicad_pcb`** (`legality.py:983-993`). Once per ref.

## Measured -- glasgow_revC, 24 fine-pitch refs, clearance 0.09

The absolute seconds are load-sensitive on a busy machine, so the robust
figure is the ratio and the totals, both of which reproduce:

* the 24-ref sweep takes **12.6 - 15.8 s**;
* with `graded_parts_from_file` + `build_part_pads` + `part_copper_geometry`
  hoisted out of the loop it takes **~0.44 s**;
* per ref, the rebuild is **~117%** of what the ledger itself costs
  (min-of-5 interleaved: rebuild 0.415 s against ledger 0.354 s). A single
  paired sample on a loaded machine ranges 72% - 149%, so quote the min-of-N,
  not one pair.

So `check_channels` on this board spends slightly more time re-deriving
geometry it already had than computing lane supply, and roughly 97% of the
sweep is avoidable.

Not a crisis -- but `check_channels` is in the placement fix loop, and 15 s per
call is felt there.

## Shape of a fix

Two optional keywords on `face_lane_ledger` (`graded=` and `parts=`, or one
resolved-context object), built once by `check_channels` around its ref loop
and passed in. `escape_ledger` already uses exactly this pattern -- it resolves
`board_side_map`, `board_container_refs` and `board_copper_geometry` once and
threads them into every `part_escape` (`escape.py:947-950`), with the reason
recorded in `board_side_map`'s docstring.

The keywords must default to None and rebuild, so the two existing test callers
(`tests/test_run6_check_channels.py:49`,
[...]

Issue #847, opened 2026-09-02
Placement: the escape band floor was calibrated against courtyards, and it is now the parameter that decides the starved-face gate


edgehero:
**Found while:** fixing #841 (one obstruction rectangle for both lane ledgers).
**Verified against:** the #841 branch, and `upstream/main` @ 8aa76378 for the before column.

## The mechanism

Both lane ledgers look a fixed depth off a face for neighbours:

```python
# routability.py:951-952
band = (escape_band_mm if escape_band_mm is not None
        else max(1.0, 4 * pitch_routed))
# escape.py, part_escape
reach = reach_mm if reach_mm is not None else max(lane * 4.0, 1.0)
```

At `--clearance 0.09` the routed pitch quantizes to 0.20 mm, so `4 * pitch` is
0.80 and the **1.0 mm floor** is what binds.

That floor was chosen while a neighbour contributed its **courtyard**. A
courtyard is a body plus an assembly skirt, so shrinking the charged rectangle
to the copper (#841) silently narrows the search by the width of that skirt --
the band and the rectangle are one model, and only one of them was changed.

## Measured

`wk/run7/glasgow_revC/{rL_repair,perturbed}.kicad_pcb` at `--clearance 0.09`,
the wrong-basin fixture `tests/test_run8_starved_face_gate.py:91-102` uses.
U30 is an 8.35 x 8.35 mm pad field inside an 11.0 x 11.0 mm courtyard: a
1.325 mm skirt. Its courtyard reached into U1's 1.0 mm band; its copper sits
1.775 mm out and does not.

All at `--clearance 0.09`, track width 0.127, grid 0.05 -- ONE basis, because
the CLI resolves the board's own track width and reports different supplies
(28 / 2 / 2 / 2, U30 at 37.88 lanes) for the same conclusion.

| escape band | U1 east supply @ finest | top blocker | `--gate` exit | NEW |
|---|---|---|---|---|
| 1.0 (the shipped floor) | 25 | C14 | 0 | 0 |
| 1.5 | 25 | C14 | 0 | 0 |
| **2.0** | **1** | **U30 (30.3)** | **4** | **2** |
| 2.5 | 1 | U30 | 4 | 2 |
| 3.0 | 1 | U30 | 0 | 0 |
| **4.0** | 1 | U30 | **4** | **1** |

Two things follow, and they point different ways:

* **The detection is band-gated, not gone.** The wrong-basin board still has a
[...]

Issue #841, opened 2026-09-02
Placement: the two lane ledgers still disagree about which rectangle a neighbour contributes, and it is worth 2-7x on five boards


edgehero:
**Found while:** fixing #834/#835 (PR pending). The side axis of the two lane
ledgers' disagreement is fixed there; this is the axis that is left.
**Verified against:** `upstream/main` @ 8aa76378 plus the #834/#835 branch.

## The disagreement

`escape.escape_ledger` and `routability.face_lane_ledger` both answer "how much
of this face's span is eaten, and by whom". After #835 they share one kernel
(`escape.span_eaten`) and agree on the interval union, the clamp, the symmetric
side test and the container exemption. They still disagree on **which rectangle
a neighbour contributes**:

| | `escape._blocked_span` | `routability.face_lane_ledger` |
|---|---|---|
| neighbour rect | pad bbox (`_part_rect`) | courtyard (`graded_parts_from_file`) |

`routability.py:862-864` states its choice and justifies it by consistency with
the assembly channel — *"lanes eaten by NEIGHBOR BODIES … via the same
courtyard geometry the assembly channel uses"* — not by what obstructs a track.

## Measured

`escape_ledger` deficit lanes, with **only** the neighbour rectangle swapped
(the escaping part's own face geometry is held at its pad bbox, so this is the
isolated variable), on the #834/#835 branch:

| board | pad bbox (today) | courtyard |
|---|---|---|
| glasgow_revC | 17 | **125** |
| orangecrab_ext_pll | 70 | **147** |
| rp2350_fpga_eensy_prePlane | 53 | **91** |
| watchy | 25 | **88** |
| tigard | 41 | **75** |
| ulx3s | **0** | 19 |
| kit-dev-coldfire-xilinx_5213 | **0** | 18 |
| splitflap_driver | 0 | 0 |

Two to seven times, and `kit-dev-coldfire` and `ulx3s` stop being
zero-deficit boards — `kit-dev` is the control
`docs/placement-predictors.md` uses to argue the escape instrument is inert
where the halo term is largest.

## Why it is not obvious which is right

The two rectangles answer different questions, and a *lane* ledger wants
the first:

- What obstructs a **track** is foreign **copper**.
- What obstructs a **part** is a foreign **body**.

A courtyard is an *assembly* keep-out — pick-and-place and rework margin, drawn
deliberately beyond the copper. A track may legally run under a neighbour's
[...]

PR #863, opened 2026-09-03
#850/#848: one face rule for both placement lane ledgers

### LINKED ISSUE #848: Placement: a lane ledger charges a cross-side through-hole neighbour its whole pad box, not just the pads that reach the escaping part's face
**Found while:** fixing #841. Deliberately left out of that PR, with the
measurement below, because it is a fourth rectangle refinement and would have
confounded a 2-4x change with a 0x one.
**Verified against:** the #841 branch.

## The defect

After #835 a neighbour is skipped when it shares no board side with the
escaping part, and a **through-hole** neighbour is kept because its barrels
occupy both faces. That much is right. What it then contributes is
`PartPads.extent` -- the box over **all** its pads, both sides.

So an F.Cu part escaping past a B.Cu connector with a few through-hole pins is
charged the connector's whole SMD pad field on the back, not just the drilled
pins that actually reach the front.

`legality` already makes exactly this distinction for courtyards:

```python
# legality.py:280-287
def rect_on(side_wanted, own_side, courtyard_rect, tht_rect):
    """Its own side gets the full courtyard; the opposite side gets only the
    drilled-pad bounding box (None when the part has no drilled pads)."""
```

and #834 added the per-side extent the copper channel would need:

```python
# legality.py:2292
def extent_side(self, x, y, rot, side): ...
```

Neither is used by either lane ledger.

## Measured -- and the effect today is very nearly nothing

Over the nine boards the #841 PR tabulates, at clearance 0.2:

| board | (ref, side) boxes that differ from `extent` | CHARGED blocker instances on a deficit face that would shrink |
|---|---|---|
| glasgow_revC | 5 | 0 |
| watchy | 5 | 0 |
| rp2350_fpga_eensy_prePlane | 2 | **1** |
| tigard | 2 | 0 |
| ulx3s | 2 | 0 |
| orangecrab_ext_pll | 1 | **2** |
| esp_prog | 1 | 0 |
| kit-dev-coldfire / splitflap_driver | 0 | 0 |
| **total** | **18** | **3** |

**Eighteen boxes differ at all, three of them are ever charged against a face
in deficit, and the reported deficit moves on zero of nine boards.**

That is the whole reason this is filed rather than built: it is correct, it is
cheap, and today it buys nothing measurable. It is worth having when a board
finally exercises it, and worth NOT having ridden along with a change whose
effect it would have muddied.

*(An earlier draft of this issue said "72 neighbour boxes shrink". That number
did not reproduce under any scoping and is retracted -- it came from a
different, unstated basis. The table above is the measurement, with its scope
named.)*

## Shape of a fix

`face_lane_ledger` and `escape._blocked_span` already resolve the escaping
part's own `sides_occupied(...)`. Charge `extent_side(..., side)` per occupied
side and union the results, rather than `extent`. Holes stay in both sides'
boxes -- `extent_local_side` already does that, and a drill removes copper on
every layer.

## Verification

A board where an SMD-heavy part on one face carries a few through-hole pins
and sits opposite a fine-pitch part. On the current corpus the change is
inert, so the arm must be a hand-built fixture, and the table above is the
negative control: nine boards, zero movement in the reported deficit.

### LINKED ISSUE #850: Placement: the two lane ledgers now agree on obstruction and still disagree on which face a pad points at -- one has a tolerance and an interior bucket, the other has neither
**Found while:** fixing #841, which settled the *obstruction* half of the same
disagreement. This is the demand half, and it is the one that is left.
**Verified against:** the #841 branch.

## The disagreement

Both ledgers bucket a part's pads to faces. They do it differently, and
neither file records that the other exists.

`escape._face_of` (`escape.py:631-660`) has a tolerance and an interior bucket:

```python
tol = max(pitch / 2.0, INTERIOR_EPS)
d = {'west': px0 - minx, 'east': maxx - px1,      # pad COPPER edges, #841
     'north': py0 - miny, 'south': maxy - py1}
if min(d.values()) > tol:
    return None                                   # interior: needs a via
```

`routability.face_lane_ledger` (`routability.py:942-946`) has neither, and
measures pad CENTRES:

```python
d_edges = {'N': abs(pad.global_y - ext[1]), 'S': abs(pad.global_y - ext[3]),
           'W': abs(pad.global_x - ext[0]), 'E': abs(pad.global_x - ext[2])}
demand[min(d_edges, key=d_edges.get)].add(nid)
```

So on the routability side **every** netted pad is demand on some face -- a
BGA's interior balls included, which cannot escape sideways at any pitch and
need a via. `escape` reports those separately and says why (its module
docstring: *"Rolling it into a face's demand would blame the face for a fanout
problem"*).

## Measured -- pads whose face changes if the centre becomes the pad's copper edge

| board | netted pads | assignment moves on |
|---|---|---|
| glasgow_revC | 460 | **60** |
| orangecrab_ext_pll | 613 | **40** |
| ulx3s | 586 | **20** |
| rp2350_fpga_eensy_prePlane | 190 | 8 |
| watchy | 121 | 4 |
| tigard | 146 | 0 |

That is only the centre-vs-edge half. The missing interior bucket is separate
and larger: `escape` classifies 402 of ulx3s's pads and 306 of orangecrab's as
interior, and every one of them is face demand to `face_lane_ledger`.

## Why it matters

`check_channels`' verdict is `deficit = demand - supply`. #841 made the two
ledgers agree on **supply**; they still disagree on **demand**, so the same
face can be short in one instrument and fine in the other for a reason neither
file states. `check_channels`' deficit is the classification input the
placement fix loop reads as *"is this failure floorplan-shaped"*.

## Shape of a fix

`escape._face_of` is already the shared-kernel shape: it takes a rect, a pad
box and a pitch and returns a face or None. `face_lane_ledger` could call it
directly -- `legality.pad_box(geom[ref], pad)` gives it the pad rect, and it
already has the extent and can get the pitch from `escape.pad_pitch`.

Two things to decide rather than assume, which is why this is filed rather
than done inside #841:

1. **Does this ledger want an interior bucket at all?** Its demand model is
   deliberately different in a second way -- it takes nets with >= 2 owners and
   <= `DISPLACEMENT_MAX_FANOUT`, where `escape` takes an `ignore_net_ids` list.
   Adding one changes `check_channels`' published deficit on every dense board.
2. **The grid.** This ledger's supplies are grid-quantized and `escape`'s are
   not (`routability.py:1039-1042` records why the layer term was kept out for
   the same reason). Sharing the demand half does not oblige sharing the rest.

## Verification

`check_channels --json-out` on glasgow_revC, ulx3s and rp2350 before and
after, reporting demand and deficit per face. A face that loses demand must
lose it to the interior bucket, and the interior count must match
`escape_ledger`'s `interior_pads` for the same ref -- which is the assertion
that the two instruments have actually been reconciled rather than merely
moved closer.

edgehero:
## The last two axes on which the two lane ledgers disagreed

`escape.escape_ledger` and `routability.face_lane_ledger` both answer *"how
much of this face's span is eaten, and by whom"*. #835 gave them one
arithmetic kernel and one side test, #841 one obstruction rectangle, #849 one
hoist, #847 one escape band. Two axes were left, both filed out of #851 with
their measurement:

Closes #850
Closes #848

**#850** is the demand half: `escape` measured each pad by its own copper edge
against the part's copper box, within `max(pad_pitch/2, INTERIOR_EPS)`, and
reported a pad boxed in on every side as INTERIOR — it needs a via, not a lane,
and charging a face for it blames the face for a fanout problem.
`face_lane_ledger` took `min` over `abs(pad CENTRE − extent edge)`: no
tolerance, no interior case, so **every** netted pad was demand on some face,
a BGA's inner balls included. Neither file recorded that the other existed.

**#848** is the rectangle a cross-side neighbour contributes. The side *filter*
has been per-side since #835; the *rectangle* was the box over all that part's
pads on both faces, so an F.Cu part escaping past a B.Cu connector with a few
through pins was charged the connector's whole back-side pad field.

Fifteen commits, each independently revertible -- including one that reverts an
experiment two others made, kept in history rather than squashed away because
what killed it is the most useful thing in this branch.

### #848 first, because it is the inert one

Land #850 first and #848's inertness becomes unfalsifiable — every baseline it
would be compared against has just moved.

`legality.rect_on_sides(geom, sides)` is the copper analogue of the courtyard
channel's `rect_on`, reading a new `CopperGeometry.rect_sides` built from
`PartPads.extent_side` (#834, until now called only by `pair_shortfall`). **It
lands in `legality`, not in either ledger**, because `escape._blocked_span`
charges the same rectangle — a per-side box in one of them re-opens exactly the
disagreement #841 closed. Both call sites bind the intersection they were
already computing and discarding. Holes stay in **both** side boxes, because a
drill removes copper on every layer, which is why a single-sided part's
per-side box equals its whole-part box bit-for-bit.

**#848's own census reproduces board for board.** 20 (ref, side) boxes differ
from `rect` over the 22 tracked boards at clearance 0.2 — glasgow 5, watchy 5,
rp2350 2, tigard 2, sonde_u 2, ulx3s 2, orangecrab 1, esp_prog 1. That is the
issue's 18 on the nine boards it tabulated, plus sonde_u, which it did not
cover.

**But it is not the value no-op the issue predicts, and that is worth saying
rather than repeating the claim.** #848 says the reported deficit moves on zero
of nine boards. The *deficit* half is right — no `deficit_*` moves on any of
the 22, on either ledger. Supply and blocker attribution do:

| | before | after |
|---|---|---|
| `RN3 E` supply | 1 | **2** |
| `RN3 E` charges `U8` | 2.63 lanes | **0.75** |
| `RN5 E` charges `U8` | 1.12 lanes | **not at all** |

on orangecrab_ext_pll, where U8 is a through-hole part whose back-side field was
being billed to two front-side resistor networks. `eaten_by` is advertised as
the move target, so a neighbour charged 2.63 lanes of copper it does not have
on this face sends a fix at the wrong part — the same defect class #847
recorded for a mis-named blocker, arriving through the rectangle instead of the
band. Both ledgers move together, on the same two refs, which is the property
that says #841 is still closed.

### #850: one face rule, and what it does

`escape._face_of` becomes public `face_of`, and the loop that PAIRS it with a
rect, a pad box and a pitch is lifted into `assign_faces`. Publishing the
kernel alone would not have been enough: what disagreed is not the arithmetic
but what the caller feeds it, and there were **three** copies of that pairing —
`part_escape`, `face_lane_ledger`, and `tests/stress/demand_halo_study.py`,
whose own docstring called itself *"the defect #841 exists to remove"*. All
three now call one function. That commit is asserted to be a value no-op: all
22 boards, both ledgers, every published key, zero movement.

Then `face_lane_ledger` calls it. Three things change and they do not all push
one way — the copper-edge measurement and the tolerance can only take a pad
OFF a face, but the `escape.FACES` tie order (N, E, S, W, against this dict's
N, S, W, E) **moves** it between two, so a face's demand can rise. 54 faces do,
the largest by 2 (rp2350 U5 east, 0 → 2; several tie there, so that is an
example rather than a unique witness). Reporting the
falls alone would have measured half the change, which is why the measure
script prints the rise count as a column.

| board | refs | demand | deficit@finest | interior | starved |
|---|---|---|---|---|---|
| glasgow_revC | 24 | 315 → 256 | 75 → **48** | 108 | 0 → 0 |
| orangecrab_ext_pll | 22 | 320 → 185 | 147 → **66** | 306 | 1 → 1 |
| haasoscope_pro_max | 3 | 333 → 105 | 44 → **0** | 586 | 0 → 0 |
| routed_output | 3 | 333 → 105 | 44 → **0** | 586 | 0 → 0 |
| ulx3s | 9 | 339 → 117 | 68 → **0** | 402 | 0 → 0 |
| rp2350_prePlane | 10 | 98 → 83 | 53 → **40** | 37 | 1 → 1 |
| tigard | 10 | 91 → 89 | 37 → 36 | 18 | 0 → 0 |
| watchy | 9 | 64 → 64 | 10 → 9 | 1 | 0 → 0 |
| esp_prog | 1 | 7 → 4 | 0 → 0 | 5 | 0 → 0 |
| kit-dev-coldfire | 2 | 130 → 130 | 0 → 0 | 0 | 0 → 0 |
| **TOTAL** | **97** | **2034 → 1142** | **478 → 199** | **2054** | **2 → 2** |

**The three boards that go to zero are the point, not a symptom.** ulx3s,
haasoscope_pro_max and routed_output are exactly where the two instruments most
disagreed: `escape_ledger` reported deficit **0** on all three while this
ledger reported 68 / 44 / 44 lanes short, and the whole of that gap was
interior pads charged to faces. Spot-checked at the largest fall: U3 on
haasoscope is a **BGA-529 23×23**, whose perimeter is 88 balls — the ledger
assigns 23 + 22 + 22 + 21 = 88 and calls the other 441 interior. Those 441
need vias.

**ulx3s is the exception, and it is not a win — it is issue #862.** That board
agrees at zero for a bad reason: eight *unnetted* 0.127 × 0.508 mm alignment
marks sit 0.954 mm outside U1's ball field and set the box every pad is
measured against, so all 379 of its netted balls read as interior and the part
reports demand 0 on all four faces. `escape` has said that since it was
written; this PR makes the lane ledger agree with it, which is exactly what
#850 asks for, and the agreement is on a wrong number. A bounding box cannot
tell eight corner dots from an enclosing ring, so the fix is an occupancy test
rather than a different box — its own change, on its own evidence. **I tried a
different box on this branch and reverted it**: measuring against the NETTED
pads only fixes ulx3s and breaks `qfn_interior_pads`, the fixture that exists
for this exact question, whose four genuinely-interior pads (`INT1`, `INT2`,
GND, GND — 1.075 mm inside the part's south edge, behind the QFN's unnetted
south pin row) then read as escaping south. Both boards, both directions, in
#862.

**The reconciliation, exact.** `interior_pads` on the row equals
`escape.PartEscape.interior_pads` for the same ref at the same clearance, on
**97 of 97** fine-pitch refs across all 22 boards. It holds only because the
classification is geometric FIRST and the owner filter second — classify the
other way round and the published count is over a subset escape does not
share.

**And demand is not silently eaten.** tigard U3 has 18 interior pads and lost
demand on NO face, because every one of those nets has another pad that IS on
a face. `interior_demand_nets` is 0 there and 13 on rp2350 U2. A net loses its
demand only when it has nowhere else to go, and that is published per row and
printed once per ref, so a fall always names where it went.

**Isolation, asserted.** `length_mm`, both `supply_*`, `eaten_by` and all
three `escape_band_*` keys are bit-identical on every row of every board.
`ext` is still the whole-part extent; only assignment moved to the copper box.
So this commit changed the demand model and provably nothing else — and #848's
supply movement cannot be hiding inside it.

### The ranked risk did not happen, and it was checked rather than assumed

All three `check_channels` gate predicates read `demand_nets`, and two
threshold at `GATE_MIN_DEMAND = 7`. Demand fell corpus-wide by 41%. So the
question that decides whether this ships is whether the gate went quiet.

* **The tracked true positive is byte-identical.**
  `tests/fixtures/run23/tigard_{damaged,placed}` still exits 4 naming the same
  two faces with the same numbers: `U3 W` lost 21% at demand 10, `U3 E` lost
  52% at demand 9. It is the only clean-clone positive `--gate` has.
* **`_starved_faces` moves on no board** (2 → 2 corpus-wide, the same two
  refs).
* **The #847 calibration does not move.** Re-run against the committed
  `measure_847_calibration.json`: every band row of every comparable pair
  identical, `supply drop @ demand>=7 / 0.20` still `adopt: True`,
  `boards_firing` unchanged at demand ≥ 5, 7, 9 and 11. So
  `GATE_MIN_SUPPLY_DROP` and `GATE_MIN_DEMAND` are **not** re-pinned — there is
  nothing to re-pin them on, and #847's rule was that a value without a
  measurement behind it does not move.
* What does move is the unfiltered demand≥1 ladder, reported rather than
  buried: glasgow 24 → 16, orangecrab 37 → 32, tigard 21 → 19, rp2350 11 → 13
  (up), and rp2350 at demand≥5 goes 2 → 1.

### Published-schema change

`check_channels --json` dumps ledger rows verbatim, so five new keys are a
published-schema change: `interior_pads`, `interior_nets`,
`interior_demand_nets`, `face_pitch_mm`, `face_pitch_source`. The last two
because the two ledgers resolve different *lanes* (#847) and the face
tolerance is a **second, separate** basis — reported before it can become
another invisible difference. Measured, `face_pitch_source` is `pad_lattice`
on every tracked ref; the fallback branch is documented and reachable by no
tracked board.

No consumer breaks: the three in-repo readers of `ledgers`
(`tests/test_849_lane_context.py`, `tests/test_847_escape_band.py`,
`tests/test_run6_check_channels.py`) all read named keys rather than asserting
a key set, and the skill drivers use the tool through its exit code and its
printed text. (An earlier draft of this section said *"the only reader is
`test_849`"* — a fact-check found two more, so the conclusion now rests on
three files that were checked rather than one that was assumed.) No top-level
interior map was added to the JSON — it is derivable from `ledgers`, and a
second copy is how two numbers come to disagree.

`GOLDEN_PRE_HOIST` is **split, not re-recorded**: its supply and `eaten_by`
columns still assert equality against `b5c567c7`, which keeps #849's claim
checkable forever, and its demand column moves to a sibling
`DEMAND_AFTER_850` asserted as a transition with the mechanism named. Editing
four numbers inside the old table would have retired the oracle for the other
sixteen and left no record that anything happened.

### Verification, and what it found in my own work

* `tests/test_850_demand_face_of.py` — nine sections; the file **prints its own
  check total** (`OK -- 63 checks in 9 sections`) rather than stating one in
  prose. It stated one at first, and a fact-check falsified it: I had counted
  `check(...)` call *sites*, and several are inside loops over boards, refs and
  pads, so the emitted number is larger and moves with the corpus. Same
  correction in `test_848_side_obstruction.py`. A count maintained by hand is a
  claim; a printed one is a measurement.

  Its spy checks the rect handed to the kernel **by object identity**, not by
  equality: on a part with no NPTH the `rect` and `copper` boxes coincide, so
  `==` would pass against either and the one pairing mistake that matters would
  be invisible — which is why the NPTH refs (rp2350 J2, watchy SW1–SW4) are in
  its board list. Beside it is the invariant derived from the pads rather than
  from the function: every edge of the box `face_of` measures against is
  attained by some pad, which is the property its docstring rests on.
* `tests/test_848_side_obstruction.py` — six sections, 29 printed checks. A
  corpus positive the issue said did not exist, a `span_eaten` spy asserting both ledgers are
  handed an identical obstacle list (their published values cannot do this —
  different pitch, band and quantization), and a runtime fixture for the three
  semantics the corpus cannot discriminate. Reverting `rect_on_sides` fails
  eight checks.
* `tests/measure_850_848_faces.py` — both arms in ONE process, carrying its own
  copy of each replaced rule, so no table depends on checking out a parent and
  the "before" column re-measures every run. `int_R`/`int_E` printed adjacent
  is its negative control. **Its own first run was a lesson**, recorded in it:
  it patched `rect_on_sides` on the wrong module, all 22 boards raised, and it
  printed a table of zeroes under the words *"they agree on every board
  above"*. It now refuses with exit 2 if any board fails to measure.
* `tests/measure_847_calibration.py --diff` had **two gaps of the same shape**,
  both found by this change: the corpus ladder was compared on the *set* of
  firing boards, so glasgow going 24 → 16 firing faces printed `0 value(s)
  moved`; and a pair present in BEFORE and absent from AFTER was silent, which
  is not a corner case — the committed JSON was recorded with the gitignored
  `wk/` family present, so a clean clone drops three of its six pairs and the
  comparator reported 0 anyway. Both fixed; dropped pairs are named and do not
  count as movement, because missing evidence is not evidence of sameness.
* `tests/mutate_850_848.py` — a new battery rather than rows appended to
  `mutate_834_835.py`, because these need different witnesses and therefore a
  different unmutated gate. **The first run was 18 killed, 1 survived, and the
  survivor was a hole in my own tests.**

  `the-face-geometry-moves-to-the-copper-box` builds `faces` from the copper
  box instead of `pp.extent`, moving the face length, the escape band and the
  obstruction span to a smaller box. Two witnesses each should have caught it
  and each could not: the isolation arm toggles the face *rule* and diffs the
  static keys, and this mutation moves `length_mm` in **both** arms equally —
  #849's own warning about equivalence checks, landing in my arm one commit
  after I quoted it approvingly. And `GOLDEN_PRE_HOIST` pins its two refs
  (rp2350 U2, tigard U3), neither of which carries an NPTH hole, so
  `rect == copper` on both and the mutation is a value no-op on exactly the
  refs that are pinned.

  Closed, not re-recorded: a new arm asserts `length_mm` against the
  extent-derived face on the **five** refs corpus-wide where the two boxes
  actually differ (rp2350 J2, watchy SW1–SW4), and asserts the precondition
  `rect != copper` first so a future corpus change cannot turn it into a
  tautology. Final run: **17 rows, 17 killed, 0 survived, 0 undecided, 0
  broken, 0 disagreeing** — six rows left with the experiment they guarded,
  rather than being kept to guard code that is gone.

### What the reviewers found, including the change I had to take back

Three independent passes, none given my reasoning. Between them they falsified
five things, and one of them cost a commit.

**A blind code review killed an entire experiment, and it was right.** Two
commits on this branch made the box a pad is measured against the union of the
**netted** pads' copper, on the argument that an unnetted pad cannot demand a
lane so it must not decide where the netted ones point. The reviewer showed
the argument is true about DEMAND and irrelevant to ESCAPE — unnetted copper
still blocks a track — and that the change makes the interior gate strictly
more lenient, since distances to a subset's box are `<=` distances to the whole.

The corpus proves it on the fixture that exists for the question.
`qfn_interior_pads` U1's pads 34–37 — `INT1`, `INT2`, GND, GND — sit 1.075 mm
inside the part's south copper edge, behind the QFN's unnetted south pin row.
They are interior; the board is in the corpus to say so. The netted box's own
south edge is 12.8 because no netted pad reaches further, so all four landed on
it at distance 0 and were reclassified as escaping south. Its interior count
went 5 → 1.

**And my own census had printed that number as a win.** `qfn_interior_pads U1
5 -> 1` appeared in my "interior pads recovered" column and I never asked which
four pads had moved or whether they should have. The reviewer asked. The whole
line is reverted — `escape.face_of` is byte-identical to `e239e067` again — and
the ulx3s finding that motivated it is filed as **#862** with both boards, since
they fail in opposite directions.

**A fact-check falsified three published numbers.** Two were check counts I had
got by counting `check(...)` call *sites* rather than emitted checks; both files
now print their own totals. The third was *"the only reader of `ledgers` is
`test_849`"* — there are three, all of which read named keys, so the conclusion
held on evidence instead of on an assumption.

**An independent verifier confirmed the load-bearing claims** — the 97-of-97
reconciliation at five different bases including clearance 0.09 (where
`check_channels` actually runs the corpus), the gate's byte-identity across
`--min-demand` 1/5/7/9/11 and four escape bands, and `interior_pads` being
provably non-increasing — and found the same defect the blind review did, from
the other end.

### Suite and gates

`tests/run_all.py -j 4` at HEAD: **519 passed, 0 failed, 0 skipped**, 3775 s,
plus one self-skip (`test_run8_starved_face_gate.py`, whose fixtures live in
the gitignored `wk/`) and one timeout — `test_826_portfolio_lattice.py`, which
**passes alone in 378 s against its own 900 s budget**. It timed out only under
4-way contention, it is not in this diff, and the harness prints "a timeout is
not evidence of a broken test" for exactly this case. Named rather than folded
into a total.

Mutation battery: **17 rows, 17 killed, 0 survived, 0 undecided, 0 broken.**

### Not affected

No `kicad_routing_plugin/` code references either ledger, and this PR adds no
CLI flag, no GUI-facing engine parameter and no post-engine pass. Both wx-free
parity gates were run and pass, and are **reported rather than offered as
coverage** — they are structurally blind to this class of change.
`.gui-parity-checked` is deliberately not advanced: no parity audit was
performed.

`docs/placement-predictors.md` measured as **unmoved** and says so with a dated
paragraph — both its tables re-derived by the two commands it already names,
reproducing 72.2 %, 14 of 80 uncharged pairs, and every deficit/worst pair. It
reports the ESCAPE ledger, and #850 moved the *other* instrument onto the rule
this one already used.

Filed, not fixed: **#862**, the bounding-box enclosure defect that makes a
379-ball BGA report demand 0 on every face. It reaches the published number
through this PR for the first time, so it is disclosed here rather than left
for a reader to find, and it is pinned in `test_850`'s golden so the day it is
fixed that row fails and says so.

Claude-Session: https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

### DIFF
--- a/docs/placement-predictors.md
+++ b/docs/placement-predictors.md
@@ -230,6 +230,24 @@ kit-dev-coldfire 0 / 0 -> 1 / 7, glasgow 14 / 2 -> 16 / 4, rp2350 8 / 11 ->
 10 / 13, orangecrab 18 / 9 -> 20 / 10, tigard 8 / 3 -> 9 / 8, watchy 4 / 4 ->
 6 / 9; only esp_prog stays at 0 / 0.
 
+*After #850* (2026-09-03) **nothing on this page moved, and that is a
+measurement rather than an omission.** #850 changed which face a pad points
+at in both ledgers -- corpus-wide, face demand fell from 2034 nets to 1203 --
+and it moved neither column of either table above. Both were re-derived at
+that tip with the two commands named above: every deficit / worst pair is
+identical, the 72.2% share is identical, and so is 14 of the 80 uncharged
+pairs.
+
+The reason is worth knowing, because "the demand model changed and the
+demand-derived table did not" reads like an error. This page reports
+`deficit = demand - supply`, and the parts whose demand moves hardest have
+supply to spare, so neither number reaches a deficit. On this page's own
+instrument, ulx3s U1 goes from 0 face-demand nets to 18 / 18 / 13 / 18
+(N / E / S / W) against a supply of 21 / 31 / 31 / 29 -- deficit 0 on every
+face before and after. And the one board whose argument here turns on demand,
+kit-dev-coldfire, has no interior pads at all: 0 of its 244 netted pads, on
+either of its two fine-pitch parts.
+
 **What that does to the argument, stated rather than absorbed.** The
 "already charged" share falls 93.6% -> 72.2%, and the pairs the halo misses
 that the demand term WOULD reach go from 0 of 26 to 14 of 93. So the first
--- a/docs/placement-predictors.md
+++ b/docs/placement-predictors.md
@@ -248,8 +248,8 @@ ulx3s U1 with demand 0 on all four faces and 379 of its 379 netted balls
 interior, because the box each pad is measured against is set by eight
 UNNETTED 0.127 x 0.508mm alignment marks sitting 0.954mm outside the ball
 field. A bounding box cannot tell eight corner marks from an enclosing ring.
-That is older than either ledger and is filed rather than fixed here; the
-board's `1 / 5` row above comes from its other parts.
+That is older than either ledger and is filed as **#862** rather than fixed
+here; the board's `1 / 5` row above comes from its other parts.
 
 **What that does to the argument, stated rather than absorbed.** The
 "already charged" share falls 93.6% -> 72.2%, and the pairs the halo misses
--- a/docs/placement-predictors.md
+++ b/docs/placement-predictors.md
@@ -238,15 +238,18 @@ that tip with the two commands named above: every deficit / worst pair is
 identical, the 72.2% share is identical, and so is 14 of the 80 uncharged
 pairs.
 
-The reason is worth knowing, because "the demand model changed and the
-demand-derived table did not" reads like an error. This page reports
-`deficit = demand - supply`, and the parts whose demand moves hardest have
-supply to spare, so neither number reaches a deficit. On this page's own
-instrument, ulx3s U1 goes from 0 face-demand nets to 18 / 18 / 13 / 18
-(N / E / S / W) against a supply of 21 / 31 / 31 / 29 -- deficit 0 on every
-face before and after. And the one board whose argument here turns on demand,
-kit-dev-coldfire, has no interior pads at all: 0 of its 244 netted pads, on
-either of its two fine-pitch parts.
+The reason is that this page reports the ESCAPE ledger, and #850 changed only
+`routability.face_lane_ledger` -- it moved that instrument onto the rule this
+one already used, rather than changing this one. Every number here is
+re-derived at the tip and identical.
+
+Worth knowing while reading the ulx3s rows in particular: `escape` reports
+ulx3s U1 with demand 0 on all four faces and 379 of its 379 netted balls
+interior, because the box each pad is measured against is set by eight
+UNNETTED 0.127 x 0.508mm alignment marks sitting 0.954mm outside the ball
+field. A bounding box cannot tell eight corner marks from an enclosing ring.
+That is older than either ledger and is filed rather than fixed here; the
+board's `1 / 5` row above comes from its other parts.
 
 **What that does to the argument, stated rather than absorbed.** The
 "already charged" share falls 93.6% -> 72.2%, and the pairs the halo misses
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -1259,6 +1259,48 @@ face against `--baseline`), **3** (nothing had a ledger, so the gate did not
 run — this is *not* a pass), and **2** (unreadable board). Without `--gate`
 it is 0 throughout.
 
+### Which face a pad points at, and the interior bucket
+
+Demand is *"nets with a pad **on** this face"*, and "on" is
+`placement.escape.assign_faces` — the same rule the escape ledger uses
+(#850). Each pad is measured by its **own copper edge** against the box over
+the part's **netted** pad copper, within `max(pad_pitch / 2, 0.001 mm)`.
+
+**A pad that is not on any edge of that box is INTERIOR**, and counts toward
+no face's demand. It cannot leave sideways at any pitch — it needs a via — and
+charging a face for it blames the face for a fanout problem. On a BGA-529 that
+is 441 of 529 balls; the ledger assigns the 88 that form the perimeter.
+
+Three keys report it, on every row (they are part-level facts, repeated the
+way `escape_band_mm` is), and the tool prints them once per ref:
+
+| key | |
+|---|---|
+| `interior_pads` | netted pads on no face — **equal to `escape_ledger`'s `interior_pads` for the same ref at the same clearance** |
+| `interior_nets` | distinct nets among them |
+| `interior_demand_nets` | the subset that had **no** pad on any face, i.e. what the four faces actually lost |
+
+The third is the one to read when a demand looks low. A net with one pad
+interior and another on a face still has to leave through that face, so it is
+still demand; only a net with nowhere to go is off the books. Measured, tigard
+U3 has 18 interior pads and `interior_demand_nets` **0**.
+
+`face_pitch_mm` / `face_pitch_source` report the tolerance pitch and whether
+it came from the part's pad lattice or fell back to the lane. This is a
+**second** basis alongside the escape band's, and it is reported for the same
+reason: two ledgers graded at different bases are not comparable, and until
+#847 nothing in the output said which either used.
+
+*Before #850* this ledger took `min` over the distance from each pad's
+**centre** to the whole-part extent edge, with no tolerance and no interior
+case, so every netted pad was demand on some face. Corpus-wide that was 2034
+face-demand nets against 1203, and 478 deficit lanes at the finest grid
+against 199; the boards where the two instruments most disagreed (ulx3s,
+haasoscope_pro_max, routed_output — 68 / 44 / 44 lanes short here against 0 on
+the escape ledger) now agree. Regenerate with
+`tests/measure_850_848_faces.py --table demand`, which prints the two
+ledgers' interior counts adjacent as its negative control.
+
 ### The escape band
 
 Supply is not just face length over lane pitch: a neighbour parked off the
@@ -1285,7 +1327,10 @@ Two things worth knowing before tuning it (#847):
   reported rather than hidden. `escape` uses the raw `track + clearance`;
   `routability` uses the grid-quantized pitch. They disagree on 19 of the 22
   tracked boards (2.2 mm against 2.4 mm at the `routing_defaults` fallback).
-  The `basis` field says which.
+  The `basis` field says which. Since #850 there is a **second** such basis,
+  the face tolerance's — `face_pitch_mm` / `face_pitch_source` above — and it
+  is *not* this one: it is the part's own pad pitch, which is a property of
+  the footprint rather than of the routing parameters.
 * **Deepening the band raises the false-positive rate.** Measured in
   `tests/measure_847_calibration.py`: at a 2.0 mm band the legitimate-restore
   control itself reports a 0.435 loss of escape. The band is a screening
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -1264,7 +1264,16 @@ it is 0 throughout.
 Demand is *"nets with a pad **on** this face"*, and "on" is
 `placement.escape.assign_faces` — the same rule the escape ledger uses
 (#850). Each pad is measured by its **own copper edge** against the box over
-the part's **netted** pad copper, within `max(pad_pitch / 2, 0.001 mm)`.
+the part's pad copper (`CopperGeometry.copper`), within
+`max(pad_pitch / 2, 0.001 mm)`.
+
+**A known limit of that box, since it is a box and not an occupancy test:** a
+few small pads lying outside the main pad field set it for everything inside.
+On `ulx3s` U1 — an LFE5U BGA — eight *unnetted* 0.127 × 0.508 mm alignment
+marks sit 0.954 mm beyond the ball field on all four sides, and all 379 netted
+balls therefore read as interior, so the part reports demand 0 on every face.
+That predates both ledgers and this section does not fix it; it is recorded
+here because the number is published and a reader should know what it means.
 
 **A pad that is not on any edge of that box is INTERIOR**, and counts toward
 no face's demand. It cannot leave sideways at any pitch — it needs a via — and
@@ -1302,7 +1311,7 @@ printed text, not its JSON.
 *Before #850* this ledger took `min` over the distance from each pad's
 **centre** to the whole-part extent edge, with no tolerance and no interior
 case, so every netted pad was demand on some face. Corpus-wide that was 2034
-face-demand nets against 1203, and 478 deficit lanes at the finest grid
+face-demand nets against 1142, and 478 deficit lanes at the finest grid
 against 199; the boards where the two instruments most disagreed (ulx3s,
 haasoscope_pro_max, routed_output — 68 / 44 / 44 lanes short here against 0 on
 the escape ledger) now agree. Regenerate with
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -1291,6 +1291,14 @@ it came from the part's pad lattice or fell back to the lane. This is a
 reason: two ledgers graded at different bases are not comparable, and until
 #847 nothing in the output said which either used.
 
+The three interior keys and the two pitch keys were **added** to the row by
+#850, and `--json` dumps rows verbatim, so that is a published-schema change.
+It is additive: the three in-repo readers of `ledgers`
+(`tests/test_849_lane_context.py`, `tests/test_847_escape_band.py`,
+`tests/test_run6_check_channels.py`) all read named keys rather than asserting
+a key set, and the skill drivers use the tool through its exit code and its
+printed text, not its JSON.
+
 *Before #850* this ledger took `min` over the distance from each pad's
 **centre** to the whole-part extent edge, with no tolerance and no interior
 case, so every netted pad was demand on some face. Corpus-wide that was 2034
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -1272,8 +1272,9 @@ few small pads lying outside the main pad field set it for everything inside.
 On `ulx3s` U1 — an LFE5U BGA — eight *unnetted* 0.127 × 0.508 mm alignment
 marks sit 0.954 mm beyond the ball field on all four sides, and all 379 netted
 balls therefore read as interior, so the part reports demand 0 on every face.
-That predates both ledgers and this section does not fix it; it is recorded
-here because the number is published and a reader should know what it means.
+That predates both ledgers and this section does not fix it -- **#862** --
+and it is recorded here because the number is published and a reader should
+know what it means.
 
 **A pad that is not on any edge of that box is INTERIOR**, and counts toward
 no face's demand. It cannot leave sideways at any pitch — it needs a via — and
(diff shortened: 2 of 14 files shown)


edgehero:
Superseded by #867, which carries every commit from this branch plus the fix for #862.

Closing rather than stacking, because #862 is not a separate change from this one in any useful sense:

- it edits the same `escape.face_of` that #850 makes shared, so a second PR would conflict with this one by construction;
- its golden row was recorded *on this branch* (`test_850_demand_face_of.py`) specifically so that it would fail the day #862 was fixed — and it does, which is how the transition is now asserted;
- and its evidence needs this branch's measurement rig and mutation battery. A stacked PR would have had to rebuild both to say anything.

The concrete gain from landing them together: **this PR's own table stops carrying an exception.** #850's corpus result had ulx3s going from 68 deficit lanes to 0, and the body had to say that is *not* a win — the two ledgers agreed on a number that was wrong, because eight unnetted alignment marks set the copper box 0.954 mm outside a BGA's ball field and made all 379 of its balls read interior. In #867 that ref reads 379 → 308 interior with demand on all four faces, and the agreement is on a number that is right.

#867 is rebased onto current `main` (v0.22.0) and carries `Closes #850`, `Closes #848`, `Closes #862`.

https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

PR #867, opened 2026-09-04
#862: an occupancy test beside the bounding box, so a BGA's balls are not all 'interior'

### LINKED ISSUE #862: Placement: eight unnetted alignment marks make all 379 of a BGA's balls "interior", so it reports demand 0 on every face

edgehero:
## A bounding box cannot tell eight corner dots from an enclosing ring

`escape.face_of` decides whether a pad can escape sideways by measuring its
copper edge against `CopperGeometry.copper` — the bounding box over **every**
pad's copper — and calls a pad further than `max(pad_pitch/2, INTERIOR_EPS)`
from all four edges INTERIOR: it needs a via, not a lane.

Closes #862

Rebased onto `main` at `9dda4114` at the maintainer's request, so this is now
purely the #862 delta — 13 commits, 13 files. The #850/#848 half it used to
carry is on `main` already, via the #863 merge (`fd1db046`); I compared each
file's delta over its own base before and after the rebase and all six are
byte-identical.

#850 made this worse by publishing it through a second instrument.

`ulx3s` U1 is an LFE5U BGA: 389 pads, **379 netted balls**, pitch 0.8 so the
face tolerance is 0.4. Eight **unnetted** alignment marks sit at its corners —
four 0.127 × 0.508 mm and four the same oval rotated, stacked in pairs to form
four crosses — **0.954 mm outside the ball field on all four sides**, and they
set `CopperGeometry.copper` — the box every pad's distance is measured against.
0.954 > 0.4, so every one of the 379 balls read INTERIOR and the part reported
**demand 0 on all four faces**: a 379-ball BGA that asks nothing of any of its
channels, and a `check_channels` deficit of 0 for the best possible reason and
the worst possible cause.

**The rule is now a union of two sufficient conditions.** A pad escapes a
direction when the band it must cross is shallower than
`max(pad_pitch/2, INTERIOR_EPS)` — the old test, kept verbatim — **or** when
the largest **contiguous** free run across that band, in the pad's own strip,
is at least one `track_width` wide with every other pad of the part inflated by
`clearance`. A pad that escapes in no direction is interior.

**Why the box half is kept rather than replaced**, which is the load-bearing
design decision:

* its error is **one-way**. Widening the box is all an outlying unnetted pad
  can do, and that only ever *increases* a pad's distance to it, so the box
  rule can manufacture a false INTERIOR and never a false ESCAPE. #862 is
  exactly that one direction, so a second sufficient condition corrects it
  without taking on a new error.
* the tolerance turns out to have a job the corridor cannot do: it is the depth
  below which a straight-shadow model does not apply, because a track leaving
  a pad turns before it has travelled that far. **Measured**, replacing the box
  test instead *gains* interior pads on orangecrab_ext_pll U10 and
  rp2350_fpga_eensy_prePlane U4 — both a 0.325 mm part with 5 copper pads
  (`Texas_X2SON-4`), whose central GND pad is 0.58 × 0.58 at 45°, so an
  **0.8202 mm** axis-aligned extent, sitting **0.0699 mm** from the copper box
  on west and east: plainly on the boundary, yet its axis-aligned corridor is
  closed by the flanking pads once they are inflated.
* it makes `interior_pads` provably non-increasing, and makes "consult the
  corridor only for box-interior pads" a **proof rather than an optimisation**:
  when `min(d) <= tol` the argmin direction is itself in the escaping set, so
  the filtered argmin is the unfiltered one.

### Measured

`tests/measure_862_enclosure.py`, both arms in one process, 97 fine-pitch refs
over the 22 tracked boards:

| clearance / track | box | union | refs moved | grew | `ulx3s:U1` | `qfn_interior_pads:U1` |
|---|---|---|---|---|---|---|
| 0.09 / 0.10 *(the corpus runs here)* | 2054 | 1953 | 9 | 0 | 379 → **308** | 5 → 5 |
| 0.20 / 0.20 *(the census basis)* | 2054 | 1953 | 9 | 0 | 379 → **308** | 5 → 5 |
| 0.25 / 0.30 *(the CLI's defaults)* | 2054 | 1957 | 8 | 0 | 379 → **308** | 5 → 5 |
| 0.05 / 0.10 *(swept)* | 2054 | 1914 | 19 | 0 | 379 → **308** | 5 → **0** |
| 0.40 / 0.40 *(swept)* | 2054 | 1960 | 8 | 0 | 379 → **308** | 5 → 5 |

The nine at 0.20/0.20: `esp_prog:USB1` 5→0, `orangecrab:U3` 227→223,
`orangecrab:U8` 4→2, `orangecrab:U9` 1→0, `rp2350:U7` 1→0, `ulx3s:GPDI1` 2→0,
`ulx3s:SD1` 12→1, `ulx3s:U1` 379→308, `ulx3s:U9` 4→0.

**ulx3s U1 is the point.** 71 balls recover and its faces go from
`N0 E0 S0 W0` to **`N17 S11 W18 E16`**. A full 20 × 20 ring would be 76 balls;
nine sites are empty, so the ring holds **67** — and all 67 recover, plus **4**
second-row balls that escape through those empty sites. That last four is the
thing a ring-vs-interior rule could not do. 308 remain interior because they really are, and those balls need vias. Its
four faces carry supply 21/31/29/31 at the finest grid, so the board gains real
demand and is still not short.

**`qfn_interior_pads` U1 is the other side of the gate**, and it is checked as
the SAME five pads rather than the same count. They sit behind that QFN's
unnetted south pin row and are genuinely enclosed. The two boards fail in
opposite directions, which is why the issue insists on both.

### Two figures in the issue that do not survive measurement

Both were in the fix's favour, which is the direction to distrust:

* the issue says **eight** unnetted marks; U1 carries **ten** pads with copper
  and no net — the eight corner marks that do set the box, plus `W10` and
  `W11`, two 0.4 × 0.4 mm circles *inside* the ball field, which do not touch
  the box but are obstacles to the corridor.
* the issue predicts the outer ring recovers **~88 balls**. The ring holds
  **67** (a full 20 × 20 ring would be 76; nine sites are empty), and the
  measured recovery is **71**.

### One limit of the model, which is a cliff rather than a slope

The strip a corridor measures is the pad's **own** span, so a pad narrower
than `track_width` can never be rescued however empty the field in front of
it. Physically a track wider than its pad is routine — it necks down at the
joint — and this model says no. Measured on ulx3s U1, whose balls are 0.4 mm
wide, at clearance 0.2:

| track | balls recovered |
|---|---|
| 0.20 … 0.400 | **71** |
| 0.401 | **0** |

That is this PR's headline falling off an edge one thousandth of a millimetre
wide, and `check_channels` floors the track *up* to the fab floor, which can
only move toward it. It is stated in `pad_band`'s docstring and pinned by four
rows in the gate, so it is found from a test rather than from a board.

Not fixed here, deliberately: widening the strip to admit a necked track needs
its own evidence, and the two measured alternatives — a pitch-wide window and
a zero threshold — both break `qfn_interior_pads`.

### The semantics change, disclosed rather than discovered

**`interior_pads` is now a function of `clearance` and `track_width`**, where
before it was a function of the part's geometry alone — because the enclosure
test asks whether a *track* can leave. That is a published-semantics change on
a key `check_channels --json`, `board_brief` and `routability.health` all
carry, so six keys are published on **both** instruments, spelled identically
so the two JSON outputs diff key for key:

| key | |
|---|---|
| `interior_pads_box` | the box rule alone — **basis-free**, the fanout fact without the parameter fact |
| `face_corridor_escapes` | how many the corridor freed |
| `face_corridor_source` | `caller` / `not_measured` / `unmodelled` / `no_pad_boxes` — four facts, not one `unknown` |
| `face_corridor_clearance_mm`, `face_corridor_track_mm` | the basis it ran at |
| `face_pitch_mm` on `PartEscape` | #850 added this to the routability row only, so until now the two instruments could not be compared on the basis they are asserted EQUAL at |

The precedent is `supply_routed_grid` / `supply_finest_grid`, not
`face_pitch_*`: the corridor half is monotone in both arguments, so **an
interior pad that survives the box rule is a fanout fact, not a parameter
fact**. All three counts are published though one is derivable, because
`interior_pads + face_corridor_escapes == interior_pads_box` is then a
conservation law over each ledger's own population — a count that falls with
nothing naming where it went is how a ledger stops looking and reads as a fix.

**There is no clamp**, and that is deliberate: `qfn_interior_pads` U1's pin
rows are 0.25 mm pads on a 0.5 mm pitch, so the gap between pins is 0.25 mm,
and at clearance 0.05 that leaves 0.15 mm — which really does pass a 0.1 mm
track. U1 reading 0 there is the right answer for a board etched at that
floor. The repo
has refused this move twice before (`signal_layer_count`, `via_pitch`) for the
same reason.

### The reconciliation held, and its precondition grew

`interior_pads` is equal on **97 of 97** fine-pitch refs — but "the same ref at
the same clearance" is now "at the same clearance **and the same track
width**". That is not a formality: leave `track_width` out of `test_850`'s
reconciliation arm and `part_escape` resolves orangecrab's own 0.3 against the
ledger's 0.2, and U3 reads 227 against the row's 223. One ref out of 97, and
only that arm would have said so. The measure script's `int_R`/`int_E` negative
control caught the same defect independently, printing a WARN.

### The ranked risk, answered

All three `check_channels` gate predicates are demand-gated
(`GATE_MIN_DEMAND = 7`), so demand *rising* is the direction that could make
`--gate` newly exit 4. Measured, box rule vs union, at all three bases in use:

| | starved faces | deficit faces |
|---|---|---|
| 0.09 / 0.10 | 0 → 0 | 57 → 57 |
| 0.20 / 0.20 | 2 → 2 | 102 → 102 |
| 0.25 / 0.30 | 2 → 2 | 134 → 134 |

Not one face moves in either census. The tracked `--gate` pair
(`tests/fixtures/run23/tigard_{damaged,placed}`) is byte-identical: exit 4, the
same two NEW faces, `U3 W` demand 10 supply 24→19 at 21% and `U3 E` demand 9
supply 27→13 at 52%. *(I had predicted `ulx3s GPDI1 E` would gain a reported
deficit line. It does not. The measurement is the claim.)*

### Goldens: split, not overwritten

`test_850`'s `GOLDEN` is **byte-identical**. Its four columns are the
`e239e067` oracle and the #850 recording, and editing numbers inside it would
have retired that oracle for the four rows that did *not* move and left no
record that #862 happened. The single moved row goes to a sibling
`GOLDEN_AFTER_862` asserted as a **transition**, with the direction (#862 may
only ADD demand and only REMOVE interior pads) as its own check — the part a
plain re-record cannot state. Same shape as #849's `GOLDEN_PRE_HOIST` /
`DEMAND_AFTER_850`.

`test_835`'s `DEMAND` moves on three of eight rows and `EXPECTED` on two; the
unchanged rows are the negative control inside the same table. Its two
DIRECTION asserts are untouched and get *stronger*, and `DEMAND_AT_PAD_CENTRES`
does not move at all — checkably, because that arm passes
`obstruction_rects={}`, so `geom` is None and no corridor is ever built.
`test_849` needs no change **for #862** — both its pinned refs are measured
unchanged, which makes it the strongest negative control in the tree. (It does
change in this PR, in #850's commit, where `GOLDEN_PRE_HOIST`'s demand column
moved to a sibling `DEMAND_AFTER_850`.)

`docs/placement-predictors.md` said *"After #850 nothing on this page moved …
every number here is re-derived at the tip and identical"*. **That claim is now
false**, and the fix is not to delete it: it held because #850 changed only the
other instrument, and #862 changes the rule both share. It is scoped to the tip
it was measured at, and an *After #862* paragraph carries the re-derivation by
the two commands the page already names — deficit table `ulx3s 1 / 5 → 2 / 5`,
halo share **72.2% → 71.8%**, uncharged pairs **14 of 80 → 16 of 82**. The
share *fell*, so this re-measurement weakens the page's own argument by 0.4
points, exactly as the #841 re-measurement did.

### `has_interior_pads` is deleted

Zero callers anywhere — one grep hit, the definition. It answered this same
question from rounded pad-CENTRE local coordinates with no copper, no clearance
and no obstacles. Shipping #862 while leaving an uncalled fourth copy in the
module whose last commits were about there being *one* is an invitation for the
next author to pick the cheap-looking one.

---

## Verification, and what it found in my own work

**A mutation battery: 26 rows, 26 killed, 0 survived, 0 undecided, 0 broken.**
Two of the 17 existing anchors went stale the moment the corridor landed, and
are re-anchored in the same commit rather than found 50 minutes into a run.
Nine rows are new, and **witness bases are chosen per row, not defaulted**,
because two of them would survive a gate pinned only at the bases in use:

* `the-corridor-replaces-the-box-instead-of-uniting` leaves *both* named
  witnesses intact (ulx3s 308, qfn 5) and is identical to the union at
  0.09/0.10; only the 0.20/0.20 direction column sees it.
* `the-threshold-epsilon-is-dropped` only bites where a free run lands exactly
  on the threshold, which on this corpus is clearance 0.05 and 0.40 and nowhere
  else — so **every basis anything actually routes at is blind to it**.

**A mutation run found a hole in my own gate.** The epsilon row *survived* all
68 checks at first, because every basis I was testing is one a caller uses. It
is closed by testing 0.40/0.40 explicitly *for the mutation*, where a bare
comparison puts 33 of ulx3s U1's recovered balls back (308 → 341).

**A blind code review and a fact-check ran against the finished branch, and
between them falsified fifteen things.** Every one is fixed in `8aba0cff`;
the ones worth naming here, because they were defects in the *verification*
rather than in the rule:

* **No golden could see the central design decision.** The whole argument is
  "a union, not a replacement" — and all six `GOLDEN_862` refs read the same
  under *both* rules at *every* basis. The arm claiming to catch it was true by
  construction: `assign_faces` recomputes a face only `if bf is None`, so a
  box-accepted pad can never become interior. Fixed by adding the two refs that
  do discriminate (orangecrab U10, rp2350 U4) and an arm that RUNS the
  replacement and requires the answer to move.
* **A direction arm compared two literals.** `test_850`'s "#862 only REMOVED
  interior pads" was `308 < 379`, both from tables in the same file; no code
  change could fail it. It reads the live ledger now.
* **A 12× performance regression, worst where it bought nothing.** haasoscope
  U3 is a BGA-529 with 441 interior balls and nothing to rescue, and paid
  ~0.9M inner iterations for that answer. `PadCorridors` now carries an exact
  candidate index — **asserted equivalent on 388 ref-bases, 0 differing** —
  and the board goes 0.824 s → 0.177 s.
* **`demand_halo_study` was on the box rule** while both ledgers were on the
  union, which is the third-copy defect its own docstring exists to record,
  arriving through the argument list instead of a reimplementation.
* **`PartEscape` published the wrong pitch** under `face_pitch_mm` — the raw
  lattice pitch rather than the one the rule ran at — which also emitted
  `inf` and made the row invalid strict JSON.
* and the numbers: `legality.rect_sides`'s docstring carried #848's
  nine-board figure (18) labelled as the corpus — it is **20**, the extra pair
  being `sonde_u`, which that issue's table did not cover; the maintainer's
  own census agrees. `docs/utilities.md` carried a value this very change
  moves (2034 → 1142, now **1215**) with its own regeneration command three
  lines below it. And two commit citations named pre-rebase SHAs unreachable
  from this branch, so the battery's recorded verdict could not be reproduced
  at the tip it named.

**An earlier verifier re-derived the whole census from the prose alone and
found a fifth free parameter nobody had named**: whether an obstacle
"intersecting" the band counts when it merely *touches*. My read-only prototype
read it closed; the shipped `free_run` reads it open, and open is right —
`tigard` U3's exposed pad is sub-rects inside a full-size EP rect sharing the
edge `y = 58.225`, and that EP lies entirely *south* of the sub-pad, so it
cannot block a *northward* escape. My published 0.05/0.10 figure was wrong as a
result (1926; it is **1914**), and the shipped code agreed with the verifier
all along. The stake is not a few pads: the obstacle list deliberately includes
the pad's own rect, because open-ness makes self-exclusion free — so the closed
reading blocks every pad against itself and **collapses the union back to the
box rule exactly, 2054 at every basis, both witnesses still reading 308 and 5,
every corpus golden still green, and the rule doing nothing at all**. It is now
pinned directly.

The verifier also corrected two claims I had overstated: "the box total is 2054
at every basis" is *structural*, not a measurement (`CopperGeometry.copper`
carries neither clearance nor track width, so it cannot move), and monotonicity
is a *tautology* of the union's shape rather than a geometric finding — still
worth asserting, but only as a check that nobody turned the union into a
replacement.

**And the measurement script made the exact error it exists to catch.** It
keyed its pad sets on `pad_number`, which is not unique inside a footprint — a
thermal pad drawn as sub-rects, a row of NC pads — so the sets silently merged
them. It printed box 1994 against the true 2054, undercounting by 60 pads and
dropping `orangecrab_ext_pll:U8` off the moved list entirely. Caught by
cross-checking against an independent sweep, not by the script. (My commit
message for that fix says "two of the nine"; a fact-check reconstructed the
buggy form and it is one.)

Two guards earned their keep while I broke them: the measure script REFUSED to
print a table when 14 boards raised on a stale stand-in signature, and
`run_utils.evidence` refused a `--diff` against a json that run had not written.

### Gates

`tests/test_862_enclosure.py` — **121 checks in 9 sections**, printing its own
total. Every expected value in the unit sections is derived in the comment above
it from the fixture's own numbers. Three arms exist because they are not true by
construction: a spy asserting the corridor ran *at all* and was never asked
about a pad the box half accepted; the assertion that
`tests/test_escape_ledger.py`'s synthetic parts are `modelled=False` with an
empty pad table, so its `interior_pads == 9` oracle is preserved by the
degradation path **and witnesses nothing about this change**; and the
`interior_pads_box` invariance across three bases.

`tests/measure_862_enclosure.py` — the committed tables, both arms in one
process, `--out` / `--diff`.

### Suite

`tests/run_all.py -j 4` at the tip: **532 passed, 0 failed**, 6164 s, plus one
self-skip (`test_run8_starved_face_gate.py`, whose fixtures live in the
gitignored `wk/` — it self-skipped before this branch too) and four timeouts
under 4-way contention: `test_431_board_gates` (its own 1200 s budget),
`test_826_portfolio_lattice` (its own 900 s), `test_713_plane_score_ranking`
and `test_interf_u` (the 600 s default). Named rather than folded into a
total. **None of the four references `escape_ledger`, `face_lane_ledger`,
`assign_faces`, `part_escape` or `check_channels`** — they are not in this
change's blast radius — and **all four PASS when re-run alone**, which is what the
harness's own "a timeout is not evidence of a broken test" line asks for.

Mutation battery: **26 rows, 26 killed, 0 survived, 0 undecided, 0 broken, 0
disagreeing with expectation.**

### Not affected

No `kicad_routing_plugin/` code references either ledger, and this PR adds no
CLI flag, no GUI-facing engine parameter and no post-engine pass.
`.gui-parity-checked` is deliberately not advanced: no parity audit was
performed.

Claude-Session: https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

### DIFF
--- a/docs/placement-predictors.md
+++ b/docs/placement-predictors.md
@@ -175,23 +175,23 @@ is not implemented, and this is the measurement rather than a preference:
 
 *It reaches few blockers the shipped halo does not already charge.* The escape
 ledger NAMES the parts eating a face in deficit. Across seven boards there are
-288 such (face, blocker) pairs, and **208 of them (72.2%) already sit inside
+291 such (face, blocker) pairs, and **209 of them (71.8%) already sit inside
 `halo_a + halo_b`** at the A/B harness's own coefficients -- the OFF arm
-already repels them. Of the 80 it does not charge, the demand term fires on
-**14 of the 93 parts involved**. Its whole mechanism is a larger quadratic on
+already repels them. Of the 82 it does not charge, the demand term fires on
+**16 of the 96 parts involved**. Its whole mechanism is a larger quadratic on
 repulsions that already exist, which is the mode
 `docs/placement-optimization.md` records as scattering the layout.
 
 | board | named blockers | already charged |
 |---|---|---|
 | rp2350_fpga_eensy_prePlane | 77 | 55 (71.4%) |
-| orangecrab_ext_pll | 88 | 53 (60.2%) |
+| orangecrab_ext_pll | 89 | 54 (60.7%) |
 | tigard | 39 | 37 (94.9%) |
 | watchy | 48 | 40 (83.3%) |
 | glasgow_revC | 27 | 18 (66.7%) |
 | kit-dev-coldfire | 8 | 5 (62.5%) |
-| ulx3s | 1 | 0 (0.0%) |
-| **total** | **288** | **208 (72.2%)** |
+| ulx3s | 3 | 0 (0.0%) |
+| **total** | **291** | **209 (71.8%)** |
 
 *And it fires where the grader is nearly blind.* The only independent escape
 instrument the placement A/B has is `health_escape_deficit_parts` /
@@ -206,7 +206,7 @@ identical:
 | esp_prog | 0 / 0 | 3 of 16 (19%) | 2.48 mm (U1) |
 | rp2350_fpga_eensy_prePlane | 10 / **13** | **0 of 61** | -- |
 | orangecrab_ext_pll | 20 / 10 | 2 of 154 (1%) | 1.38 mm (J2) |
-| ulx3s | 1 / 5 | 6 of 226 (3%) | 3.03 mm (U2) |
+| ulx3s | 2 / 5 | 6 of 226 (3%) | 3.03 mm (U2) |
 | glasgow_revC | 16 / 4 | 1 of 257 (0%) | 1.50 mm (J5) |
 | tigard | 9 / 8 | 4 of 85 (5%) | 1.51 mm (U5) |
 | watchy | 6 / 9 | 2 of 84 (2%) | 3.03 mm (J3) |
@@ -236,20 +236,42 @@ at in both ledgers -- corpus-wide, face demand fell from 2034 nets to 1203 --
 and it moved neither column of either table above. Both were re-derived at
 that tip with the two commands named above: every deficit / worst pair is
 identical, the 72.2% share is identical, and so is 14 of the 80 uncharged
-pairs.
+pairs. (Both of those numbers move at #862 -- see below. They are left here
+because this paragraph is a record of what was true at THAT tip, not a live
+figure.)
 
 The reason is that this page reports the ESCAPE ledger, and #850 changed only
 `routability.face_lane_ledger` -- it moved that instrument onto the rule this
-one already used, rather than changing this one. Every number here is
-re-derived at the tip and identical.
+one already used, rather than changing this one. Every number was re-derived
+at that tip and identical.
 
-Worth knowing while reading the ulx3s rows in particular: `escape` reports
-ulx3s U1 with demand 0 on all four faces and 379 of its 379 netted balls
-interior, because the box each pad is measured against is set by eight
+*After #862* (2026-09-04) **that stops being true, and both tables move.** The
+sentence above held because #850 changed only the other instrument. #862
+changes the rule BOTH share: a pad the copper box calls enclosed is now asked
+a second question -- is there a track's width of clear copper from its own
+edge to outside the part -- so the escape ledger this page reports moves too.
+Re-derived at the tip with the same two commands:
+
+* the deficit table: **ulx3s 1 / 5 -> 2 / 5**, and nothing else on any of the
+  eight boards. The freed pads are on U1, U9, SD1 and GPDI1, and one more of
+  that board's parts crosses into deficit; the worst deficit is unchanged.
+* the halo share: **72.2% -> 71.8%** already charged, and **14 of 80 -> 16 of
+  82** uncharged pairs the demand term reaches. Both denominators move because
+  a part with no face demand named no blockers, and four parts now do.
+
+**The direction is worth stating rather than leaving to the reader**: the
+share fell, so this re-measurement WEAKENS the "already charged" argument by
+0.4 points, the same way the #841 re-measurement did. It is recorded here for
+the same reason -- a page that only ever reports its numbers holding is not
+reporting.
+
+Worth knowing while reading the ulx3s rows in particular: `escape` USED to
+report ulx3s U1 with demand 0 on all four faces and 379 of its 379 netted
+balls interior, because the box each pad is measured against is set by eight
 UNNETTED 0.127 x 0.508mm alignment marks sitting 0.954mm outside the ball
-field. A bounding box cannot tell eight corner marks from an enclosing ring.
-That is older than either ledger and is filed as **#862** rather than fixed
-here; the board's `1 / 5` row above comes from its other parts.
+field, and a bounding box cannot tell eight corner marks from an enclosing
+ring. **#862** fixed that -- see the *After #862* paragraph below -- and U1
+now reads 308 interior with demand on all four faces.
 
 **What that does to the argument, stated rather than absorbed.** The
 "already charged" share falls 93.6% -> 72.2%, and the pairs the halo misses
--- a/docs/placement-predictors.md
+++ b/docs/placement-predictors.md
@@ -232,7 +232,8 @@ kit-dev-coldfire 0 / 0 -> 1 / 7, glasgow 14 / 2 -> 16 / 4, rp2350 8 / 11 ->
 
 *After #850* (2026-09-03) **nothing on this page moved, and that is a
 measurement rather than an omission.** #850 changed which face a pad points
-at in both ledgers -- corpus-wide, face demand fell from 2034 nets to 1203 --
+at in both ledgers -- corpus-wide, face demand fell from 2034 nets to 1142
+at that tip (1215 at this one, since #862 gave some of it back) --
 and it moved neither column of either table above. Both were re-derived at
 that tip with the two commands named above: every deficit / worst pair is
 identical, the 72.2% share is identical, and so is 14 of the 80 uncharged
@@ -257,7 +258,11 @@ Re-derived at the tip with the same two commands:
   that board's parts crosses into deficit; the worst deficit is unchanged.
 * the halo share: **72.2% -> 71.8%** already charged, and **14 of 80 -> 16 of
   82** uncharged pairs the demand term reaches. Both denominators move because
-  a part with no face demand named no blockers, and four parts now do.
+  a part with no face demand named no blockers, and ulx3s **GPDI1** now
+  does -- that board's named-blocker count goes 1 to 3 (`GPDI1/D51` and
+  `GPDI1/SW1`, both uncharged) and one more pair is charged on orangecrab,
+  taking the corpus total 288 to 291 and the uncharged-pair part population
+  93 to 96.
 
 **The direction is worth stating rather than leaving to the reader**: the
 share fell, so this re-measurement WEAKENS the "already charged" argument by
@@ -273,8 +278,10 @@ field, and a bounding box cannot tell eight corner marks from an enclosing
 ring. **#862** fixed that -- see the *After #862* paragraph below -- and U1
 now reads 308 interior with demand on all four faces.
 
-**What that does to the argument, stated rather than absorbed.** The
-"already charged" share falls 93.6% -> 72.2%, and the pairs the halo misses
+**What that does to the argument, stated rather than absorbed.** (This
+paragraph and the two above it describe the #841 re-measurement; the #862 one
+is in the *After #862* paragraph further up.) The
+"already charged" share falls 93.6% -> 71.8%, and the pairs the halo misses
 that the demand term WOULD reach go from 0 of 26 to 14 of 93. So the first
 claim is weaker than it was: the term is no longer measurably redundant, it is
 mostly redundant. The conclusion does not turn over -- 72% is still most of
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -1304,14 +1304,47 @@ Demand is *"nets with a pad **on** this face"*, and "on" is
 the part's pad copper (`CopperGeometry.copper`), within
 `max(pad_pitch / 2, 0.001 mm)`.
 
-**A known limit of that box, since it is a box and not an occupancy test:** a
-few small pads lying outside the main pad field set it for everything inside.
-On `ulx3s` U1 — an LFE5U BGA — eight *unnetted* 0.127 × 0.508 mm alignment
-marks sit 0.954 mm beyond the ball field on all four sides, and all 379 netted
-balls therefore read as interior, so the part reports demand 0 on every face.
-That predates both ledgers and this section does not fix it -- **#862** --
-and it is recorded here because the number is published and a reader should
-know what it means.
+**A pad the box calls enclosed gets a second question (#862).** A box cannot
+tell a few small pads lying outside the main field from a ring that encloses
+it: on `ulx3s` U1 — an LFE5U BGA — eight *unnetted* 0.127 × 0.508 mm alignment
+marks sit 0.954 mm beyond the ball field on all four sides, so the box rule
+called all 379 netted balls interior and the part reported demand 0 on every
+face. So the rule is a **union of two sufficient conditions**: a pad escapes a
+direction when the band it must cross is shallower than the tolerance **or**
+when a track's width of clear copper crosses that band, with every other pad
+of the part — *including the unnetted ones, which still block a track* —
+inflated by the clearance. A pad that escapes in no direction is interior.
+
+The box half is kept rather than replaced, and that is deliberate. Its error is
+**one-way**: widening the box only ever increases a pad's distance to it, so it
+can manufacture a false interior and never a false escape, which is exactly the
+defect above. Keeping it also keeps the tolerance doing a job the corridor
+cannot — it is the depth below which a straight-shadow model does not apply,
+because a track turns before it has travelled that far. Measured, replacing the
+box test instead *gains* interior pads on two corpus parts whose central pad
+sits 0.07 mm from the box edge.
+
+Measured over the 22 tracked boards, 97 fine-pitch refs at clearance 0.2 /
+track 0.2: interior pads **2054 → 1953**, nine refs move and every one moves
+down. `ulx3s` U1 goes **379 → 308** — 71 balls recover, more than the 67-ball
+outer ring of its 20 × 20 lattice, because some second-row balls escape through
+sites where the outer row has no ball — and its four faces then read demand
+17/11/18/16 against supply 21/31/29/31 at the finest grid, so the board gains
+real demand and is still not short. `qfn_interior_pads` U1 stays at **5**, the
+same five pads: they sit behind that QFN's unnetted south pin row and are
+genuinely enclosed.
+
+> **`interior_pads` now depends on the clearance and the track width.**
+> Before #862 it was a function of the part's geometry alone. It is not any
+> more, because "can a track leave" is a question about track width and
+> clearance — so **a number without its basis is not a number**, and every row
+> publishes the basis it ran at. The dependence is monotone (a coarser basis
+> can only add interior pads) and bounded above by `interior_pads_box`, so *an
+> interior pad that survives the box rule is a fanout fact, not a parameter
+> fact* — the same distinction `supply_routed_grid` / `supply_finest_grid`
+> draws for supply. There is no clamp: at clearance 0.05 a 0.2 mm gap between
+> QFN pins really does pass a 0.1 mm track, and `qfn_interior_pads` U1 reading
+> 0 there is the right answer for a board etched at that floor.
 
 **A pad that is not on any edge of that box is INTERIOR**, and counts toward
 no face's demand. It cannot leave sideways at any pitch — it needs a via — and
@@ -1323,9 +1356,13 @@ way `escape_band_mm` is), and the tool prints them once per ref:
 
 | key | |
 |---|---|
-| `interior_pads` | netted pads on no face — **equal to `escape_ledger`'s `interior_pads` for the same ref at the same clearance** |
+| `interior_pads` | netted pads on no face — **equal to `escape_ledger`'s `interior_pads` for the same ref at the same clearance AND the same track width** (#862 added the second term; measured, price one ledger at track 0.2 and let the other resolve orangecrab's own 0.3 and `U3` reads 227 against 223) |
 | `interior_nets` | distinct nets among them |
 | `interior_demand_nets` | the subset that had **no** pad on any face, i.e. what the four faces actually lost |
+| `interior_pads_box` | the same count under the **box rule alone** — basis-free, and the number to read when you want the fanout fact without the parameter fact |
+| `face_corridor_escapes` | how many pads the corridor freed. `interior_pads + face_corridor_escapes == interior_pads_box` is a conservation law, not a second copy: a count that falls with nothing naming where it went is how a ledger stops looking |
+| `face_corridor_clearance_mm` / `face_corridor_track_mm` | the basis the enclosure test ran at |
+| `face_corridor_source` | `caller` when the corridor ran; `unmodelled`, `no_pad_boxes` or `not_measured` naming which degradation happened instead — three different facts, not one `unknown` |
 
 The third is the one to read when a demand looks low. A net with one pad
 interior and another on a face still has to leave through that face, so it is
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -1329,7 +1329,10 @@ track 0.2: interior pads **2054 → 1953**, nine refs move and every one moves
 down. `ulx3s` U1 goes **379 → 308** — 71 balls recover, more than the 67-ball
 outer ring of its 20 × 20 lattice, because some second-row balls escape through
 sites where the outer row has no ball — and its four faces then read demand
-17/11/18/16 against supply 21/31/29/31 at the finest grid, so the board gains
+17/11/18/16 against supply 21/31/29/31 at the finest grid **as
+`check_channels` resolves that board (track 0.3 / clearance 0.25)** — demand
+is basis-invariant here but supply is not, and at 0.2/0.2 the same faces read
+43/43/40/43 — so the board gains
 real demand and is still not short. `qfn_interior_pads` U1 stays at **5**, the
 same five pads: they sit behind that QFN's unnetted south pin row and are
 genuinely enclosed.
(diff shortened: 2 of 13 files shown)


drandyhaas:
Heads-up before you touch this branch, and an apology: **#863's commits are on `main`** as of `9dda4114`.

That was my mistake. I re-fetched #863's head before merging and confirmed it hadn't moved, but I never re-checked the PR's *state* — you had closed it 13 minutes earlier, at 12:22:08Z, right after opening this one. So your supersede plan got run over by a merge that should not have happened. Sorry for the disruption.

**Nothing was lost or altered.** I compared each branch's delta over its own base, per file: the #850/#848 half here is byte-identical to what landed. The only differences are hunk line-number offsets, because #530 shifted lines in `legality.py` and `check_channels.py`:

```
py_placer/placement/escape.py       IDENTICAL (191 lines)
py_placer/placement/routability.py  IDENTICAL (207 lines)
tests/stress/demand_halo_study.py   IDENTICAL (30 lines)
tests/measure_847_calibration.py    IDENTICAL (68 lines)
py_placer/placement/legality.py     offsets only (@@ -2376 vs @@ -2393)
py_tools/check_channels.py          offsets only (@@ -444  vs @@ -449)
```

## The ask

Could you rebase onto current `main`, dropping the 15 now-redundant #850/#848 commits, so this PR is just the #862 delta?

```
git fetch origin
git rebase --onto origin/main 52cee5ac fix/862-850-848-enclosure-occupancy
```

Merging as-is produces **10 conflicted files**, including add/add collisions on `test_850_demand_face_of.py`, `mutate_850_848.py` and `measure_850_848_faces.py` — pure duplication, no real disagreement, but not something worth hand-resolving.

I tested that rebase locally: **all 13 #862 commits replay with zero conflicts.**

## What I verified on the rebased result

* 10/10 targeted gates pass — `test_862_enclosure.py` (`OK -- 121 checks in 9 sections`), plus test_850, test_848, test_849, test_847, test_835, test_run6_check_channels, test_escape_ledger, test_run8_channels_gate_honesty, test_549_routability.
* `check_channels --gate` verdict **unchanged on all 22 boards**, non-vacuously — 6 of them fire (rc=3). The further demand-model change does not quiet the gate.
* **#862's own claim reproduces**, and it is a live defect on `main` right now:

  | | ulx3s U1 (BGA-529) |
  |---|---|
  | `main` today | `interior_pads=379`, demand **0 / 0 / 0 / 0** |
  | this branch, rebased | `interior_pads=308`, demand **18 / 18 / 17 / 18** |

  71 perimeter balls recovered against the 88 a 23x23 grid has, so the order is right. Until this lands, the largest part on that board reports no escape demand at all and every channel measurement on it is a sweep of green zeroes.

* Your fact-check correction is real and is **currently wrong on `main`**: `legality.rect_sides`'s docstring says `18 (ref, side) boxes differ from rect` where the measurement is 20. My own independent census of #848 found movers on sonde_u, which the issue's nine-board table did not cover — consistent with 20, not 18.

Ping me when it's rebased and I'll re-run the full suite against it.

https://claude.ai/code/session_01UvXZ4VyCb7r1TzDjkDWsNJ

edgehero:
Rebased and force-pushed: **`11ce7d64`**, 13 commits, zero conflicts — matching what you measured. No apology needed; I closed #863 without telling you first, which is the half of that race I own.

`git rebase --onto upstream/main 52cee5ac …` here (`origin` is my fork), otherwise exactly your command.

## What I checked before pushing

Your per-file comparison, run from this side over each branch's own base — the six files #862 touches, delta hashed with `index`/`@@` lines stripped:

```
py_placer/placement/escape.py       IDENTICAL
py_placer/placement/routability.py  IDENTICAL
py_tools/check_channels.py          IDENTICAL
py_placer/placement/legality.py     IDENTICAL
tests/test_862_enclosure.py         IDENTICAL
tests/measure_862_enclosure.py      IDENTICAL
```

The PR is now 13 files, +1951 / −122 — purely the #862 delta.

Re-run against the new base (which has both #863 and #860 under it), because the base moving is exactly when a recorded number stops being a measurement: **10/10 targeted gates**, including `test_862_enclosure.py` `OK -- 121 checks in 9 sections`, plus test_850/848/849/847/835, escape_ledger, run6_check_channels, run8_channels_gate_honesty and board_brief. Full suite running now; I'll post the number rather than let you find it.

## One reconciliation on your table, and it is not a disagreement

You report ulx3s U1 demand **18 / 18 / 17 / 18**; the PR body says **N17 S11 W18 E16**. Both are right — they are the two instruments:

| | per-face demand | `interior_pads` |
|---|---|---|
| `escape_ledger` | N18 E18 S17 W18 | 308 |
| `face_lane_ledger` / `check_channels` | N17 S11 W18 E16 | 308 |

They reconcile on `interior_pads`, which is what #850 asserts, and they are *expected* to differ on per-face demand: escape takes an `ignore_net_ids` list, the lane ledger applies the `>= 2 owners` / `DISPLACEMENT_MAX_FANOUT` filter. Worth having in the thread so a later reader doesn't file it as a defect.

Two figures in your note are haasoscope U3's rather than ulx3s U1's: U1 is a **BGA-381 on a 20 × 20 lattice**, not BGA-529 23 × 23, and its perimeter is **67**, not 88 — a full ring would be 76 and nine sites are empty. So the 71 recovered is 67 ring balls **plus 4** second-row balls escaping through the empty sites, which is the part I find most convincing about the occupancy rule: it recovers more than any ring-vs-interior rule could. (The 88/529/23×23 numbers are real, they are just U3's — the docs use it as the worked example, which is probably where they came from.)

## `rect_sides` 18 → 20

Confirmed independently here too, and thank you for cross-checking it: `sonde_u` (`J1/B`, `J2/F`) is the pair #848's nine-board table did not cover. **That fix is in this branch** (`265e4cdd`), so it lands with #862 — the docstring on `main` says 18 until then.

While you are in there, three more corrections from the same pass that are now on `main` via #863 and are fixed here:

* `docs/utilities.md` says corpus demand is `2034 → 1142`; #862 moves it to **1215**, and its own regeneration command three lines below already disagrees with it.
* `docs/utilities.md` re-defines `interior_pads` as the box rule 35 lines after defining it as the union — that paragraph is `interior_pads_box` now.
* `docs/placement-predictors.md`'s *"After #850 nothing on this page moved … every number here is re-derived at the tip and identical"* stops being true at #862. It is scoped to the tip it was measured at, with an *After #862* paragraph carrying the re-derivation (halo share 72.2% → **71.8%**, uncharged pairs 14 of 80 → **16 of 82**, deficit table ulx3s `1/5` → `2/5`). The share *fell*, so that re-measurement weakens the page's own argument by 0.4 points, the same way #841's did.

## One thing I would rather you saw from me than from a board

The rule has a **cliff**, not a slope. The strip a corridor measures is the pad's own span, so a pad narrower than `track_width` can never be rescued however empty the field in front of it. On ulx3s U1, whose balls are 0.4 mm:

| track | balls recovered |
|---|---|
| 0.20 … 0.400 | 71 |
| 0.401 | 0 |

`check_channels` floors the track *up* to the fab floor, which can only move toward that edge. It is stated in `pad_band`'s docstring and pinned by four rows in the gate. Not fixed here deliberately — widening the strip to admit a necked track needs its own evidence, and the two alternatives I measured (a pitch-wide window, a zero threshold) both break `qfn_interior_pads`.

Ready for your suite run.

https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

edgehero:
Two follow-ups while you were reading, both bookkeeping I created:

**#850 and #848 were still OPEN.** Closing #863 by hand meant its `Closes` directives never fired when you merged it — GitHub only runs them for a PR that is itself merged. I have closed both against `fd1db046` with a note saying the fix shipped there, so the issue list matches `main` again.

**This PR's body and title now describe only #862**, since the #850/#848 half is no longer in the diff. It said `Closes #850 / Closes #848` and "supersedes #863", which would have been three false statements sitting on a 13-commit PR.

One thing I did **not** do, deliberately: I did not leave `Closes #850`/`Closes #848` in this PR to tidy those issues up on merge. It would have worked, and it would have recorded that this PR fixed them, which it does not. The commit that fixed them is `fd1db046`, and that is what their close comments name.

`mergeable_state` is `clean`. Full suite running here too; I will post the number either way.

https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

drandyhaas:
Merged as **`2828476c`**. `#862` closed itself through the linkage, so all six issues across the three PRs are now closed and the issue list matches `main`.

## Full suite, your base

**539 passed, 0 failed, 0 timed out, 0 skipped (+1 self-skipped)**, 2319 s. The count reconciles as 538 (main at `9dda4114`) + 1 for `test_862_enclosure.py`, so nothing dropped out of the run silently.

Your rebase produces a tree **byte-identical** to the one I had verified locally, and fast-forwards onto `9dda4114` with no conflicts.

## My error, and thank you for catching it

You are right, and the footprint says so outright — `lfe5bg381:BGA-381_pitch0.8mm_dia0.4mm`:

```
ulx3s U1: pads=389 netted=379 unnetted=10
          netted lattice 20x20, perimeter ring = 67 (a full ring would be 76)
haasoscope_pro_max_test U3: G529:BGA-529_23x23, 529 pads, lattice 23x23, perimeter 88
routed_output           U3: G529:BGA-529_23x23, 529 pads, lattice 23x23, perimeter 88
```

The 529 / 23x23 / 88 figures are U3's. I took them from #863's body, where they are correctly attributed to haasoscope U3, and carried them onto a different part without re-deriving them — which is the exact failure mode I had just told you to watch for, one comment earlier. Your reading of the 71 is also the better one: 67 ring balls **plus 4** second-row balls escaping through empty ring sites, which no ring-vs-interior rule could recover. That is a stronger argument for the occupancy test than the one I made.

## Your two other claims, reproduced here

**The two instruments.** Both reproduce exactly, and they reconcile where #850 says they must:

| | per-face demand | `interior_pads` |
|---|---|---|
| `escape_ledger` | N18 E18 S17 W18 | 308 |
| `face_lane_ledger` | N17 S11 W18 E16 | 308 |

Agreed that this belongs in the thread rather than being left for someone to file as a defect.

**The cliff.** Reproduced, and it is as sharp as you describe — worth having the numbers on the record:

| track (mm) | balls recovered |
|---|---|
| 0.200 – 0.400 | 71 |
| 0.401 | **0** |

Recorded in the merge commit as a disclosed property, not a regression. In practice fine-pitch signal tracks sit at 0.1–0.15 mm, well clear of a 0.4 mm ball — but your note that `check_channels` floors the track *upward*, i.e. toward the edge, is the part that makes it worth stating.

## On the bookkeeping

Closing #850/#848 against `fd1db046` rather than letting this PR sweep them up was the right call, and the reason you gave is the one that matters: it would have worked, and it would have recorded something untrue. The stranding was mine to begin with — I re-fetched #863's head before merging and concluded nothing had changed, having never checked the PR's *state*.

https://claude.ai/code/session_01UvXZ4VyCb7r1TzDjkDWsNJ

edgehero:
Full suite on the rebased branch (`11ce7d64`, base `9dda4114`):

```
539 passed, 0 failed, 0 timed out, 0 skipped (+1 self-skipped) in 3747.5s
```

The self-skip is `test_run8_starved_face_gate.py`, whose wrong-basin fixtures live in the gitignored `wk/` — it self-skips on `main` too, so it is not this branch's.

Two differences from the run I reported before the rebase, both explained rather than left as noise:

* **532 → 539 passed.** The +7 are `main`'s own new test files, arriving with the #860 and #863 merges. Nothing in this branch adds a test file that was not already counted.
* **4 timeouts → 0.** The earlier run had `test_431_board_gates`, `test_826_portfolio_lattice`, `test_713_plane_score_ranking` and `test_interf_u` time out under 4-way contention; all four passed standalone then, and all four pass in-suite now. That is machine load, not the branch — which is what the harness's own "a timeout is not evidence of a broken test" line predicts, and it is worth saying that the prediction held rather than quietly enjoying the greener number.

Ready when you are.

https://claude.ai/code/session_01A4yGaUMwXoP7xzedneLMW3

Issue #837, opened 2026-09-01
Placement: nothing declares whether a board is single- or double-sided, so the back-side area credit is granted to every board with nothing charged against it


edgehero:
**Found while:** asking whether the placer can be told to use one side or both.
**Verified against:** `upstream/main` @ 3fb9c05f.

## The defect

`grow_board` accumulates part area per side and scores the board on the busier face:

```python
# py_placer/placement/options.py:230-234
# ... A part on B.Cu does not compete for F.Cu area.
busiest = max(per_side.values()) if per_side else 0.0
util = (busiest / usable) if usable > 0 else float('inf')
fits = busiest <= usable
```

The reasoning is right for a double-sided board, and the comment records the measurement that motivated it. But it is unconditional, and there is no way to say a board is single-sided: `_TOP_LEVEL_KEYS` (`floorplan.py:106-110`) has no assembly key, and no `place_*.py` has a side flag. So a board that will be built with one reflow pass is scored as if it had a whole second face available, and nothing in the run discloses the assumption.

`check_assembly.py` — despite the name — has no per-side census at all: no `parts_by_side`, no reflow-pass count.

## Measured

Local corpus, 51 distinct designs: **16 strictly single-sided, 35 double-sided.**

Refined by pad count, three of those 35 carry only zero-pad graphics on the back. `esp_prog` is one of them, so its often-quoted "F=17 / B=3" is **electrically single-sided** and it is the wrong board to cite as a two-sided example.

## Why it matters

A second reflow pass is a real cost the tool models nowhere. Today a placement can be reported as fitting because of area on a face the fab will never populate, and there is no channel — flag, intent key or report field — through which anyone can say otherwise.

## One correction to the obvious schema

`"sides": "single"` is under-specified, because single-sided does not mean front-sided. `ulx3s` is back-dominant, F=66 / B=168, so a rule reading "single implies all on F.Cu" flags **168 of 234 parts on a shipping board**. The face has to be named: either `sides: "F" | "B" | "both"`, or `single` plus a `primary_side` defaulted to the observed majority and disclosed rather than assumed.

## Scope, and the half that is not shippable yet

1. **The declaration and the arithmetic.** An `assembly` key in `_TOP_LEVEL_KEYS`, a `rule_assembly_side` in `RULES` (`floorplan.py:2838`), a per-side census in `check_assembly.py`, and `options.py:232` summing rather than maxing when the policy says one side. All of this is reachable now.

[...]

Issue #836, opened 2026-09-01
Placement: a part's side is fixed at construction, so no move the engine can make puts a component on the other face


edgehero:
**Found while:** asking whether the placer can be told to use one side or both, and to move parts between them.
**Verified against:** `upstream/main` @ 3fb9c05f.
**Depends on:** #714 (the writer half) and, through it, #726.

## The mechanism

`_Part.side` is assigned once and reassigned nowhere:

```python
# py_placer/placement/quench.py:626-628
self.side = footprint_side(fp)
self.has_tht = footprint_has_through_pads(fp)
self.sides = sides_occupied(self.side, self.has_tht)
```

Every move signature is `(x, y, rot)`: `_candidate_positions` (`quench.py:2492`), `_candidate_rotations` (`:2547`), `candidate_valid` (`:1682`), `apply_move` (`:2261`). Side is a filter throughout and a variable nowhere, which the engine says about itself — "`zone_side` is invariant under every move this engine can make" (`quench.py:75`) — and which `docs/floorplan-intent.md:400` records as "**vacuous, not conservative**".

The move is already specified and already marked unbuilt:

> *side flip* (optional; mirrored courtyard): treated as a first-class move
> in recent PCB literature. **Not implemented as a move.**
>
> — `docs/placement-optimization.md:347-348`

#714 asks for the writer that can emit a flip. This is the other half: the search that can *choose* one. Neither is useful alone.

## Why it is worth having

Moving a passive to the back is how a dense board fits. Over the 51 distinct designs in the local corpus, 35 are double-sided, and **62% of all their back-side parts carry two pads or fewer** — precisely the decap-and-testpoint move the search cannot make. Eight designs put at least half their parts on the back (ulx3s 168 of 235; glasgow revD0 and castor_pollux more).

The area credit for that is already granted and never earned. `options.py:232` scores utilisation as `max(per_side)`, so a part on the back reduces it — while nothing in the objective can perform the move that would.

## What it needs beyond the writer

Not a fourth element on a tuple. The parts that are easy to miss:

- **Four snapshot/restore sites store a 3-tuple** and restore with `apply_move(r, *pose)` — `seeder.py:878`/`:936-937` (eviction rollback) and `reconstruct.py:138`, `:1058`, `:1343`, the last calling itself an "exact snapshot restore". A *rejected* flip would not roll back.
[...]

Issue #829, opened 2026-09-01
Placement: the movable set is gated on pads, so a footprint that owns Edge.Cuts can be moved and silently redefine the outline


edgehero:
The placement stack's stated contract is that it **never changes the board outline** (CLAUDE.md, and `plan-pcb-placement`'s description). Nothing enforces it. The movable set is gated on **pads**, not on whether a footprint owns board geometry — so a pad-bearing footprint that carries Edge.Cuts shapes is freely movable, and moving it silently redefines `board_bounds`.

## The mechanism

`kicad_parser._footprint_edge_points` returns *"GLOBAL points of footprint-embedded Edge.Cuts shapes (fp_line/fp_rect/fp_arc/fp_circle/fp_poly/fp_curve), **transformed by each footprint's (at x y rot)**"*, and `extract_board_bounds` folds them into the board bounds by min/max. So the outline is a function of a footprint's pose.

The gate on what may move is pads only:

```python
# py_placer/placement/portfolio.py -- free_refs
    for ref, fp in pcb_data.footprints.items():
        if not fp.pads:
            continue
```

and `quench.py:813-823` likewise drops or locks a zero-pad footprint. **`grep -rn "Edge.Cuts" py_placer/`** finds nothing in the movable-set logic — the only hits are in `design_brief.py` and `fanout_clearance.py`, neither of which gates movement.

## Measured

Inject one Edge.Cuts `fp_line` into `C1` — a real 2-pad part — of `splitflap_driver`, then move it +40 mm in x exactly as a placement candidate would:

```
C1 pads in original: 2
footprint-embedded Edge.Cuts points: 2
C1 in free_refs   : True
free set size     : 65
injected bounds   : (25.4, -5.71, 223.52, 55.88)
moved bounds (+40): (25.4, -5.71, 240.66, 55.88)
BOUNDS CHANGED    : True
```

`free_refs` is what `portfolio.py` uses to pick what candidates perturb. This is not an exotic part shape: a USB-C receptacle, a card slot, an edge connector, or a castellated module with a milling cutout all plausibly carry Edge.Cuts *and* pads.

## What it costs, beyond the outline itself

[...]

PR #869, opened 2026-09-04
#714: placement.writer can mirror a footprint to the other board face

### LINKED ISSUE #714: Placement: placement.writer cannot mirror a footprint to the other side, so a real layer flip is unavailable
Split out of #411, whose corpus and recovery machinery have shipped. That issue asked to "move a subsystem to the wrong side of the board"; what exists is a point reflection through the board centre, which produces the wrong **half**, not the wrong copper layer. `py_placer/placement/perturb.py:38-43` explains why, and names this as the prerequisite:

> **Layer flip (F.Cu <-> B.Cu).** `placement.writer.write_placed_output` rewrites `(at ...)` and pad angles only: it writes no `(layer ...)` and mirrors no pads. A "flip" through it produces a board whose pads contradict its side, which corrupts `footprint_side` / `sides_occupied` and therefore every overlap number downstream. If the wrong SIDE is wanted rather than the wrong HALF, that is a writer capability first.

So side is currently read-only throughout placement: everywhere it appears it is a filter, never a variable. `_Part.side` is set once at construction and no move, swap or rotation touches it.

**Why it is worth having.** Three consumers are blocked on it, in increasing order of value:

1. **A real `wrong_side` damage kind for the #411 ladder.** The corpus is heavily double-sided, so this is a realistic damage class the rig cannot currently stage: ulx3s is 168 of 234 parts on the back, glasgow_revC 94 of 266, orangecrab_ext_pll 84 of 165.

2. **`rule_zone_side` is graded and unsatisfiable.** The intent schema carries `blocks[].side` and `floorplan.py:755-768` fires a violation when a part sits on the wrong side, but the seeder consults `keepouts[].sides` and never `z.side`, and no engine op could act on it if it did. `docs/floorplan-intent.md:139` claims every rule measures with the geometry the optimizer gates on; this is the one exception, and it is a permanent red mark no run can clear.

3. **Side assignment as a placement decision.** On a dense board, moving a passive to the back is how the board fits at all, and it is currently outside the search space entirely.

**What the capability actually needs**, and the part that makes it more than a one-liner: rewriting the footprint's `(layer ...)` is the easy half. The rest is mirroring every pad's layer list front-to-back including mask and paste, mirroring pad local X, mirroring `F.CrtYd`/`B.CrtYd`, `F.Fab`/`B.Fab` and `F.SilkS`/`B.SilkS`, negating the stored rotation per KiCad's convention, and adding `(justify mirror)` to texts.

The risk is that getting the mirror convention subtly wrong is **silent**: every downstream overlap number stays plausible while being computed from geometry that contradicts itself. That is exactly what `perturb.py` warns about.

**Suggested verification**, following the pattern `tests/gui_parity/test_ref_label_pcbnew_parity.py` already establishes for a two-path field: flip a spread of footprints (SMD, THT, castellated, various rotations and shapes) through both the text writer and pcbnew's own `FOOTPRINT.Flip()`, and assert the resolved `Pad.global_x/global_y/size_x/size_y/layers` and `footprint.layer` agree pad for pad. Plus a round-trip assertion that flip-and-flip-back is byte-identical through `write_placed_output`, and a spot check that KiCad's own DRC accepts the result.

Given the silent-failure mode, I would treat the pcbnew parity test as the acceptance criterion rather than a nice-to-have. Without it the only oracle is "the numbers still look reasonable", which is precisely the failure this would introduce.

edgehero:
## A writer that rewrote poses could not say which face it was writing them on

`placement.writer.write_placed_output` rewrote a footprint's `(at x y rot)` and
its pad **angles** and nothing else. It wrote no `(layer ...)` and mirrored no
geometry, so board side was read-only throughout placement — a filter
everywhere, a variable nowhere. This gives it the missing half.

Closes #714
Refs #836
Refs #837

The two dependants stay open on purpose. #836 (side as a *search* variable) is
gated on its own pre-registered measurement, which was not **buildable** until
this writer existed; #837 says its own acting half is unshippable without both.

## Verified

Rebased onto `main` @ `bc0ca9bc`; the diff hash is identical before and after
the rebase, so only the base moved.

* **Full suite: 545 passed, 0 failed, 0 timed out, 0 skipped (+1
  self-skipped).** Reconciles exactly: `main` collects 540 `tests/test_*.py`
  (539 pass + the pre-existing `test_run8_starved_face_gate` self-skip), this
  adds 6, and 539 + 6 = 545.
* **Acceptance gate**: 45 passes over 15 fixtures — 1833 pads and 2502 non-pad
  nodes compared node-for-node against pcbnew 10.0.0, all five required
  surfaces present (`fp_arc`, `fp_curve`, `fp_poly`, `fp_circle`, `model`).
* **Mutation battery**: 24 rows, 20 killed, 4 survived, **0 broken, 0
  disagreeing with expectation**; 24/24 anchors match exactly once and
  `--selftest` proves the `.pyc` defence rather than asserting it.
* **CLI/GUI parity**: `test_manifest_plan_parity.py` and
  `test_cli_postpass_coverage.py` both run, both unchanged.
* The defect was live on `main`: `perturb --kind wrong_side` could only produce
  the wrong HALF, and `write_placed_output` had no way to express the other
  FACE at all.

## The API is one optional dict key

`new_side`, `'F'` or `'B'`. Absent or `None` and the layer is left alone and the
write is byte-identical to one built before this existed — which is what keeps
the ~20 producers of these dicts (`quench`, `seeder`, `portfolio`, `relocate`,
`perturb`, `fanout_clearance`, `place_reconstruct`, `place_seed`, `converge`,
`net_rescue`) unchanged. Anything other than `'F'`/`'B'` raises at the boundary;
`'B.Cu'` is the mistake a caller actually makes.

`new_rotation` still means the **final** orientation. pcbnew's `Flip` negates
it; this does not, because a hidden negation would make the same call mean
different things depending on the input side.

## Why the obvious test cannot see the bug

The issue proposes comparing resolved `Pad.global_x/global_y/size_x/size_y/
layers` plus `footprint.layer` through both parse paths. That is blind to
exactly the failure it is written for:

`local_to_global` has **no mirror term** — KiCad stores B-side children
pre-mirrored. Toggle the footprint `(layer ...)` to `B.Cu` and toggle the pad
`(layers)`, but forget the local y, and every pad's **global** position comes
out exactly as before, on **both** parse paths, while `footprint.layer` reads B.
Every field the issue names passes. `legality.footprint_side` reads `fp.layer[0]`
and nothing cross-checks the pads, so the board grades plausibly and wrongly
forever.

Three more reasons the comparison had to move down a level:

* Comparing two parsers over one file tests the parsers. The oracle has to be a
  different **document** — pcbnew's own flip of the original.
* PCBData models a minority of a block. Of 1319 pad-bearing tracked footprints,
  1206 carry `fp_line`, 1215 `model`, 1077 `fp_text`, 132 `fp_circle`, 85
  `fp_poly`, 54 `fp_arc`, 46 `fp_rect`. No pad field compares any of them.
* `fp_arc`'s start/end swap is **geometrically invisible**: mirroring
  start/mid/end without exchanging start and end gives the same curve through
  the same three points.

So the acceptance gate compares the emitted **block** against the block pcbnew
writes after `FOOTPRINT.Flip(pos, FLIP_DIRECTION_TOP_BOTTOM)` + `SaveBoard()`,
as canonical trees — numbers to integer nanometres (both sides are nm-quantised,
so the tolerance is exactly **zero**, not a float epsilon), children sorted
because pcbnew re-serialises in its own order, and the uuid multiset **asserted**
rather than normalised away.

## The gates were written first, against the writer that could not flip

Six of them, committed before the engine. Five were red then, one green, and
that is the evidence they are gates rather than decoration.

| gate | before | now |
|---|---|---|
| IDENT — no-side-key write byte-identical, 22 boards / 1349 footprints | GREEN | GREEN |
| SELF — a flipped part vs its own pads' layers | RED | GREEN |
| VAC — the mirror asserted numerically, on fixtures *proven* to discriminate | RED | GREEN |
| RT — flip and flip back, byte-identical | RED | GREEN |
| REF — 13 refusals, each naming its construct | RED | GREEN |
| PAR — 45 passes vs pcbnew, node for node (1833 pads, 2502 non-pad nodes) | RED | GREEN |

## The vacuity problem

Measured over the 22 tracked boards, the symmetric fraction depends on the
predicate, and the three answer different questions:

| predicate | | |
|---|---|---|
| position `{(x,y)}` closed under `y → -y` | 1198/1319 | **90.8%** |
| tuple `(num,type,shape,x,y,angle)`, angle negated | 532/1319 | **40.3%** |
| the same tuple, **angle held identical** | 1011/1319 | **76.6%** |

The third is the one an implementer needs — *how often is the y-mirror itself
unobservable*, against the mutant that negates angles and layers but forgets y.
**A fixture drawn from ordinary parts passes every geometric assertion with the
mirror deleted**, so VAC proves per surface that its fixtures can see what its
assertions claim to test, and refuses if they cannot.

## Two rounds of adversarial verification, and what they found

**Round 1** (against the phase-0 measurement) found a live bug — pcbnew adds
`(justify mirror)` only to a text on a *sided* layer, and this added it
unconditionally. 28 flip-eligible texts on the corpus sit on a User layer and
**none of the 13 parity fixtures carried one**, so the acceptance gate was green
with the bug in it. It also falsified four of my numbers, all four in the
feature's favour.

**Round 2** (against the whole branch diff) found a bug that was **already
shipping in this PR's own consumer**:

> Pads used the general angle composition `a' = R_new − (a − R_old)`; texts used
> the constant `180 − a`. Those agree **only** when `new_rot == −old_rot` —
> pcbnew's bare flip, which is the only thing the parity gate ever asked for.
> `perturb`'s `layer_flip` holds the pose, so its delta is 2R: **9 of tigard's
> 33 flipped parts shipped a reference designator rotated 180° from where KiCad
> puts it**, relative to their own pads, which had rotated correctly. Now 0/33.

Each fixture is now compared three times — bare flip, +90, +180 — against
`Flip` + `SetOrientationDegrees`. Reintroducing the old formula produces 90
differences, *every one on a composed pass and none on the bare one*.

Round 2 also turned **five silent skips into refusals** — an unparseable
four-token `(at … unlocked)`; a text with no `(effects (font …))`; a **pad
dispatch that was a blacklist**, so the head that refuses at footprint level was
emitted verbatim from inside a pad (KiCad 9/10's `padstack` is the live case); a
block with no top-level `(layer …)`; and `(\S+)\)` swallowing a closing paren,
which made every compactly-serialised `fp_poly` unflippable.

## Mutation battery: 24 rows, 20 killed, 4 survived, 0 broken

**Neither gate is sufficient alone**, and the rows prove it in both directions:

* `the-negation-goes-through-float` is killed by the **round trip alone** — every
  value test passes, because the numbers are right and only their spelling
  changed.
* `fp_arc-mirrors-without-the-start-end-swap` is killed by the **parity gate
  alone** — the round trip survives it (still an involution) and so does every
  geometric check.

**All four survivors carry their reason in the row above them**: two
redundant-guard pairs created by the round-2 fixes, the `zone` refusal covered
twice, and the #829 outline term that no tracked board can exercise.

The battery corrected one of my own justifications. The comment said a float
round trip would rewrite `0.500` as `0.5` — measured, **all 148417 coordinate
literals on the 22 tracked boards are already in `_fmt_mm` minimal form and none
carries more than six decimals**. The example was invented. The sign toggle is
still right (exact *by construction* over the declared grammar, where the float
path is exact only because of a property of these particular files), so the
discriminating input is now synthesised rather than found.

## Deliberate divergences, disclosed

**pcbnew loses a nanometre on an arc's mid point when it flips; this does not.**
A plain `LoadBoard` + `SaveBoard` preserves rp2350 SW1's `(mid 1.40384 0)`
exactly, but `fp.Flip(...)` + `SaveBoard` writes `1.403839` — pcbnew re-derives
mid from centre/angles. This transform never touches an arc's X. PAR waives that
rather than matching it (adopting the rounding would make the transform lossy
and break RT's byte identity); the waiver is `mid` only, one nanometre only,
every instance printed, and capped.

**`docs/floorplan-intent.md`'s `zone_side` row is unchanged and still correct.**
It says the quench never flips a side, and the quench gains no lever here.

## What it refuses rather than guesses

`SideFlipUnsupported`, raised and never skipped — `write_placed_output` returns
True unconditionally, so a construct quietly passed through is indistinguishable
from success at every call site. Both the footprint-level and pad-level
dispatches are whitelists, because you cannot test for a node kind you have not
thought of.

Refusing is cheap, measured: of 1349 tracked footprint blocks the refusal set
touches at most 7 (`primitives` 6, `zone` 1). `chamfer`, `rect_delta`,
`fp_text_box` and `group` have **no witness at all**, which is the argument
*for* refusing them — but only because they describe SHAPE. Twelve passthrough
heads also have no witness and are passed through deliberately: they are
scalars, identifiers and flags a reflection cannot act on, and that distinction
is now written down rather than left as an apparent inconsistency.

## CLI / GUI

No new CLI flag — the capability rides on a dict key — so
`tests/gui_parity/test_manifest_plan_parity.py` and
`test_cli_postpass_coverage.py` are unchanged. Both were **run** rather than
assumed: the first reports OK on placement-block flags, the second 8 registered
passes / 0 unreachable / 0 failures.

One GUI path does change behaviour without changing: `placement_gui._apply_pose`
already compares `fp.IsFlipped()` against the result board's parsed layer and
calls `fp.Flip(...)`. That branch has been **dead for CLI-produced boards**
because the writer never wrote a differing layer. It is now live.

`.gui-parity-checked` is not touched — this is a placement-only change with no
routing reach, matching how PR #867 was handled.

## Provenance

`record_write` derived `moved` from x/y/rot only and wrote `poses_written` as
`[x, y, rot]`, so a flip that holds its pose landed in `refs_written` and **not**
in `refs_moved`, with a pose claim byte-identical to the incumbent — the ledger
recorded the flip nowhere and `fence_audit` would grade a flipped board as
untouched. Found independently by both design reviews, which is why it is in
this PR rather than a follow-up.

Fixed additively: a side term in the `moved` predicate and a separate
`sides_written` key. `SCHEMA` does **not** move — `read_ledger` is shape-agnostic,
`poses_written` keeps its `[x, y, rot]` shape, and the new key is optional and
absent from every existing row.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -413,6 +413,19 @@ same change — and vice versa. When adding a flag, grep the
 through there too.
 
 **Parity gates (run these when touching CLI/GUI routing):**
+- `tests/gui_parity/test_714_mirror_pcbnew_parity.py` — needs KiCad python; the
+  acceptance gate for the placement writer's **layer flip**. Compares the
+  footprint block `write_placed_output` emits for a side change against the
+  block pcbnew writes after `FOOTPRINT.Flip(pos, FLIP_DIRECTION_TOP_BOTTOM)` +
+  `SaveBoard()`, as canonical trees: numbers to integer nanometres (so the
+  tolerance is exactly zero, not a float epsilon), `at` angles folded, children
+  sorted, and the uuid multiset asserted rather than normalised away. **Exits 2
+  when pcbnew is absent and does NOT self-skip** — `tests/mutate_714.py` uses
+  it as a killer gate, and a killer gate that exits 0 reports every row it
+  guards as SURVIVED. Run it whenever you touch `placement/writer.py`'s flip
+  path or `kicad_parser.flip_layer_token`; ~2 min. Its wx-free siblings are
+  `tests/test_714_{mirror_discriminates,side_self_consistency,flip_roundtrip,
+  refusals,identity_write_unchanged}.py`, which `run_all.py` does collect.
 - `tests/gui_parity/test_manifest_plan_parity.py` — no wx; asserts every CLI
   `--flag` survives `manifest_to_plan` into the GUI plan step (plan→params).
 - `tests/gui_parity/test_cli_postpass_coverage.py` — no wx; asserts every CLI
--- a/docs/api-kicad-parser.md
+++ b/docs/api-kicad-parser.md
@@ -375,7 +375,7 @@ sided layer `F.<x>` / `B.<x>`, not the list of which ones exist. Measured over
 the 22 tracked boards, the layer tokens appearing inside a `(footprint ...)`
 block are `F.Cu F.Mask F.Paste F.SilkS F.CrtYd F.Fab F.Adhes B.SilkS B.Fab B.Cu
 B.CrtYd B.Mask B.Paste *.Cu *.Mask Dwgs.User Cmts.User Eco1.User Eco2.User
-User.1`; this flips the twelve sided ones and returns the other eight as they
+User.1`; this flips the thirteen sided ones and returns the other seven as they
 are.
 
 Two things it deliberately does **not** do, because both would be guesses:
@@ -383,13 +383,13 @@ Two things it deliberately does **not** do, because both would be guesses:
 - `*.Cu` and `*.Mask` pass through. They are KiCad's ALL-copper / all-mask sets
   and are already their own mirror; narrowing `*.Cu` to `B.Cu` would silently
   drop 66 pads on `rp2350_fpga_eensy_prePlane` U8 alone.
-- Inner copper passes through unchanged, which is the identity answer and not
-  the right one. `In1.Cu`'s mirror image is `In<n+1-k>.Cu`, which depends on
-  the board's inner-layer count *and* on `remove_unused_layers` padstack
-  semantics, and no tracked board carries an explicit `In<n>.Cu` in a footprint
-  pad. Callers that must decide about one are expected to **refuse** rather
-  than take this function's answer —
-  `placement.writer` does exactly that.
+- Inner copper passes through unchanged, and that matches KiCad: probed
+  against pcbnew 10.0.0 on a six-layer board, `FOOTPRINT::Flip` leaves a pad on
+  `In1.Cu` and an `fp_line` on `In2.Cu` where they are. So the identity answer
+  is correct here, not merely conservative. `placement.writer` nonetheless
+  **refuses** a pad naming an explicit `In<n>.Cu`, for a different reason: no
+  tracked board carries one, so nothing here would notice if a future KiCad
+  began remapping them.
 
 ## Utility functions
 
--- a/docs/api-kicad-parser.md
+++ b/docs/api-kicad-parser.md
@@ -356,6 +356,41 @@ Converts footprint-local pad coordinates to absolute board coordinates.
 KiCad's rotation convention means the angle is **negated** inside the
 standard rotation matrix — use this helper rather than rolling your own.
 
+There is **no mirror term**, and that is correct rather than an omission: KiCad
+stores a B-side footprint's children *pre-mirrored*, so a plain rotate and
+translate resolves them. A consumer that adds a mirror of its own for B-side
+parts double-applies it.
+
+## Board side: `flip_layer_token`
+
+```python
+flip_layer_token(name: str) -> str
+```
+
+The other face's spelling of a layer token (#714): `F.SilkS` → `B.SilkS`,
+`B.Cu` → `F.Cu`, and **anything else unchanged**.
+
+A prefix rule, deliberately not a table — the knowledge is that KiCad spells a
+sided layer `F.<x>` / `B.<x>`, not the list of which ones exist. Measured over
+the 22 tracked boards, the layer tokens appearing inside a `(footprint ...)`
+block are `F.Cu F.Mask F.Paste F.SilkS F.CrtYd F.Fab F.Adhes B.SilkS B.Fab B.Cu
+B.CrtYd B.Mask B.Paste *.Cu *.Mask Dwgs.User Cmts.User Eco1.User Eco2.User
+User.1`; this flips the twelve sided ones and returns the other eight as they
+are.
+
+Two things it deliberately does **not** do, because both would be guesses:
+
+- `*.Cu` and `*.Mask` pass through. They are KiCad's ALL-copper / all-mask sets
+  and are already their own mirror; narrowing `*.Cu` to `B.Cu` would silently
+  drop 66 pads on `rp2350_fpga_eensy_prePlane` U8 alone.
+- Inner copper passes through unchanged, which is the identity answer and not
+  the right one. `In1.Cu`'s mirror image is `In<n+1-k>.Cu`, which depends on
+  the board's inner-layer count *and* on `remove_unused_layers` padstack
+  semantics, and no tracked board carries an explicit `In<n>.Cu` in a footprint
+  pad. Callers that must decide about one are expected to **refuse** rather
+  than take this function's answer —
+  `placement.writer` does exactly that.
+
 ## Utility functions
 
 ```python
--- a/docs/placement-optimization.md
+++ b/docs/placement-optimization.md
@@ -345,10 +345,19 @@ precedent), which is simpler and may capture most of the value:
     wins; mixed-size swaps are usually illegal anyway, so restrict by
     footprint compatibility)
   - *side flip* (optional; mirrored courtyard): treated as a first-class move
-    in recent PCB literature. **Not implemented as a move.** Board side *is*
-    now modelled by the clearance/halo terms (#456): a part occupies its own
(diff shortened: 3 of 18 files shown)


edgehero:
Superseded by #873, which carries all ten commits of this branch unchanged and builds #837 (the assembly-side declaration) and #836 (its pre-registered measurement) on top.

Three loose ends this PR left open are fixed there:

* `tests/stress/qualify_subject.py` drew `rng.choice(P.KINDS)`, so adding `layer_flip` — whose applied dose is 0 by construction — spent 1 draw in 6 on a kind its material gate can never pass, silently depressing the land rate it exists to measure.
* `provenance_audit.poses()` compared `(x, y, rotation)`, so a part flipped in place read as untouched and an unclaimed flip could never be a VIOLATION. This PR fixed the ledger half and said so; that is the consumer half.
* `write_placed_output`'s `Args:` block never mentioned `new_side`.

Closing in favour of #873.

Issue #828, opened 2026-09-01
Routing: the wx-main-thread board save reports a timeout and a save failure as the same bare False


edgehero:
`py_router/ui_thread.save_board_on_ui_thread` returns a bare `False` for two facts that call for opposite responses:

```python
        wx.CallAfter(_save_and_signal)
        if not done.wait(timeout_s):          # SAVE_BOARD_UI_TIMEOUT_S = 120
            abandoned.set()
            print(f"  (#688 guard{label}: the wx main thread did not run the "
                  f"board save within {timeout_s}s -- skipping rather than "
                  f"blocking)")
            return False
    if 'err' in box:
        print(f"  (#688 guard{label}: board save failed: {box['err']})")
        return False
```

A 120 s expiry says **this machine, this moment** — the UI thread was not pumping, and the same call may well succeed next time. A `SaveBoard` exception says **this board** — it will fail again. Both return `False`, so no caller can branch: a timeout is retryable and a save failure is not, and nothing downstream can act on the difference.

**Each arm does print distinctly**, so a human reading a log can tell them apart. Verified by driving both paths against a stubbed wx/pcbnew:

```
PATH A (main thread never pumps): return False (bool)
PATH B (SaveBoard raises):        return False (bool)
ra == rb: True | same type: True
```

The defect is the **programmatic** channel, not the operator-facing one.

## Two callers, and the GUI one is silent

`py_router/kicad_parser.py:4934` — `_live_fill`, the GUI's `exact_fill_provider`:

```python
            if not _save688(tmp, _board, label=" live-fill"):
                raise RuntimeError("could not save the live board on the UI "
                                   "thread (#688 guard)")
```

One `RuntimeError`, one message, caught below and turned into one warning, then a fall-through to `live_fill_islands` — the in-process fill with clearances resolved at board-LOAD time, i.e. the #627 divergence.

`kicad_routing_plugin/swig_gui.py:3427` — `_stage_live_board`, reached via the alias `gui_utils.save_board_via_ui_thread`, and this is **the site `ui_thread.py`'s own module docstring names as the original #688 deadlock**:

```python
                    if not save_board_via_ui_thread(_p, _b):
                        try:
                            os.unlink(_p)
                        except OSError:
                            pass
                        return None
```

[...]

Issue #726, opened 2026-08-23
Placement: footprints sharing a reference collapse to one, and the placement writer moves them all


edgehero:
`PCBData.footprints` is a `Dict[str, Footprint]` keyed by reference. When a
board carries two footprints with the SAME reference, one silently wins and the
other disappears from every placement instrument — and `write_placed_output`,
which matches by reference in the file text, then moves **both**.

This is not exotic. Five of the in-repo boards have it, and on `watchy` the
duplicated references are **real test points**, not placeholders:

| board | duplicate refs | copies, at these poses | parser keeps | footprints lost |
|---|---|---|---|---|
| `watchy` | `TP4` | (93.92, 105.82), (86.73, 92.13) | (86.73, 92.13) | 2 |
| | `TP5` | (93.92, 108.36), (84.87, 92.87) | (84.87, 92.87) | |
| `glasgow_revC` | `REF**` | 7 copies, all distinct | (59.5, 107) | 6 |
| `orangecrab_ext_pll` | `G***` | 3 copies, all distinct | (172.887, 102.465) | 2 |
| `ulx3s` | `EMARD` | (125.58, 88.8), (147.58, 89.3) | (147.58, 89.3) | 1 |
| `esp_prog` | `Ref*` | (141.2, 95.9), (118.1, 99.9) | (118.1, 99.9) | 1 |

`watchy`: 86 footprint blocks in the file, `len(pcb.footprints) == 84`.

## The half that writes copper

Worse than an invisible part. `placement/writer.py:write_placed_output` finds
footprint blocks and matches `(property "Reference" "XX")`, so ONE placement
entry moves EVERY footprint sharing that reference. Measured on `esp_prog`,
passing a single placement for `Ref*` that nudges it 0.40 mm:

```
Modified 2 footprint positions
```

```
original: (at 141.2 95.9)              -> 1 occurrence
output:   (at 141.2 95.9)              -> 0 occurrences
output:   (at 117.705442 99.965760)    -> 2 occurrences
```

The second fiducial was **teleported ~23 mm** onto the first, and the two now
sit stacked at one point. Nothing in the run named it, nothing reported it.

## Why it matters for placement specifically

- The lost footprints are absent from `grade_pad_legality`, `grade_body_overlap`
  and the quench gate, so their copper is never censused for clearance and
  never charged — a clean report on a board that is not clean.
- `build_part_pads` / `LegalityContext` / `QuenchState.parts` are all keyed by
[...]

Issue #328, opened 2026-07-07
NPTH pads with a net are treated as copper barrels at several drill>0 sites


drandyhaas:
## Summary
Several code paths gate "this pad connects every copper layer" purely on `pad.drill > 0`, without excluding `pad_type == 'np_thru_hole'`. An NPTH pad has **no copper barrel** — its hole ties nothing — so a *net-carrying* NPTH pad (e.g. a mounting hole deliberately tied to GND in the schematic) would be treated as an all-layer connection point that doesn't exist.

Found by the #289/#315 generalization audit (the #289 fix in `_route_hybrid_leg` and the `get_stub_info` has_pad_via fix both exclude `np_thru_hole` explicitly; these sibling sites don't).

## Sites (as of e3cca97)
- `diff_pair_routing.py:244-249` — `_endpoint_launch_layer_indices`: `if pad.drill > 0: spanned |= _span_all()`
- `stub_layer_switching.py` — `swap_would_orphan_smd_pad`: `if pad.drill > 0: return False  # through-hole pad spans all layers`
- `route_planes.py:1074` — plane-seed `on_layer = pad.drill > 0 or ...`, and `:1313` TH-pad→plane tie
- `route_disconnected_planes.py:638` — `if pad_layer is None or pad.drill: continue`
- `plane_pad_tap.py:1033` (`if pad.drill > 0: continue`), same assumption at `:609` / `:684`

## Why it's benign today
The parser assigns NPTH pads `net_id = 0` in the common case, so they don't appear in `pads_by_net` for a real net. It takes a deliberately net-tied mechanical hole (KiCad allows it) to trigger any of these.

## Failure modes when it does trigger
- A plane pass "connects" GND to a mounting hole with no barrel → counted as connected copper that isn't there (phantom connectivity, real open).
- A diff-pair/leg endpoint offered launch layers it can't actually reach → routes that terminate on nothing.
- Repair/reconnect skips a pad as "plane-tied on every layer" when it's tied to nothing.

## Fix shape
Mirror the #289 exclusion at each site: treat `drill > 0` as all-layer copper only when `pad_type != 'np_thru_hole'` (`Pad.pad_type` is already parsed; NPTH pads have no copper even when `layers` lists `*.Cu` — see CLAUDE.md pad notes). Consider a shared helper (e.g. `pad_is_plated_through(pad)`) so future sites can't re-introduce the bare `drill > 0` idiom.

drandyhaas:
[...]

PR #873, opened 2026-09-05
#837/#836/#714: a board can declare which faces the fab populates, and the flip move is measured rather than built

### LINKED ISSUE #714: Placement: placement.writer cannot mirror a footprint to the other side, so a real layer flip is unavailable

### LINKED ISSUE #836: Placement: a part's side is fixed at construction, so no move the engine can make puts a component on the other face

### LINKED ISSUE #837: Placement: nothing declares whether a board is single- or double-sided, so the back-side area credit is granted to every board with nothing charged against it

edgehero:
## A board could not say which faces the fab will populate, so the back one was free

`options.grow_board` scores area utilisation on the **busier face**
(`options.py:232`). That is right for a board built with two reflow passes — a
part on B.Cu does not compete for F.Cu area, and the comment there records the
two shipping boards that regression failed — and wrong for a board built with
one, where every part has to fit the single face that gets populated.

Nothing could say which kind of board it was. `floorplan._TOP_LEVEL_KEYS` had
no assembly key, no `place_*.py` had a side flag, and `py_tools/check_assembly.py`
— despite the name — had no per-side census at all. So a placement could be
reported as fitting on the strength of area the fab will never use, and nothing
in the run disclosed the assumption.

Closes #714
Closes #837
Closes #836

This supersedes #869, which carried #714 alone. All ten of its commits are here
unchanged, plus three loose ends it left open.

## What ships

| | |
|---|---|
| **#714** | the writer that mirrors a footprint to the other face — #869 verbatim, plus three gaps found while building on it |
| **#837** | `assembly.sides` in the intent, `rule_assembly_side`, the per-side census in `check_assembly`, and `grow_board` charging the declared face |
| **#836** | its own pre-registered kill switch, run to the end. **Answer: do not build the flip move.** The evidence ships as a committed measurement, not a claim |

## The face is named, and `single` is refused

`"sides": "single"` is under-specified and the corpus says so: `ulx3s` is
back-dominant, 163 of its 226 pad-bearing parts, so a rule reading "single
implies F.Cu" would be wrong about most of a shipping board. The enum is
`F` / `B` / `both`, and `single` is refused **by reason** — "name the face" is
the whole correction.

## Two decisions measurement reversed

**`both` RUNS the rule rather than skipping it.** Skipping looks like the
honest "this cannot fire, so do not claim to have graded it" — but the intent
DECLARES the key, so a skip lands in the abstention channel as *declared and
ungraded*: `rules_skipped_arm 0 → 7`, `boards_clean_but_ungraded 0 → 2`. And
running it is not vacuous: "every part is on a declared face" is a true answer
to a real question, the way `zone_containment` is on a board whose parts are
all inside their zones.

**The rule reads the census's own ref list, not `ctx.parts`.** The two agree on
every tracked board and are not the same set by construction: `QuenchState`
admits a zero-pad footprint that draws a courtyard, and the pad-bearing census
excludes it. One such board and `check_assembly` would print a count
`check_floorplan` contradicts.

## The arithmetic is inert until someone declares

| declared | charged |
|---|---|
| nothing, or `both` | `max(F, B)` — today's arithmetic, exactly |
| `F` or `B` | `F + B` |

Measured over all 23 boards, **the undeclared path is bit-identical**:
`fits_by_area`, `utilisation` and `shortfall_mm2_at_least` unmoved on every
one, because the two bases agree exactly when the far face is empty and 15 of
the 22 tracked boards have zero back-side area. Declaring one face moves the 7
that do not and flips the verdict on the 2 whose sum exceeds their usable area:
`ulx3s` 0.9132 → 1.3609, `orangecrab_ext_pll` 0.9458 → 1.2950 (both at the
board's own floor; at a flat 0.2/0.5 orangecrab reads 0.9186 → 1.2531).

`busiest_side_area_mm2` keeps its name and its meaning on both bases. It stops
being the number `utilisation` divides, which `charged_area_mm2` now is.

## Three defects an adversarial review found in my own census

Each is a case where the obvious helper is the wrong one:

* **`drill > 0` is not the assembly question.** It is
  `footprint_has_through_pads`, whose own docstring says it answers
  OBSTRUCTION. `watchy` printed *5 through-hole parts … need wave or hand
  soldering*; SW1–SW4 are SMD switches with unplated 0.75 mm alignment posts.
  `pad_is_plated_through` says 1 — CLAUDE.md's own #328 rule.
* **A third solder bucket.** `flat_hierarchy`'s 6 NPTH mounting holes are
  soldered by no process at all. Counting them as SMD reported 1 reflow pass
  for a board whose 58 real parts are every one through-hole; the honest
  answer is 0.
* **The printed arithmetic did not close on 7 of 22 boards**, because only the
  BACK zero-pad blocks were listed. `interf_u` reported "24 pad-bearing of 25
  blocks" and explained nothing — its missing block is a *front* graphic.

A second review round found six more, of which two are worth naming here: the
`not_modelled` disclosure was nested where nothing prints it, and forcing the
new basis bool into the text digest **evicted `part_area_mm2` from 21 of 22
boards**. It also caught a sentence of mine that was simply false — see
*Corrections* below.

## One correction to #837's own acceptance criterion

It asks for **94** violations on `glasgow_revC` declared `F`. The measured
answer is **92**, and 94 is not reachable under any single basis: 94 is the
back-side *block* count, and two of those blocks carry no pads. Every per-side
number for this corpus is ambiguous without its counting rule — #726 moved
`ulx3s` from 234 blocks to 235 and glasgow from 266 to 272 — so the census
declares its basis in both channels, and the issue's number wants the
correction rather than the code.

## #836: measured, and not built

#836 pre-registered its own kill switch: *"on at least three corpus boards, how
many parts does a flip help that a re-seat does not? If the answer is near
zero, this issue should be closed rather than built."*

`tests/measure_836_flip_vs_reseat.py` is that check. Its thresholds were
committed in `484ae103`, **before the results existed**, so git makes the order
checkable.

**The routed arbiter says the flips do not pay.** Of the 8 parts the screen
called EXCLUSIVE — helped by a flip and by no same-side move within reach — a
real scoped route gives **1 better, 4 WORSE, 3 unmoved**, every one on the last
rung (vias). `E_material`, the count improving any rung above it, is **0**.
That is not what the instrument always says: of 8 CONTROL parts flipped and
routed identically, **6 moved the ladder**.

**#836's premise about the objective is backwards, and that is the
load-bearing correction.** The issue says *"the objective cannot price the move
— the only term that responds to a flip at all is 2-D MST length."* Measured:
`_halo_pair_penalty` returns `0.0` for a cross-side SMD pair
(`quench.py:1344`), `candidate_valid`'s pair loop skips cross-side pairs
entirely (`:1798`), `_body_contained_at` skips them (`:2122`), `pair_min_gap`
returns `None` for an unshared side (`:698`), and `align` has no side filter at
all (`:1117`).

**207 of ulx3s's 226 movable parts and 237 of glasgow_revC's 243 have their
whole same-side halo charge zeroed by a flip.** A greedy quench handed this
move and this objective would flip nearly every part on the board, every run.
The risk was never that the search could not see the move; it is that it could
only see the upside. Hence the verdict is a route, and the cost delta is
published beside it as the artefact it is.

**What the flip was compared against.** Not `reseat.py` as shipped:
`clusters_from_tethers` sources only decap tethers and reaches **23% of ulx3s
and 35% of glasgow_revC**, so comparing against it alone would have
manufactured a win on two thirds of each board. The move arm is the union of
native reseat and `pose_score.rank_poses`.

**Said plainly:** the pre-registered CLOSE arm asked for `E ≤ 1` and E is 8, so
this conclusion rests on the routed arm the design named as decisive rather
than on the screen count. Both numbers are in the baseline so a reader can
disagree with the reading rather than with the data. Four things are recorded
as NOT measured, including `orangecrab_ext_pll`, which the rig **refuses** to
route because it ships 742 segments a flip would strand — a routed A/B there
measures copper damage, not placement.

`tests/test_836_flip_census.py` re-derives both mechanisms in five seconds and
fails naming #836 if an engine change ever makes a flip cost something. That
test is the reopening signal, and it is a test rather than a sentence in a doc.

## Corrections

I published a mechanism that was false and a review caught it. The `both`
decision was justified in the commit message *and in the code comment* by
"glasgow_revC and orangecrab_ext_pll went `pass: true → false`". A re-run at
the base commit, with none of this code present, showed those boards already
failing: the committed census baseline had gone stale four days earlier for
unrelated reasons and I read that drift as my own effect. The decision stands;
the comment now carries the measured mechanism and says what it used to say.

## Verification

* full suite **547 passed / 0 failed** at this HEAD and **545 / 0** at #869's
  head, same machine, same router (0.22.0) — the +2 are the two new test files
* `tests/mutate_837.py`: **26 rows, 24 killed, 2 disclosed survivors, 0
  broken**; `--verify-anchors` 26/26; `--selftest` proves the `.pyc` defence
  rather than asserting it
* the change detectors that legitimately moved were **re-stated, never
  re-derived** from the code they guard: `rules_skipped 11 → 12`, `KEY_SETS`,
  `_ASSEMBLY_KEYS`, the `713_abstention_census.json` re-record,
  `test_712_edge_centering`'s `READER_VERSION` literal and `mutate_711`'s
  anchor on it
* `tests/gui_parity/test_manifest_plan_parity.py` and
  `test_cli_postpass_coverage.py` both run, both unchanged

## Disclosure: `.gui-parity-checked`

This branch touches a `PARITY_SURFACE` — `py_router/kicad_parser.py` gains
#714's pure `flip_layer_token` helper — so the release gate is tripped by path
prefix. The marker is already 61 commits behind for unrelated reasons (#530,
#860, #862, #828), so this is not the PR to move it. Flagged rather than
repeating #869's "placement-only, no routing reach", which does not survive the
path-prefix rule.


edgehero:
Four things this work turned up that are **not** fixed here, filed separately rather than smuggled in:

* **#876** — `tests/test_run6_check_assembly.py` calls `unittest.main()` at line 72 with two test classes defined after it, so 3 of its 7 tests never run and it reports `OK (skipped=3)`. Effective CI coverage of this file is one test. Found while looking for the gate that would catch a mistake in the census this PR adds.
* **#877** — 24 mutation-battery anchors across 8 of 37 batteries match nothing, so their rows mutate nothing and report KILLED. Only 5 batteries have `--verify-anchors`. Found by re-anchoring the `mutate_711.py` row that this PR's `READER_VERSION` bump staled, and wondering how many others already were.
* **#878** — `grow_board` charges a through-hole part's area to one face; `check_pockets` charges the whole courtyard to both and `rule_keepout` charges the drilled-pad rect to the far one. Three instruments, three answers. This PR **defers** that deliberately and discloses it in `not_modelled` — folding it in would have moved all 17 through-hole boards in the same commit as the basis change and destroyed attribution. Measured impact on today's default basis: 2 of 23 boards, no verdict flips.
* **#879** — `713_abstention_census.json` has no gate keeping it fresh, and this PR is the worked example of the cost: I re-recorded a four-day-stale baseline and cited the resulting delta as my own result, in a commit message and a code comment. Corrected here; the missing gate is filed there.

Deliberately not filed, for the record: `render_placement.py` and `fanout_clearance.py` each re-derive the side rule from the raw layer string instead of calling `legality.footprint_side`. They agree with it on every tracked board, so it is latent duplication rather than a live bug — noted inside #878, which is the pass that would touch the same code.

PR #883, opened 2026-09-05
#882: a POSIX-only import at module scope made the suite permanently red on Windows

### LINKED ISSUE #882: tests: fill_timing_census imports POSIX-only 'resource' at module scope, so test_831_fill_preflight_census fails at import on every Windows machine
**Found while:** running the full suite on a Windows machine for an unrelated PR, and checking whether the failure was mine.
**Verified against:** a clean `upstream/main` worktree @ `4b487455`.

## The defect

`tests/stress/fill_timing_census.py:44` imports `resource` at module scope. `resource` is **POSIX-only** — it does not exist on Windows — so the module cannot be imported there at all, and `tests/test_831_fill_preflight_census.py` dies before it runs a single assertion:

```
Traceback (most recent call last):
  File "tests\test_831_fill_preflight_census.py", line 23, in <module>
    import fill_timing_census as ftc                    # noqa: E402
  File "tests\stress\fill_timing_census.py", line 44, in <module>
    import resource
ModuleNotFoundError: No module named 'resource'
```

`test_831_fill_preflight_census.py` is collected by `run_all.py`, so **the suite has a standing failure on every Windows machine**. Arrived with `e2643179` (#831, PR #868), 2026-09-04.

## Why it is a one-line class of fix

`resource` is used in exactly **one** function:

```python
# tests/stress/fill_timing_census.py:103
def _child_cpu():
    ru = resource.getrusage(resource.RUSAGE_CHILDREN)
    return ru.ru_utime + ru.ru_stime, ru.ru_maxrss
```

whose only callers are `measure_one` (`:121`, `:125`) — the census-**running** path. Everything `test_831_fill_preflight_census.py` touches is `ftc.analyse(...)`, a pure function over already-recorded rows that never reaches it. So the platform dependency is real for *running* a census and entirely spurious for *analysing* one, which is all the test does.

## The repo already has the pattern, twice

This is the only unguarded POSIX-only import in the tree — I swept for `resource`, `fcntl`, `termios`, `pwd`, `grp` and `posix`:

```
tests/stress/fill_timing_census.py:44:import resource        <-- unguarded, top level
py_router/memory_debug.py:12:    import resource            <-- guarded
tests/stress/redo_stress_test.py:805:        import resource as _res   <-- local to the caller
```

`py_router/memory_debug.py` does the try/except with a `_HAS_RESOURCE` flag; `redo_stress_test.py` imports it inside the function that needs it. Either shape fixes this. So the ask is not "add Windows support" — it is "one file departs from a convention the repo already follows in both directions".

## Suggested fix

Move the import into `_child_cpu` (the `redo_stress_test.py` shape), or guard it at the top (the `memory_debug.py` shape) and have `_child_cpu` return a disclosed `None` on a platform without it, so a census run on Windows reports "CPU not measured here" rather than a wrong zero. The second is slightly better: `measure_one` writes `cpu_s` into the recorded rows, and a silent 0.0 there would be a number someone later compares against a real one.

Whichever, the recorded rows should say which platform produced them — `_machine()` (`:141`) already captures `platform.platform()`, so a reader can tell.

## Verification

`python3 tests/test_831_fill_preflight_census.py` must pass on Windows, and `python3 tests/run_all.py` must stop carrying it as a failure there. On Linux/macOS nothing should change: `_child_cpu` still returns real numbers and `run_census` still records them.

## Why this is worth fixing ahead of larger things

A permanently-red test on a platform the project is developed on is worse than its own size. It trains the reader to skim the failure list, which is where the next real failure goes unnoticed — the same reasoning as #876, where 30 tests reported OK without running.

edgehero:
## The suite was permanently red on Windows, at import, before a single assertion

`tests/stress/fill_timing_census.py` imported `resource` — a POSIX-only
standard-library module — at module scope. Windows has no such module, so the
file could not be imported there at all, and
`tests/test_831_fill_preflight_census.py` died taking it with it:

```
  File "tests\test_831_fill_preflight_census.py", line 23, in <module>
    import fill_timing_census as ftc
  File "tests\stress\fill_timing_census.py", line 44, in <module>
    import resource
ModuleNotFoundError: No module named 'resource'
```

That test is collected by `run_all.py`. The expensive part is not the one test:
a permanent red teaches the reader to skim the failure list, which is where the
next real failure goes unnoticed.

Closes #882

## The dependency was spurious for what the test does

`resource` is used in exactly one function — `_child_cpu` — reached only from
`measure_one`, the census-**running** path. Everything the failing test touches
is `analyse()`, a pure function over already-recorded rows that separates on
`fill_elapsed_s` / `timeout_s` and reads neither CPU field. Real for *recording*
a census; entirely spurious for *analysing* one.

## Guarded, and the absence is a value rather than a zero

`_HAS_RESOURCE` with a try/except — the shape `py_router/memory_debug.py`
already uses for this same module.

`_child_cpu()` returns `(None, None)` where it is unavailable, **not `0.0`**.
A zero would be written into the recorded rows as `child_cpu_s: 0.0`,
indistinguishable from a fill that genuinely cost nothing — and those rows get
compared against ones recorded on a POSIX machine. The row also carries
`cpu_source`, so it states *not measured: no `resource` module on Windows*
rather than leaving a reader to infer it from a null.

Two printers would then have raised on the `None` — `run_census`'s progress
line and `report()`'s `{r['child_cpu_s']:7.1f}` — and print `-` instead, so an
unmeasured cost cannot be read as a small one.

## The gate, because a sweep cannot prevent the next one

`test_718_static_test_hygiene.py` says exactly this in its own docstring:
*"A sweep cannot prevent that — only a standing gate can."*

`test_no_module_scope_posix_only_import` walks **967 files** — the whole repo,
not just `tests/`, because the same import in `py_router/` would break the
router itself on Windows rather than one test — and refuses a top-level import
of `resource`, `fcntl`, `termios`, `pwd`, `grp`, `posix`, `crypt`, `syslog`,
`nis` or `spwd`. An import inside a function or a try/except is the correct
shape and is not flagged; this was the only unguarded instance in the tree.

## Both arms mutation-checked, and the second one exists because I got it wrong

| mutation | result |
|---|---|
| restore the unguarded `import resource` | gate fails, naming `tests/stress/fill_timing_census.py:56: resource` |
| append a syntax error to that file | gate fails, naming it **UNPARSEABLE** |

The second arm is there because the first draft was wrong in a way worth
recording. My initial mutation happened to produce a `SyntaxError`, and the
gate printed **PASS over one file fewer** — silently skipping the very file I
had just broken, because it swallowed the parse error and moved on. `_code_only`
in the same file already states the rule: *a gate that goes quiet on a file it
cannot parse is worse than one that is slightly noisy.* The check now refuses
to clear what it could not read.

## Verified

* `tests/test_831_fill_preflight_census.py` — was dying at import on Windows,
  now **15/15 checks passed**
* `tests/test_718_static_test_hygiene.py` — passes, 5 registered checks
* Full suite on this branch is RUNNING as this opens; I will post the result
  as a comment rather than quote a number I do not yet have. The last full run
  on `main` from this machine was 545 passed / 0 failed, and the run that
  exposed this bug carried it as one of two failures.
* Nothing changes on POSIX: `_child_cpu` still returns real numbers,
  `run_census` still records them, and `analyse` never read them either way.

## One thing noticed and deliberately NOT fixed

`measure_one` computes `child_maxrss_mb` as `maxrss / (1024 * 1024)`.
`ru_maxrss` is in **bytes on macOS and kilobytes on Linux**, so that line is
correct on the platform this is developed on and off by 1024× on Linux.
Out of scope here — correcting it changes the meaning of already-recorded rows
and wants its own re-measurement — but flagged rather than silently left, and I
am happy to file it.

### DIFF
--- a/tests/stress/fill_timing_census.py
+++ b/tests/stress/fill_timing_census.py
@@ -41,9 +41,24 @@ import json
 import multiprocessing as mp
 import os
 import platform
-import resource
 import sys
 import time
+
+#: `resource` is POSIX-only. Imported at module scope it made this file
+#: unimportable on Windows, which took `tests/test_831_fill_preflight_census.py`
+#: -- a collected test that only calls the pure `analyse()` -- down with it at
+#: IMPORT time, so the suite carried a standing failure on every Windows
+#: machine (#882).
+#:
+#: Guarded rather than moved into `_child_cpu`, so the absence is a value this
+#: module can report rather than an exception at the call site. Same shape as
+#: `py_router/memory_debug.py`, which already does this for the same module.
+try:
+    import resource
+    _HAS_RESOURCE = True
+except ImportError:                                        # pragma: no cover
+    resource = None
+    _HAS_RESOURCE = False
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 REPO = os.path.dirname(os.path.dirname(HERE))
@@ -101,6 +116,16 @@ def board_features(board_file):
 
 
 def _child_cpu():
+    """(cpu seconds, peak RSS) charged to reaped children, or (None, None).
+
+    None rather than 0.0 where `resource` is absent. A zero here would be
+    written into the recorded rows as `child_cpu_s: 0.0`, indistinguishable
+    from a fill that genuinely cost nothing -- and those rows are compared
+    against ones recorded on a POSIX machine. An absent measurement has to
+    look absent.
+    """
+    if not _HAS_RESOURCE:                                  # pragma: no cover
+        return None, None
     ru = resource.getrusage(resource.RUSAGE_CHILDREN)
     return ru.ru_utime + ru.ru_stime, ru.ru_maxrss
 
@@ -129,8 +154,18 @@ def measure_one(args):
         'wall_s': round(wall, 2),
         'fill_elapsed_s': (round(st.elapsed_s, 2)
                            if st.elapsed_s is not None else None),
-        'child_cpu_s': round(cpu1 - cpu0, 2),
-        'child_maxrss_mb': round(maxrss / (1024 * 1024), 1),
+        # None all the way through where the platform cannot measure it, and
+        # SAID so in the row rather than left for a reader to infer from a
+        # null. `analyse()` reads neither field -- it separates on
+        # `fill_elapsed_s` / `timeout_s` -- so a census recorded without them
+        # still answers the question this module exists for.
+        'child_cpu_s': (round(cpu1 - cpu0, 2)
+                        if cpu0 is not None and cpu1 is not None else None),
+        'child_maxrss_mb': (round(maxrss / (1024 * 1024), 1)
+                            if maxrss is not None else None),
+        'cpu_source': ('resource.RUSAGE_CHILDREN' if _HAS_RESOURCE else
+                       f'not measured: no `resource` module on '
+                       f'{platform.system()}'),
         'n_islands': (sum(len(v) for v in islands.values())
                       if islands else 0),
         'timeout_s': timeout,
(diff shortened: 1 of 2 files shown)


edgehero:
Full suite on this branch: **552 passed, 0 failed, 0 timed out** (3604 s), plus the one pre-existing self-skip (`test_run8_starved_face_gate.py`, which needs gitignored `wk/` boards).

That is the number the PR body deferred. For comparison on the same machine:

| tree | result |
|---|---|
| `main` before this branch | 545 passed / **1 failed** — this bug |
| this branch | **552 passed / 0 failed** |

The +7 is not from this PR: `main` moved under me while this was in flight (#873 merged, bringing its two new test files, and PR #881's dead-test work is a separate branch). What matters here is the failure column going to zero and no test regressing.

The two long tests that timed out in an earlier run (`test_431_board_gates.py`, `test_interf_u.py`) both completed inside their budgets this time — they were contending with a routing measurement I had running concurrently, which is what `run_all`'s own note about timeouts warns against reading as evidence.

Issue #831, opened 2026-09-01
Routing: exact fill needs a deterministic pre-flight predicate, so which geometry a board gets stops depending on CPU speed


edgehero:
#713 item 4 asks for two things. PR #827 did the first — `refill_islands_ex` returns a discriminated `RefillStatus`, so a 300 s `EXACT_FILL_TIMEOUT` is no longer *reported* as "the KiCad refill failed", a fact about the board. It deliberately did **not** do the second:

> *A deterministic pre-flight predicate on board size would make the path taken a function of the board alone.*

## The headline, verified

Still true after #827. `plane_fragility.py` consumes the new status **for the printed message only**, and then falls through to the drawn outlines regardless of reason:

```python
                print(f"Plane fragility: exact fill unavailable ({_st.why()}); "
                      f"using zone outlines")
                fills = {}
    ...
    if not polys:
        polys = [(z.net_id, z.layer, z.polygon) for z in pcb_data.zones]
```

There is no `is_timeout` branch anywhere in `plane_fragility.py`; the same shape holds on the GUI live-fill path in `kicad_parser._live_fill`. The two geometries are not interchangeable — the module docstring says so: *"For a full-board outline the fallback prices only a board-edge band -- near-useless; the exact fill is the real lever."* So **which geometry the router builds against still depends on whether pcbnew finished in 300 s on this machine.** Naming the cause made that visible; it did not make it deterministic.

## Corrections to my first version of this issue

Three of the "building blocks" I cited were wrong, and one was load-bearing. Correcting them here rather than leaving them to mislead whoever picks this up:

- **`_fill_cost_key` is not "used only as a memo key, never as a gate".** `kicad_oracle.py:304-307` gates on it — a call-scoped short-circuit added by #827 itself, which returns without spawning anything. The defensible phrasing is *never as a pre-flight gate on the **fill** path*. Its shape is also a **5-tuple with a leading `'fill'` discriminator**, and it has a **second shape entirely** — `('path', realpath)` on any parse failure, which is not a board signature at all and would have to be handled.
- **The `W * H > 40_000_000` guard is not "a size predicate of exactly the right shape three lines below".** It is 35 lines below, `W` and `H` are **grid cells** (`round(x * inv_step)` at a 0.1 mm default step), and it caps one zone polygon's raster at ~6325×6325 *inside the loop that runs after the geometry source has already been chosen*. It is a numpy memory guard, not a board-complexity proxy.
[...]

PR #888, opened 2026-09-07
--json-out accounts for the two runs it stayed silent about

pentolope:
A run that legitimately had nothing to do now writes its `--json-out` file.

**Scope note after merging main:** this PR originally carried a second commit
adding a `plane_fragility` account to the summary. Upstream landed the same
disclosure independently as #831 — better shaped (return-threaded `_geometry`
record, the `machine_dependent` flag mine lacked) — so the merge commit yields
that entire commit to upstream's implementation and its
`test_831_fill_geometry_disclosure`. What remains is the no-op fix below,
which #831 does not touch.

---

## The defect

`_emit_summary_min`'s own docstring states the contract: a step that did
nothing *"must still satisfy it. A missing line there is indistinguishable
from a crash, which is the one reading that would make an agent do the wrong
thing."* The **console** line honors that. The **file** did not: both
nothing-to-do early returns (`already_connected` and `no_valid_nets`) fire
before the end-of-run `if json_out:` block, so a caller that asked for
`--json-out` got *"All nets are already fully connected"*, exit 0, and no
file.

Measured impact: an external wrapper that stages a board and reads the file
as the run's account refused already-routed boards as "unaccounted" — three
attempts, three refusals, on two different boards — locking every
already-routed board out of a re-record, because a re-record's staged source
is by definition a board with nothing left to route.

## The fix

`_write_summary_min_file(json_out, status)`, the file twin of
`_emit_summary_min`, called at both early returns:

```json
{"successful": 0, "failed": 0, "status": "already_connected",
 "env_knobs": {...}}
```

Deliberately **no** `min_clearance_used`: a run that routed nothing applied
no clearance, and inventing a number would defeat a reader's floor check.
Reconciliation sub-runs cannot clobber the parent's file — `json_out` is
already in the stripped `_reconcile_kwargs` set — and the write goes through
`write_summary_file`'s all-or-nothing publish.

## Compatibility

The one behavioral delta is that a file now exists where none did. Readers
were checked:

- `py_placer/converge.route_verdict` scores it `(None, 'unreadable summary')`
  — a non-verdict, exactly as the no-file path scored — because the document
  carries none of the five verdict keys and the key-presence guard (the
  documented schema-drift tripwire) refuses to fabricate a clean 0. A
  `no_valid_nets` run cannot be mistaken for a clean route by the placement
  judge.
- The early returns emit exactly one `JSON_SUMMARY_MIN` and one file with the
  same tally.
- Wrappers that treated a missing file as "unaccounted" can now accept the
  no-op deliberately, and still refuse a contradictory account
  (`already_connected` with a nonzero failure tally).

## Testing

- `tests/test_json_out_summary.py::test_a_nothing_to_do_run_still_writes_the_file`
  — the tally, the status, the env echo, the deliberate absence of
  `min_clearance_used`, `None` as a no-op, and (source-pinned, like the
  file's existing json_out-pop guard) that **both** early returns call the
  writer.
- Post-merge with `grid_router` v0.22.0: `test_json_out_summary` ALL PASS
  including the end-to-end file-equals-merged-stdout route;
  `test_831_fill_geometry_disclosure` 23/23; `test_713_refill_status` 61/61.
- Downstream, the external wrapper's full suite consuming these summaries
  runs green, including an end-to-end accounted no-op that reaches adoption.

### DIFF
--- a/py_router/route.py
+++ b/py_router/route.py
@@ -340,6 +340,40 @@ def _emit_summary_min(gate_report: Optional[dict] = None,
         print(f"  WARNING: could not emit JSON_SUMMARY_MIN: {_e}")
 
 
+def _write_summary_min_file(json_out: Optional[str], status: str) -> None:
+    """Write the --json-out file for a run that legitimately did nothing.
+
+    The console contract above ("exactly one JSON_SUMMARY_MIN per outermost
+    run") has a file twin: a caller that asked for --json-out reads the FILE,
+    and an early return that prints the tally but skips the write leaves that
+    caller unable to tell "nothing to do" from a crash. That is not
+    hypothetical: a wrapper staging an already-routed board got "All nets are
+    already fully connected", exit 0, and no file - and refused the run as
+    unaccounted, three times over, on two different boards.
+
+    The document carries the same empty tally the console line prints, the
+    `status` naming WHY it is empty, and the env-knob echo the normal
+    end-of-run summary carries. Deliberately NO min_clearance_used: a run
+    that routed nothing applied no clearance, and inventing a number here
+    would defeat a reader's floor check.
+    """
+    if not json_out:
+        return
+    try:
+        from route_summary import write_summary_file
+        document = {'successful': 0, 'failed': 0, 'status': status}
+        try:
+            import env_knobs as _ek653
+            document['env_knobs'] = _ek653.active_env_knobs()
+        except Exception:                                       # noqa: BLE001
+            pass
+        write_summary_file(json_out, document)
+        print(f"  route summary written to {json_out}")
+    except Exception as _e:                                     # noqa: BLE001
+        print(f"  WARNING: could not write --json-out {json_out}: "
+              f"{type(_e).__name__}: {_e}")
+
+
 def _late_orphan_sweep659(pcb_data, output_file, return_results, results_data,
                           protect_unfinished, keep_input_copper, skip_routing):
     """Sweep pad-less copper islands off the FINAL board (#659).
@@ -1332,6 +1366,7 @@ def batch_route(input_file: str, output_file: str, net_names: List[str],
         print("No valid nets to route!")
         if final_reconcile:
             _emit_summary_min(status='no_valid_nets')
+        _write_summary_min_file(json_out, 'no_valid_nets')
         if return_results:
             return 0, 0, 0.0, _empty_results_data()
         _write_passthrough_output(input_file, output_file)
@@ -1440,6 +1475,7 @@ def batch_route(input_file: str, output_file: str, net_names: List[str],
         print("All nets are already fully connected - nothing to route!")
         if final_reconcile:
             _emit_summary_min(status='already_connected')
+        _write_summary_min_file(json_out, 'already_connected')
         # The sweep runs HERE too (#659). The fragment gate diverts a net whose
         # extra fragments are all pad-less to this sweep instead of the router,
         # so a step whose WHOLE scope is diverted lands on this early return --
--- a/tests/test_json_out_summary.py
+++ b/tests/test_json_out_summary.py
@@ -153,6 +153,39 @@ def test_end_to_end_file_equals_merged_stdout():
     print("  PASS: --json-out == merged(stdout) on a real run")
 
 
+def test_a_nothing_to_do_run_still_writes_the_file():
+    """The early returns' file twin of the JSON_SUMMARY_MIN contract.
+
+    "All nets are already fully connected" printed the console tally and
+    wrote NO --json-out file, so a file reader could not tell the routine
+    nothing-to-do state from a crash -- and external wrappers refused the
+    run as unaccounted, which locked every already-routed board out of a
+    re-record. The file carries the same empty tally, the status naming
+    why it is empty, the env-knob echo, and deliberately NO
+    min_clearance_used: a run that routed nothing applied no clearance,
+    and inventing one would defeat a reader's floor check.
+    """
+    import route
+    with tempfile.TemporaryDirectory() as td:
+        js = os.path.join(td, 'summary.json')
+        route._write_summary_min_file(js, 'already_connected')
+        doc = json.load(open(js, encoding='utf-8'))
+        assert doc['status'] == 'already_connected'
+        assert doc['successful'] == 0 and doc['failed'] == 0
+        assert 'min_clearance_used' not in doc, (
+            "a run that routed nothing must not claim an applied clearance")
+        assert isinstance(doc.get('env_knobs'), dict), (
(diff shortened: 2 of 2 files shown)


drandyhaas:
Thanks for this — I reviewed both commits against current `main` (`43cc9266`), and they grade oppositely. Taking you up on your offer to split: **commit 1 we want; commit 2 has been overtaken by work that landed after your base.**

Line numbers below are as of `43cc9266`.

## Commit 1 (`b3f960be`) — confirmed, wanted, merge it

The defect reproduces exactly as described. The `--json-out` write sits at `route.py:5779`, and both nothing-to-do early returns (`no_valid_nets` at 1338, `already_connected` at 1458) return before it. Your framing is the right one: `_emit_summary_min`'s own docstring says a missing account is indistinguishable from a crash, and the file half wasn't honouring what the console half promised.

Everything I checked holds:

- **Applies cleanly to current `main`** despite the stale base — cherry-picked with no conflict.
- **`tests/test_json_out_summary.py` is 7/7**, including `test_end_to_end_file_equals_merged_stdout` on a real routed board.
- **The `route_verdict` compatibility argument is correct.** `py_placer/converge.py:216` requires one of five verdict keys before it will score anything; your document carries none of them, so it lands on `(None, 'unreadable summary')` — a non-verdict, precisely as the no-file path scored. A `no_valid_nets` run can't be laundered into a clean route by the placement judge.
- **No sub-run clobbering**, as you say — `json_out` is in the stripped `_reconcile_kwargs` set at `route.py:660`, which is also why calling the writer unconditionally is safe where `_emit_summary_min` is gated on `final_reconcile`.
- Omitting `min_clearance_used` is the right call for the reason you give.

One note, not a change request: the third early return at `route.py:723` also skips the file, but that's the `KICAD_DUMP_BATCH_KWARGS` debug escape hatch rather than a legitimate no-op run, so leaving it uncovered is right.

## Commit 2 (`94249b16`) — please drop it

This one is on us for timing. `main` already carries this feature: **#831, commit `e2643179` (2026-09-04)** — three days after your base `3fb9c05f`, so you were working blind to it. It emits the same `summary['plane_fragility']` key at `route.py:3796`, sourced from `config._plane_fragility_geometry`, covering the same four arms of the same decision.

Two reasons to prefer what's already there:

**It carries more.** `{source, islands, refill_status, why, machine_dependent}` against your `{mode, refill_reason, islands}`. `refill_status` is the same #713 vocabulary your `refill_reason` rides on, so that part is a wash — but `machine_dependent` (set from `_st.is_timeout`) is the field #831 exists for: it separates "priced on outlines because this board *has* no fill" from "because pcbnew didn't finish in 300 s *on this machine*". Given your fill-instability findings, that's the distinction your wrapper most wants to pin on, and it's a strictly stronger thing to refuse by than `mode`.

**Per-config beats module-global.** `_FRAGILITY_ACCOUNT` goes stale across runs in a long-lived process (the GUI routes many boards in one interpreter), and it defaults to `mode: "disabled"` — an assertion that `KICAD_PLANE_FRAGILITY_COST` was pinned off — for any run where the field never executed. That's a false statement in the exact channel the commit exists to make trustworthy.

**And the merge is quietly lossy, which is the part worth flagging.** I trial-merged it: `plane_fragility.py` conflicts in 4 hunks, as you'd expect — but **`route.py` merges cleanly** and ends up writing `summary['plane_fragility']` twice, at 3796 (#831's record) and then 4078 (yours). The second wins, so the merged tree silently downgrades #831's disclosure and drops `machine_dependent`. Nothing in the conflict markers would show a reviewer that happened.

## What we'd like

Commit 1 alone, rebased onto current `main`. Happy to take it as this PR with commit 2 dropped, or as a fresh one — whichever is less work for you.

Separately: the future-work item you held back — snapping fill vertices to `config.grid_step` before rasterizing, collapsing the divergent-polygon difference from 21 cells to 0 — is the most interesting thing in this PR, and you were right to keep it out of it. If you want to pursue that, it wants its own PR with a corpus A/B; note the house rule that a plane/oracle chain change can't be judged by a single replay pair, since that chain's own run-to-run spread is a few nets either way. Thanks also for the upstream `ZONE_FILLER` report — that's useful to us regardless.

pentolope:
Done as requested: rebased onto 43cc9266, commit 2 dropped in favour of #831's disclosure

drandyhaas:
Merged to `main` as `3c7febed`. Thanks for turning the split around so quickly — and for the original find, which was a real gap in a contract the code states about itself.

Verified before merging, beyond re-running your test:

- The rebased head is **byte-identical** to the `b3f960be` I reviewed, and carries no residue of the dropped commit.
- **Live negative control**, since the shipped test calls `_write_summary_min_file` directly and pins the call sites by source text — it would pass even if those sites were unreachable. On a genuinely already-routed board: unpatched `main` prints "All nets are already fully connected", exits 0, writes no file; with the fix the file appears carrying exactly `{"successful": 0, "failed": 0, "status": "already_connected", "env_knobs": {}}`, no invented `min_clearance_used`, output board still written. If you want to harden that test later, driving `batch_route` to the early return would close the gap.
- That **real** file feeds `converge.route_verdict` as `(None, 'unreadable summary')` — a non-verdict, exactly as the no-file path scored `(None, 'no summary')`. Your compatibility argument holds against the actual artifact, not just the schema.
- Summary/verdict readers all green: `test_json_out_summary`, `test_open_single_verdict`, `test_830_probe_summary_guard`, `test_713_probe_verdicts` (26/26), `test_run23_summary_min`, `test_run8_verdict_plumbing`, `test_run9_verdict`, `test_rip_restore_verdict`, `test_protected_skipped_summary`, `test_converge`.

One note for your wrapper: a no-op run now **overwrites** a `--json-out` file an earlier chain step wrote at the same path. That is strictly better than the old behaviour — which left the previous step's stale file to be misread as this run's account — but it does mean last-write-wins if you point two steps at one path.

The grid-snap idea from your future-work section still stands as the most interesting thing here if you want to pick it up; it wants a corpus A/B rather than a single replay pair, for the reason noted earlier.

pentolope:
@drandyhaas

On grid-snap: I can pick it up. Understood on method — corpus A/B rather than
a replay pair, with any effect needing to clear the plane/oracle chain's own
few-nets-either-way spread.

Two things I can't settle alone:

- Boards. My corpus is 8 low-complexity boards with no BGAs — so fine-pitch
  fanout, the case grid-snap most plausibly moves, is exactly what I can't
  measure. If you have boards you'd want this exercised on, I'll run them.
- Acceptance bar. Past clearing that spread, what would you want to see:
  routed length, via count, completion rate, wall time?

I can have the A/B on my own corpus late this week and will report it
either way — if it comes out neutral or worse, that's the report. Anything
you send me goes into the same run.

Issue #803, opened 2026-08-30
placement: eleven driver sites cite wk/calibration/RESULT.md as authority, and that file does not exist


edgehero:
Eleven sites in the two placement drivers cite `wk/calibration/RESULT.md` as the authority for withdrawing a threshold. **That file does not exist**, and `wk/` is gitignored, so the citation cannot resolve for any reader — including the three sites that print it to the operator at runtime.

## The sites

`git grep -n -i "RESULT\.md" -- . | grep -v '^wk/'` — 11 hits, both files git-tracked:

- `.claude/skills/plan-pcb-placement/scripts/placement_driver.py` — 5 (`:532`, `:803`, `:1024`, `:1110`, `:1326`)
- `.claude/skills/plan-pcb-placement-and-routing/scripts/loop_driver.py` — 6 (`:199`, `:295`, `:339`, `:2147`, `:2516`, `:2520`)

**Three are f-strings printed to the operator, not comments** — `placement_driver.py:1024`, `loop_driver.py:339` and `:2147`. A run today emits a refusal that names a file the reader cannot open.

They divide into two kinds of claim. Four cite it for *withdrawing a threshold* (`placement_driver.py:532` "why the threshold that used to live here was withdrawn"; `:1110` "`--congestion-ratio` is GONE… the calibration withdrew that refusal"; `loop_driver.py:199` "It used to gate, and the calibration withdrew that"; `:295` "The threshold that used to live here was withdrawn on measurement"). The rest cite it for the *underlying measurement* — that a perfect repair on piantor scored a negative hpwl gain.

## What survives

`wk/calibration/` exists locally and holds `_truth/`, `neo6502/`, `piantor/`, `urchin/`, `recovered.json` and `rows.json`. `rows.json` is the raw material: per-board `pristine`/`damaged` halo/hpwl/crossings/total plus `*_blocking`. So the finding is reconstructible; only the write-up is missing.

This matters more than a stale comment, because the withdrawn thresholds are load-bearing doctrine — `placement_driver.py`'s disposition gate exists *because* a scoring gate refused a perfect repair, and that is the only place the reasoning is recorded.

## Suggested fix

1. Author `docs/placement-calibration.md` from `wk/calibration/rows.json`, carrying the measurement and the two conclusions it supports.
2. Repoint the 11 citations at it. Mechanical.
[...]

Issue #622, opened 2026-08-12
Bus routing: assign slots and emit ribbons, instead of N attracted sequential searches


drandyhaas:
Today a "bus" is N independent sequential A\* searches that are softly *attracted* to a shared centerline. Nothing assigns each member a lane, and nothing ever computes the permutation between the two ends of the bus. The result is that members grab lanes greedily in routing order and the last members have to cross their own siblings — which, at fab-floor clearance inside a BGA courtyard, they cannot.

This issue proposes treating a bus as **one object**: plan a spine, assign slots from terminal order, emit members as pitch offsets of the spine, and use the **longest increasing subsequence** of the end-to-end permutation to decide exactly which members must change layer.

## Evidence: the failures are stragglers, not scattered congestion

From the `mez_rx` run (the board from #614, now added to the corpus as `set28/mez_rx` — 8 layers, 236 routable nets, 400-ball 0.8mm-pitch FPGA). Final: **91.1%** complete, 21 nets short, against a corpus median of 100%.

Grouping the 21 failures by net family:

| group | lost | size | failed members |
|---|---|---|---|
| `G*` (LED matrix) | 3 | 24 | G15, G3, G7 |
| `SDR_A*` | 1 | 12 | SDR_A9 |
| `RGMII_RXD*` | 1 | 4 | RGMII_RXD0 |
| `PORT2_T*` | 1 | 8 | PORT2_T0- |
| `R*` | 1 | 24 | R22 |

**7 of 21 failures are the last one-to-three members of a group whose siblings all routed.** (Of the rest, 5 are U1 power-rail escape remainders — a pour/comb problem, explicitly out of scope here — and the remainder are genuine singletons.)

"23 of 24 routed and the 24th can't get through" is a very specific signature. It is not congestion in general; it is a member being forced to cross the siblings that already took its lane. The router reports these as `no rippable blockers found`, which is true and completely uninformative.

## What the current implementation does, and where it stops short

Four concrete gaps:

**1. The permutation is never computed.** `bus_detection._order_nets_by_position()` sorts members by **source position only** — it picks the axis with the larger spread and sorts by it. The target order is never derived and never compared against the source order. `BusGroup.clique_endpoint` records *which end formed the clique*, and the other end is simply not modelled. So the single fact that determines a bus's whole strategy — do the two ends agree on order? — is not available anywhere in the code.

[...]

Issue #665, opened 2026-08-16
Iteration passes ship real DRC violations: cleanup/weld copper through THT pad rings (24 on the 6-floor board); conn-only wave grading hid them


drandyhaas:
**Discovery (2026-08-16, test-589 campaign): the celebrated iteration-champion boards carry REAL DRC violations that connectivity-only grading never saw.** Full re-grade of the wave ladder:

| board | conn issues | DRC violations |
|---|---|---|
| wave69/g7 (single pass, complete fanout) | 14 | **0** |
| wave72/i1-i3, wave73/c_both, wave74/a_iter1 (iterations) | 14→10→9 | **0** |
| **wave74/a_iter2 (the '8' pass)** | 8 | **28** ← introduced here |
| wave75/i6 ('6-floor' fixed point) | 6 | 24 |
| wave76/q2 (exported best board) | 6 | 24 |
| river arms (waves 82/85) | 11 | 2 |

**The 24 are one class: foreign segments through THT pad rings on inner layers** — e.g. a 13.4mm PUSB_VBUS trunk (width 0.3, In4.Cu) laid straight along the J1 pin row THROUGH J1.1/J1.2 pads (overlap 0.24mm), SD0_CMD through J2 pins on In3/In4, etc.

**Forensics so far:**
- Copper diff a_iter1→a_iter2: the trunk previously JOGGED around the J1 row (correct); a_iter2 replaced the jog with the straight — a staircase-collapse shape.
- `smooth_octolinear_chains` dry-run on the same input board (both scoped and full-board) **correctly REFUSES** the through-pad connector (its pad check works on file state) — so the in-run state at the responsible pass differed (stale `_foreign_pad_arr_cache`? in-memory pad view divergence? a different jog-collapse pass?). The a_iter2 run wrote TWICE; the oracle-recheck welds are exonerated (GND straps only).
- Additional distinct classes seen on other arms: same-net via drill hole-to-hole pairs 0.14mm apart on power nets (wave84: P3.3V, RAM_VDD — 2 violations), segment-endpoint-gap + via-segment singles on river arms.

**Reproduction (deterministic):** replay `route.py wave74/a_iter1.kicad_pcb` with the champion env (REAUDIT+LEGACY+plan attract pot65, clique/pref, layer-costs 1.0 6.0 1.0 2.5 1.0 1.0, dirs 5) — boards and exact commands in session scratchpad `gp589b/wave74.sh` (arm a_iter2).

[...]

PR #889, opened 2026-09-07
Smoother: trust the foreign-segment cache between its own splices; bounding boxes built once

drandyhaas:
## What

The octolinear smoother (`smooth_octolinear_chains`, #536) spent most of its time validating a cache that was never stale. `_foreign_seg_arrays` recomputes an O(segments) digest (sum of ids, a hash of every layer -- the #803 hardening) on **every** query, and the smoother makes one query per candidate span per helper: on a 2,500-segment board that was 23,577 queries and 16 of the pass's 19 s.

The smoother edits copper only by splicing `pcb_data.segments`, and already drops the cache at both splice sites. So:

- `_foreign_seg_arrays` skips the digest only while the caller has set `pcb_data._foreign_seg_arr_trust`; list identity, length and tail are still checked. Unset (the default), every other caller pays the #803 digest exactly as before.
- `smooth_octolinear_chains` sets the flag around its loop and resets it in a `finally`, dropping the cache on the way in and out. The loop body moves under a `try`; `git diff -w` shows the 14-line change.
- The per-item bounding boxes `_seg_foreign_seg_dist` windows on are built once per cache rebuild (`_foreign_seg_bboxes`) instead of four array operations per query.

## Evidence

- **Copper identical.** `route.py` on `kicad_files/splitflap_driver.kicad_pcb`, main vs this branch: 1,422 segments / 168 vias byte-identical, same JSON summary. The #622 braid boards (K35 1,752 segments, K41 1,790) byte-identical before/after.
- **Gain scales with segment count**: 16 -> 12.7 s of smoothing on the 2,500-segment board (the remaining cost is numpy per-call overhead over the windowed arrays); nothing measurable on the small parity board (7.4 s either way).
- `tests/test_smooth_route.py`, `tests/test_617_pcb_modification_hole_clearance.py`, `tests/test_760_hole_local_clearance.py` pass; the full suite was not run for this PR (by request).

## Why it is safe

The trust never outlives the smoother's own loop, and inside it the only copper mutation is the splice that drops the cache. An exception inside the loop clears the flag on the way out. No CLI flag, no GUI control, no behaviour change: a caller that does edit segments in place (the #803 site) never sets the flag.

https://claude.ai/code/session_01LttmRACcfGJchusgrdNduj

### DIFF
--- a/py_router/pcb_modification.py
+++ b/py_router/pcb_modification.py
@@ -4335,308 +4335,320 @@ def smooth_octolinear_chains(results, pcb_data: PCBData, scope_net_ids=None,
              'segs_removed': 0, 'segs_added': 0, 'saved_mm': 0.0,
              'chains_reverted': 0}
 
-    for net_id in sorted(segs_by_net.keys()):
-        if net_id in skip_net_ids:
-            continue
-        net_segs = segs_by_net[net_id]
-        if len(net_segs) < 2:
-            continue
-        if len(net_segs) > max_net_segs:
-            stats['nets_skipped_large'] += 1
-            continue
-        stats['nets'] += 1
-        net_pads = pcb_data.pads_by_net.get(net_id, [])
-        net_vias = vias_by_net.get(net_id, [])
-
-        candidates = [s for s in net_segs
-                      if not getattr(s, 'graphic', False)
-                      and (not keep_input_copper or id(s) in routed_seg_result)
-                      and math.hypot(s.end_x - s.start_x, s.end_y - s.start_y) > 1e-6]
-        if len(candidates) < 2:
-            continue
-
-        # Endpoint incidence over ALL same-net segments (any layer/width): an
-        # interior chain vertex must be touched by exactly its two chain
-        # segments -- a third endpoint there (other width, other layer via a
-        # barrel, a tee) makes it an anchor.
-        inc = defaultdict(int)
-        for s in net_segs:
-            inc[vk(s.start_x, s.start_y)] += 1
-            inc[vk(s.end_x, s.end_y)] += 1
-        via_pts = {vk(v.x, v.y) for v in net_vias}
-
-        def pad_reach(pad):
-            return math.hypot(pad.size_x, pad.size_y) / 2.0
-
-        def pad_on_layer(pad, layer):
-            return layer in pad.layers or any('*' in L for L in pad.layers)
-
-        groups = defaultdict(list)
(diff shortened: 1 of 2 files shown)


drandyhaas:
Merged to `main` as `aed981c5`, with one extra commit on top of the PR head (`5ed3b85b`, not visible in this PR's diff):

**`plane_pad_tap`'s window drops `_foreign_seg_arr_trust` too.** The #665 shallow-copy window clears a list of derived-geometry caches so it doesn't poison the parent, and the new flag wasn't in it — a window would have inherited a promise it never made. Inert today (the smoother builds no tap window, and the retained `id(segs)` check catches the window's rebind anyway), but that list is where a second trusting caller would have to look.

## Verification before merge

- **Copper identical**, reproduced independently on `splitflap_driver`: 1,422 segments / 168 vias, JSON summaries equal excluding timing. Compared on extracted geometry rather than whole-file, since output UUIDs are per-run random. Re-confirmed identical from the merged tree.
- **The trust turns out to be belt-and-braces rather than load-bearing.** The smoother makes no in-place mutation of any segment or via object anywhere in its 570 lines (direct assignment, `setattr`, and container mutation all checked), and its one copper edit at `pcb_modification.py:4631` *rebinds* `pcb_data.segments` — which moves `id(segs)`, still checked in the trusted signature. So the cache would invalidate even without the explicit drop. `apply_stub_layer_switch`, the in-place mutator #803 was hardened against, is not reachable from the smoother.
- **Direct test of the property, not just the outcome.** `_foreign_seg_arrays` was instrumented to recompute a digest *stricter* than #803's — folding in every segment's `start/end/width/net_id`, not only layers — on every call, and compare it against the digest recorded when that cache was built. Any trusted reuse that would have been stale gets flagged. Result on a real route: **13,313 calls, 174 rebuilds, 5,702 trusted hits, 0 stale reuses.** Non-vacuous.
- **The bbox hoist is exact.** `_seg_foreign_seg_dist` takes the arrays unfiltered and applies the `near` mask only afterward, to `fax`/`fay`/…, never to the bboxes — so building them per rebuild preserves length and index alignment. The `nid.size == 0` guard still short-circuits ahead of the lookup.
- **`tests/run_all.py`: 557 passed, 0 failed, 0 timed out, 0 skipped.**

The perf claim itself was not reproduced here — the gain lives on the ~2,500-segment braid boards, and `splitflap_driver` is too small to show it, as the PR notes.

Issue #887, opened 2026-09-06
Movie: stack an X-ray panel over an animated 3D isometric render, and drive a countdown from per-frame timing metadata


edgehero:
**Found while:** watching a placement+routing run's movie next to a KiCad 3D render of the same board and wanting one artifact instead of two.

> **Edited after filing** to correct three things the first version got wrong, two of which flattered the proposal. They are marked **[CORRECTED]** below.

## What is asked for

The movie should be a **two-panel frame, stacked vertically**:

```
+-------------------------------+
|  X-RAY: top-down, all copper  |   <- what animate_route draws today
|  layers, routing appearing    |
+-------------------------------+
|  3D ISOMETRIC render of the   |   <- does not exist anywhere in the repo
|  same board, animated         |
+-------------------------------+
```

and a **countdown timer animated into the frame** ("how long until done"),
driven by **per-frame timing recorded in the image metadata** rather than
recomputed by whoever plays it back.

The reference is a side-by-side pair (X-ray left, 3D iso right). The request is
explicitly for the **vertical** arrangement, not that horizontal one — a board
is usually wider than tall, so stacking keeps each panel at the full frame
width instead of halving it.

## What exists today

| piece | file | what it does |
|---|---|---|
| shared entry point | `py_router/make_movie.py:107` | `make_movie()`, called by the CLI, `tests/stress/render_run.py`, `run_plan.py --movie`, `place_route_loop --movie` and the GUI recorder, so they all produce the same movie (#506) |
| frame builder | `py_router/animate_route.py` | frame list from route traces / board deltas; new copper flashes white, restores green, rips red |
| the rasterizer | `py_router/route_render.py:401` | `BoardRenderer.frame(...)` — **the only drawer in the repo** |
| the camera | `py_router/movie_camera.py` | shot planning, part tweens, the side flip. Draws nothing; pure planner over `Action` records |
| the encoder | `py_router/animate_route.py:373-430` | `.mp4` via optional `imageio-ffmpeg`, falling back to a native Pillow `.gif` |
| a second composer | `py_tools/make_film.py:291` | splices boards + rejected laps (badged `TRIED`) from the ledger |

The movie is a **mandated artifact** of the combined skill —
`.claude/skills/plan-pcb-placement-and-routing/SKILL.md:175` — one of four, with
the note that a run producing only the board "is not finished".

### The three gaps

1. **One panel per frame.** Every frame is a single full-frame `BoardRenderer`
   image.

[...]

PR #943, opened 2026-09-12
Fix Pillow dependency probe: map pip name to its PIL import

Alurain:
## Problem

`deps_check.IMPORT_TESTS` has no entry for Pillow, so `_required_packages()`
falls back to `import <pip name>` (deps_check.py:74) and runs `import Pillow`.
That module does not exist — Pillow's import name is `PIL` — so the probe
reports Pillow missing on every machine, whether or not it is installed.

Pillow is the only entry in `requirements.txt` whose pip name differs from its
import name, which is why numpy/scipy/shapely are unaffected.

## Symptom

A startup dialog that installing Pillow cannot dismiss, because installing it
never changes the probe result:

```
KiCad Routing Tools needs the following Python packages that are not
bundled with KiCad:

  Pillow

Install them now into KiCad's Python? (/usr/bin/python3)
```

## Reproduce

```console
$ python3 -c "import PIL; print(PIL.__version__)"
10.2.0
$ python3 -c "import Pillow"
ModuleNotFoundError: No module named 'Pillow'
```

## Fix

Add the mapping, following the existing submodule-import convention that also
catches broken/partial installs. After the change, `_missing_packages()`
returns `[]` on a correctly provisioned system.

Verified on KiCad 10.0.6 / Ubuntu 24.04 / Python 3.12.3, Pillow 10.2.0.
`tests/gui_parity/test_manifest_plan_parity.py` and
`tests/gui_parity/test_cli_postpass_coverage.py` pass.

https://claude.ai/code/session_01AugPKrEHxVpjCYZ6923qhH

### DIFF
--- a/kicad_routing_plugin/deps_check.py
+++ b/kicad_routing_plugin/deps_check.py
@@ -32,6 +32,7 @@ IMPORT_TESTS = {
     "numpy": "import numpy",
     "scipy": "from scipy.optimize import linear_sum_assignment",
     "shapely": "from shapely.geometry import Polygon",
+    "Pillow": "from PIL import Image",
 }
 
 # Pattern matching the package name at the start of a requirements line.
--- a/kicad_routing_plugin/deps_check.py
+++ b/kicad_routing_plugin/deps_check.py
@@ -6,19 +6,27 @@ installed via the KiCad PCM (Plugin and Content Manager), no pip step runs,
 so we must detect missing dependencies on first invocation and offer to
 install them into KiCad's bundled Python.
 
-The list of required packages is read from `requirements.txt` at the plugin
-root, so requirements.txt is the single source of truth for both the CLI
-install path (install_plugin.py) and this runtime check.
+The list of packages is read from `requirements.txt` at the plugin root, so
+requirements.txt is the single source of truth for both the CLI install path
+(install_plugin.py) and this runtime check.
+
+Not every one of them BLOCKS, though (#943). `OPTIONAL_PACKAGES` names the ones
+whose absence disables a feature instead of stopping the plugin; only the rest
+gate `ensure_dependencies`.
 
 The pip install runs in a worker thread so the wx event loop keeps ticking
 and the progress dialog stays responsive (otherwise KiCad freezes for the
 full duration of the install, which can be minutes on a slow network).
+
+It is not offered at all on a PEP 668 interpreter (#944) -- see
+`_externally_managed`.
 """
 
 import os
 import re
 import subprocess
 import sys
+import sysconfig
 import threading
 
 import wx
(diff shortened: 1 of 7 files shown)


drandyhaas:
Confirmed, and pushed a follow-up commit onto this branch (`85b2568b`).

Reproduced the probe bug exactly as described (Pillow 12.0.0 installed here, `_missing_packages()` returns `['Pillow']`). One thing worth adding to the report: the consequence is bigger than a stuck dialog. `action_plugin.py:126` spells the gate `if not ensure_dependencies(...): return`, and the post-install re-probe at `deps_check.py:210` fails identically — so since `76ef8e7c` added Pillow to `requirements.txt`, the plugin could not be opened **at all**, and neither installing nor restarting changed that. Good catch.

I kept your `IMPORT_TESTS` entry — the probe should be correct — but didn't want to stop there, because a *working* gate on Pillow is still the wrong outcome. `startup_checks.check_render_dependencies` already records (#887) what it cost to make routing depend on the raster stack; `deps_check` reads `requirements.txt` wholesale, so it re-made that mistake one layer up, on the GUI. Pillow serves the "Make routing movie" checkbox and the placement preview, and both already disable themselves.

So the follow-up:

- **`OPTIONAL_PACKAGES` names Pillow.** Reported, installed alongside a blocking package (on the PCM path this dialog is the only installer there is), never a gate. A failed install of only the optional half no longer refuses the plugin either — that was a second route to the same dead end.
- **`check_render_dependencies` raises `RenderDependencyError(StartupCheckError, ImportError)`.** Both GUI consumers spell their degrade path `except ImportError`, and a `RuntimeError` walked straight past them — the movie reported "failed" instead of "not rendered, install Pillow", and the placement preview never set `_preview_ok = False`, so it re-attempted on every board.
- **`render_placement` imports the raster stack inside the five functions that draw.** It is also where `PlacementModel` and `legality_findings` live, so its module-scope gate made Pillow a requirement of `board_context.py` and the stress predictors — tools that grade a placement and draw nothing.

I also folded in #944, since it is the same dialog: `_externally_managed()` detects the PEP 668 marker (exempting venvs, whose base prefix carries it too) and names the distro packages instead of offering an install that cannot run. Direction (1) as you suggested, deliberately not (2) — `--break-system-packages` writes into a prefix the distro's package manager owns, which is the user's call to make, so the command is spelled out rather than run.

`tests/test_943_optional_render_dependency.py` covers all five, and each check was verified to kill its own bug by mutation rather than assumed to:

```
KILLED  A: drop the Pillow IMPORT_TESTS entry (the shipped bug)
KILLED  B: Pillow blocks again
KILLED  C: render gate back to a bare StartupCheckError
KILLED  D: venv guard removed
KILLED  B2: render_placement imports PIL at module scope again
```

The PEP 668 arms run against a planted marker rather than the host interpreter — reading the real one would be vacuous on any machine this repo is developed on.

One existing test needed a change: `test_run23_courtyard_channel` patched `RP.ImageDraw.Draw`, which is no longer a module attribute; it patches `PIL.ImageDraw.Draw` now, the same object resolved at call time.

Thanks for both reports — the Ubuntu/PEP 668 detail in #944 in particular is not something this repo's macOS-and-Windows testing would have surfaced.

PR #949, opened 2026-09-12
Floorplan: a declared edge connector inside its band is not off the board, and a receptacle's seat is its body, not its pad box

edgehero:
Two grader defects that together refused every automated seed of run 26 (esp_prog, placed from scratch) and forced the placement to be finished by hand through `place_pose`.

**1. The `oob_count` exemption that was promised and never implemented.** `rule_edge_connector`'s docstring says declaring an edge connector "stops `oob_count` from reporting it as a defect forever". `rule_legality` compared the raw courtyard count to `legality_budget.oob_count`, which `emit_intent` bakes from the board it is emitted from (0 on a pile). An intent declaring `CON1 east overhang 0..0.5` therefore refused all ten seeds on CON1's 0.25 mm overhang; `compare_seeds` had nothing to rank. Now `_Ctx.oob_exempt()` collects every `edge_claims()` entry with a `max` whose courtyard leaves the outline by an amount inside its band, measured with the same `rect_outside_amount(..., skip_rings=...)` call `legality_metrics` makes, and `rule_legality` subtracts them for `oob_count`, publishes `legality['oob_count_exempt']`, keeps the raw number as the optimizer's own, and names the exempt refs. Outside the band a part stays counted and the edge rule names it too; no `max` exempts nothing; `oob_amount` is left alone (the comment says why).

**2. The seat measured on the pad box.** The seat conjunct answered "does the mating face reach the edge" with the courtyard rect, i.e. the pad box on a courtyard-less library. A receptacle's pads sit inboard of its opening by construction: on the tracked esp_prog fixture USB1's drawn fab body is at x = 114.0 = the outline while its pad copper starts at 115.6, so the rule read a flush socket as "seated 1.30mm from the nearest edge with no overhang" and run 26 had to waive the brief's four USB1 clauses by name. For an `edge_receptacle` (or a brief row with `mount_mode: edge_mount`) the seat is now measured on the drawn body (`_Ctx.body_rect`, fab else silk); the finding reports `basis` and the courtyard reading beside it. The overhang conjunct and every other entry keep the courtyard (a flush body's edge-margin graze would read as a 0.55 mm overhang; a different currency, named rather than changed).

**Tests** (+64 lines in `test_549_floorplan_grade.py`, new `tests/test_run26_edge_seat_body_basis.py`): the budget test gains a second arm because ulx3s declares its 10 overhanging parts, so the old single assertion would have kept passing on `overlap_area` alone; a declared-band test (zero budget passes; band tightened to 0 -> counted AND named; no max -> counted); the seat test on the real fixture with anti-vacuity first (the body IS flush, the pads ARE inboard), a plain entry that still fires on the courtyard at 1.6 mm, `mount_mode`, an east claim on a west-flush part still failing, and synthetic no-body/silk/fab arms. Measured mutations: dropping the subtraction fails the declared arm; seating on the courtyard again fails four seat arms.

**Gates green:** test_549 (grade, cli, schema), test_712, test_713, test_702, test_902, test_711, test_923, test_run5_emit_guard, test_456, test_628, test_431_render_placement, test_504, test_794, test_548, test_895; every floorplan mutation battery's anchors match exactly once (702, 797, 799, 837, 893_916, 902, 705_792_794, 711, 936).

Named, out of scope: `emit_intent`'s plausibility note and `part_class.pose_plausible` still measure the courtyard, so an emitted intent still carries the "1.30 mm, no overhang" note for such a part.

Third of the run-26 series (#947, #948).

### DIFF
--- a/py_placer/placement/floorplan.py
+++ b/py_placer/placement/floorplan.py
@@ -1676,6 +1676,47 @@ class _Ctx:
         self._assembly_census = None
         self._bodies = None
         self._bodies_error = ''
+        self._oob_exempt = None
+
+    def oob_exempt(self) -> Dict[str, float]:
+        """`{ref: overhang_mm}` for every declared edge connector whose
+        courtyard leaves the outline by an amount INSIDE its own
+        `overhang_mm` band -- the parts `rule_edge_connector` has always said
+        are correct off the board, and that `rule_legality` used to count
+        against `legality_budget.oob_count` anyway.
+
+        Measured, run 26: an intent declaring `CON1 east overhang 0..0.5` and
+        `legality_budget {oob_count: 0}` (the emitter bakes the pile's own 0)
+        refused all ten of its seeds on CON1's 0.25 mm overhang, and the
+        placement was finished by hand.
+
+        The amount is `rect_outside_amount` with the part's OWN milled rings
+        skipped -- the exact call `QuenchState.legality_metrics` makes -- so
+        "counted" and "exempt" are decided on one number, never on two
+        readings of the outline. Only `edge_claims()` entries qualify (a
+        `connector_affinity` row claims no edge), only entries with a `max`
+        (an unbounded band would exempt any overhang whatever, which is the
+        run-10 160 mm case), and only when the part is actually off the board
+        (an interior connector is not "exempt" from anything). A part outside
+        its band is NOT exempt: it stays in the count AND `rule_edge_connector`
+        names it, two rules reporting one fact.
+        """
+        if self._oob_exempt is None:
+            out: Dict[str, float] = {}
+            for c in self.intent.edge_claims():
+                ref = c['ref']
+                part = self.parts.get(ref)
+                lim = c.get('overhang_mm') or {}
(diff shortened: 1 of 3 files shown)


edgehero:
Follow-up commit pushed: the NEAREST-EDGE conjunct now reads the same seat basis as the seat conjunct. Found by replaying the run-26 placement half from a hand-authored zone plan: every one of ten seeds failed its gate on two USB1 findings, one of them `USB1 sits nearest the south edge but is declared on the west edge` -- its pad box is 1.6 mm from the west edge and 1.3 mm from the south, while its drawn fab body is flush with the west edge. The basis is now decided once, ahead of both conjuncts, and the nearest-edge finding names it. +5 checks in the same test file (anti-vacuity on the pad box; body basis; a plain entry still reads the courtyard and misreads south; `mount_mode: edge_mount`; the east claim refused on the body). The other USB1 finding (declared `along_edge: center` within 0.6 mm, fixed pose 1.75 mm off centre) is the brief contradicting the mechanical fact, handled in a separate PR at the seed gate.

Issue #893, opened 2026-09-07
Placement: rotation as a search variable with a pin-facing cost


edgehero:
## Observed (run 25, esp_prog)

Rotation is not a variable anywhere in the placement engine. `placement/seeder.py:26-31` says it: the input rotation is tried in full first and kept when it fits, a rotation that is a decision "must be locked", and rotations are to be explored "deliberately" — by hand. The floorplan intent (`docs/floorplan-intent.md`) has no rotation field on a block, so a decision cannot even be declared; run 25 declared its three rotations by rewriting the board before seeding.

On a 31 x 14 mm stick the rotation IS the placement: which row of U1 faces USB1 decides the pair length (9.3 mm here vs 5.3 mm on the human-derived layout in run 24), whether CON2's pin order agrees with U1's (it does not: /DCOM 10.7 mm and /+3.3V 8.2 mm ended on the back, cutting the GND pour), and whether the LDO cluster sits between its source and its load. None of that is in hpwl, which is the only thing the quench moves toward.

## Proposal

1. Intent: `blocks[].rotation` (a decision, honoured exactly and locked by the seeder) and `blocks[].rotation_candidates` (a set the search may choose from; default the four orthogonal ones for 2-pad passives, the declared one for locked parts).
2. Seeder: when a block has candidates, evaluate each rotation at its zone with the same cost the quench uses, plus a **pin-facing term**: for each net leaving a pad, the distance from that pad to its nearest partner pad, summed — a rotated part whose pads face away from their partners pays for the detour before any route is tried. This is `net_affinity` per pad instead of per part.
3. Quench: rotation as a move class (90° steps for passives, the candidate set for others), gated by the same legality and the same cap logic as translation.
4. `converge.py poses --ref` already ranks (x, y, rot) triples; expose its rotation ranking per part as the seeder's candidate order.

Relation to #894: the pin-facing term is one of #894's placement terms (`pin_order_crossings`, `pair_length`, `cluster_to_pin`). It is implemented ONCE there, as a library function `board_score` and the seeder/quench both call; this issue adds the search moves and consumes that cost. #894 lands first. (The repo's rule: call the grader, never mirror it.)

Gate: this is a new placement objective term, so it goes through `tests/test_placement_ab.py` (three distinct boards, paired and directional) before it ships on; the pin-facing term must improve routed `blocking`/vias on the probe harness, not just the model it is computed from.

## Acceptance

[...]

PR #951, opened 2026-09-12
Seeder: an opt-in rotate-by-facing tie-break, measured on three boards and rejected as a default

edgehero:
**Verdict first: this flag was measured on three boards and REJECTED as a default.** It ships OFF, its rows stay in the A/B table as change detectors, and no skill text tells anyone to use it. The PR exists so the measurement is on the record rather than in a terminal.

**Mechanism.** The seeder keeps the FIRST pose that fits and its ladder starts at the part's input angle, which on a pile is a generator default -- run 26's regulator was seated with three pins 0.40 mm from the north edge that way. `place_seed --rotate-by-facing` lets every angle of the ladder find its own first fit and keeps the pose with the fewest connected pads on a row facing the outline with nothing beyond (`placement.edge_facing`, #950's geometry core); ties keep #893's author order. Flag unset: the search is byte-for-byte what it was (measured pose-identical on esp_prog, 21 parts).

**Measured** (`tests/test_placement_ab.py`, a new seed-engine arm so the same independent grading half judges it; signal `edge_facing_pads`, guards crossings/hpwl/inversions/body_blocking, declared before the run):

| row | mark | signal | guards |
|---|---|---|---|
| facing-seed-esp_prog | REGRESS | 3 -> 1 | crossings 77 -> 87, hpwl 272.9 -> 286.7, inversions 15 -> 18 |
| facing-seed-splitflap | REGRESS | 2 -> 0 | crossings 669 -> 712, inversions 190 -> 205 (hpwl better; unseated 6 -> 0, so the ON arm counts a larger seated population) |
| facing-seed-tigard | NEUTRAL | 1 -> 1 | identical poses |

Improve 0 of 3, regress 2. The count it ranks by falls on two boards and every guard rises with it: the inboard-facing angle seats where the ring search first finds room for THAT angle, not where the input angle would have sat, and every part seated after it inherits the shift. The first form (rank the ladder at the target, then seat the winner anywhere) moved the signal on no board and is recorded beside the corrected one.

**Harness.** `engine: 'seed'` rows (`_run_seed`: `seed_from_intent` -> write -> the same `floorplan.grade(..., with_health=True)` half as `_run`, plus crossings/hpwl from `pose_score`), `unseated` and `edge_facing_pads` in the baseline keys, self-test extended. Baseline re-recorded with the whole table and read row by row: the 11 existing rows keep their marks and every value; each gains the two keys.

**Tests.** `tests/test_run26_rotate_by_facing.py` pins the mechanism, not the verdict: rank 3 vs 0 at the two angles, OFF keeps the input angle and ON seats 180 at the same position, a declared edge connector ranks 0 everywhere, `seed_from_intent` reaches the tie-break end to end.

**Gates green:** test_placement_ab whole table (14/14 as measured) and `--self-test`, test_893_declared_rotation, the new test, the standing mutation-anchor preflight (50 batteries, 1108 anchors).

Fifth of the run-26 series (#947, #948, #949, #950). Depends on #950 (the `edge_facing` module).


edgehero:
Fix-up pushed after a pre-push review found a real defect: lifting the ring search into a closure left the per-angle whole-board fallback sweep outside the default (OFF) path, so splitflap's default seed reported 6 unseated parts where the pristine seeder seats all 65. The identity check had been run on esp_prog, where the sweep never runs. `_first_fit` now carries rings-then-sweep per angle for both paths; measured pose-identical to the pristine seeder on splitflap (65 parts, 0 unseated), and the unit test gained the far-target arm that would have caught it. Also in: `ignore_nets` was declared on the seed rows and read nowhere (now resolved to net ids for the crossings/hpwl columns); one `edge_facing.MIN_PADS` constant for the term, the rule and the tie-break; the prose said "every guard rises" where splitflap's crossings and wire length fall and only inversions rise. Whole table re-run and the baseline re-recorded: the 11 quench rows are unchanged in mark and every value; the three seed rows keep their marks (REGRESS, REGRESS, NEUTRAL) with GND-excluded columns. The verdict stands: rejected as a default, opt-in only.

Issue #797, opened 2026-08-29
Placement: zone_exclusive is graded and quench-enforced, but the seeder can seat a stranger straight into a reserved zone


edgehero:
`zone_exclusive` is graded, and since #702 the quench refuses a move that
intrudes on a reserved zone. The seeder does neither: it can seat a stranger
straight into one, and nothing refuses the pose.

This is the same shape as #701 one rule over. There, `keepout` was graded and
the seat search walked parts into it; the fix was to give `seeder.pose_ok` the
keep-out conjunct. `zone_exclusive` is now in the identical position:

```
$ grep -c exclusive py_placer/placement/floorplan.py
13
$ grep -c exclusive py_placer/placement/seeder.py
0
```

`seeder.pose_ok` (`seeder.py:71-111`) is containment, keep-outs, then
`candidate_valid`. `_try_place` applies the part's OWN zone through
`zone_gate`, so a member is kept inside its block. Nothing asks whether the
part is a **stranger** to some *other* block's exclusive zone, which is what
`rule_zone_exclusive` grades (`floorplan.py`, non-members only, threshold
`legality.EPS`, filtered by `z.side`).

So on a board whose intent reserves an area -- the "keep this region clear for
the RF section" case the rule exists for -- `place_seed` can seat an unrelated
part in the middle of it and then exit 4 against the intent it was built from.
That is the exact round trip #701 closed for keep-outs.

## Why #702 did not just do it

The predicate is already available and cheap: #702 added
`QuenchState.intent_spec_for`, which resolves the exclusive rects a ref must
avoid once per state (membership and the `z.side` filter are pose-invariant),
and `intent_clear` is the absolute-policy form the seat search wants. Wiring it
into `pose_ok` is a two-line change.

What stopped it is that it would change seat outcomes with **no verdict string
for the new refusal**. #701 established that a claim which strands a part is a
NAMED verdict rather than a silent "no pose", and `seeder.NO_POSE_VERDICTS` has
no member for this. A part refused by an exclusive zone would currently fall
through to `no_movable_neighbour`, whose prose --

> NOTHING seated is near enough to be in the way -- the outline, the zone or
> its own size refuses it, not a neighbour

-- is precisely the misleading answer that disclosure exists to replace. #701's
own follow-up work hit this and it is why `keepout_blocks` exists.

## What the change needs

1. The conjunct in `seeder.pose_ok`, using `intent_clear`'s exclusive slice
[...]

Issue #706, opened 2026-08-22
Placement: _seat_edge inherits the along-edge coordinate of the pose it repairs, and never rotates


edgehero:
When `place_seed --repair` re-seats a declared edge connector, it keeps whatever along-edge coordinate the part arrived with, and it never considers a rotation.

`py_placer/placement/seeder.py:840`:

```python
ax, ay = target if target is not None else (part.x, part.y)
```

The search from there is a fixed ladder, `:884`:

```python
for df in (0.0, 0.05, -0.05, ...):
```

and it returns the first legal seat (`:892`). No scoring, no tiebreak, no preference. `_edge_correct` (`:711-750`) only walks the edge normal, so `df = 0.0` means "keep the along-edge coordinate you came in with". And `:892` re-applies `part.rot` unchanged.

Measured on a real repair. A USB receptacle sat at (105.90, 100.60) before the repair and at (143.45, 100.60) after: a 37.40 mm move that is **pure x**. The 2.35 mm off-axis y, on a 14.5 mm edge, is the damaged pose's own coordinate carried verbatim into the shipped board. The centred seat at y = 98.25 was inside the legal band `_edge_frac_bounds` returns for that part, and was never considered, because the first `df` tried was legal.

The rotation half cost a hand-written script in the same run: at rot 180 no east-edge pose could orient the receptacle's overhang correctly, so the seat search reported "no conflict-free seat" on a measurably free strip.

Two suggested changes, both inside `_seat_edge`:

1. Make the along-edge start a preference rather than a fixed origin. Keep the current pose as the default, but when the intent declares a target (see the `center_on_edge` proposal filed separately) start at that fraction and keep the outward `df` ladder from there. The ladder already handles "the preferred fraction is occupied".
2. Add a rotation candidate loop for `edge_receptacle`-class parts: current rotation first, then the other three 90-degree multiples, accepting the first that converges to an on-board conflict-free seat, and naming in `notes` when a seat required a rotation.

Both preserve determinism, since the ladders are fixed and there is no RNG. Both are inert when nothing is declared: with no target the behaviour is bit-identical to today.

Gate (2) on the part having no legal seat at its current rotation, so it only ever fires where the tool currently returns failure. That avoids rotating a part whose orientation was deliberate.

PR #954, opened 2026-09-13
Seeder: stage 1 seats a declared edge connector clear of the parts already placed

edgehero:
The seeder's stage 1 is headed "edge connectors: spec geometry, no legality gate", and that is exactly what it did. It asked three questions about a seat -- on the board, clear of a keep-out, clear of an exclusive zone -- and never asked whether another part was already there.

**Measured on esp_prog**, seeded from a hand-authored zone plan: CON2, declared on the south edge with band 0.25-0.75, took the band's midpoint and drove pin 1 through the FIXED USB socket's ground tab. 0.198 mm² of pad intersection, a short. All ten seeds did it. Nothing downstream caught it either -- a pad conflict is not a budgeted channel, so the seed gate passed every seed, and `check_assembly` then called each one NOT BUILDABLE.

**The change.** The slide ladder stage 1 already runs -- scaled to the declared window, armed by #701's keep-out, #706's declared position and #797's exclusive zone -- gains a fourth arming condition and one more acceptance test: the seat must clear the parts whose pose is AUTHORITATIVE. That is `placed` at this point (locked in the file, or outside an explicit `seed_refs` scope), plus the connectors this stage has already seated, so two on one edge see each other. Never the pile, whose coordinates are meaningless.

A band with no clear seat keeps its declared edge and NAMES the blocker. Measured: dropping the connector to the later stages instead parks it in the board interior (the fixture's J1 lands at (16.4, 10.3) on a board whose south edge is y=20), trading a pad conflict for a lost edge and losing both. The conflict is left for the gate, which is where a decision between the band and the thing it lands on belongs.

**Result, run-27 replay, same board and seed:**

| | before | after |
|---|---|---|
| pad conflicts | 1 | 0 |
| `check_assembly` blocking | 1 | 0 |
| verdict | NOT BUILDABLE | buildable |

CON2 moved 0.92 mm east, still inside its own declared band. Run 26 found that same seat by hand and narrowed the declaration to 0.52-0.75 to force it.

**Tests.** `tests/test_run27_edge_seat_clears_placed.py`, four arms, each mutation-checked: the seat slides clear and stays in its band; with nothing locked the declared midpoint is unchanged (inert where no pose is authoritative); a band with no clear seat keeps its edge and names the blocker; and an entry declaring ONLY an edge -- no band, no keep-out, no exclusive zone, the case none of the three old arming conditions covers -- slides too. Removing the conflict test, the arming clause, the fallback, or widening the predicate to the pile each fails its own arm.

The fixture carries its `(layers ...)` section deliberately: without one, `copper_layers` is empty, `grade_pad_legality` resolves no shared layer for any pair, and every pad-conflict assertion passes on a grader that cannot see a conflict. The first draft had no layers section and its first arm passed vacuously; the arm that asserts a conflict IS found is what exposed it.

`tests/mutate_797.py`'s arming row is re-anchored onto the four-clause condition in the same commit, still removing only the exclusive clause, still KILLED.

**Gates green:** the new test, test_701_keepout_seating, test_706_seat_edge_target, test_797_zone_exclusive_seating, test_630_seeder_eviction, test_708, test_779, test_792 (both), test_compare_seeds, test_npth_seed_floor, test_place_seed, test_seeder_partner_centroid, test_run27_seed_gate_pinned, test_893_declared_rotation, test_712_edge_centering, test_549_floorplan_grade, test_894_placement_terms, the mutation-anchor preflight. The three seed rows of `tests/test_placement_ab.py` match the committed baseline exactly: that path passes no `seed_refs` and file-locks nothing, so `placed` is empty at stage 1 and the change is inert there.


edgehero:
Pre-push review findings fixed (follow-up commit). All three were prose describing what I meant rather than what the code does, plus one sub-micron consistency fix; no fixture's measured behaviour changes.

- `_shorted_by`'s docstring said "lands on". `pair_shortfall` measures a CLEARANCE shortfall, not contact, so a seat that merely crowds a placed part arms the slide too. That is deliberate (the seat can move along its own band, so preferring legal over not-quite-touching costs nothing) and the docstring now says it.
- the no-clear-seat note claimed the connector "is seated where it was declared". `_fallback` holds the first rung that is a legal seat, which equals the declared position only when rung 0.0 passed `edge_seat_ok`; otherwise it is the nearest legal rung to it. Corrected in both the note and the comment.
- the conflict was tested at the unrounded pose while `apply_move` writes `round(x, 3)` — half a micron against a 1e-6 threshold, invisible to any fixture, but the predicate and the seat should not be able to disagree at all.

The review also independently confirmed the parts that matter: `placed` is file-locked ∪ outside-`seed_refs` and grows as stage 1 seats each connector; `part.rot` is the declared angle at the predicate; the OFF path is bit-identical when `_fallback` is None (it reproduced the pre-change `(16.401, 10.284)` interior drop); and the cost is bounded — ulx3s 235 parts, default seed 15.45 s vs 16.60 s (inert, `placed` empty), worst case with 225 authoritative parts 7.33 s vs 6.53 s, +12%.

PR #955, opened 2026-09-13
Placement: the seeder places the residue, and P1 refuses to let it choose a connector's pose

edgehero:
The seeder is a greedy first-fit: it packs declared zones, drops everything else at its connectivity centroid, and keeps the first rotation that fits. That is the right shape for the many small parts, and it has **no representation at all for a pose that is a decision.** An edge band says which edge and not where along it; nothing anywhere says which way a mating face points. So a declared connector gets the band's midpoint at whatever angle it came in with, which on a pile is a generator default.

Measured on esp_prog (run 27, seeded from a hand-authored zone plan): both free connectors came out at rotation 0 — run 26 had decided 270° for one of them by hand and left the reason in a prose note no tool reads — and one of them put the band midpoint through a fixed socket's ground tab, on every one of ten seeds. The stack's own docstrings have called it a REFINER since long before this; what was missing was a stage that says so and a gate that holds.

**`placement_driver.py`**
- `_guard_zone_plan` gains the second half of its contract. It already demanded that every movable part be zoned, `must_lock`, or a declared edge connector; a declared edge connector must now also be LOCKED — `(locked yes)` in the board (what `place_pose lock` writes) or matched by the intent's `must_lock`. `--waive seed-connectors:<reason>` hands one to the seeder deliberately, and a reasonless waiver is refused like every other waiver here. The escape exists because a board whose connectors have a free run along their edge is a real case; it should be a decision on the record rather than a default nobody chose.
- P1's rung 2 becomes **"DECIDE, then seed the rest"**: a paragraph saying what the seeder is and is not, then (a) place and lock the parts whose pose is a decision, (b) seed the rest from the plan and rank several seeds, (c) face the rows. The census line reports how many declared connectors the seeder is choosing — on a plan that took the waiver, that is the number a reviewer should ask about.
- `_BODY_CEILING['P1']` 45 → 60 with the reason beside it. `_tiny_board` grows a `locked` parameter so the fixture can carry the file lock the refusal tells the reader to write. Two `_refusal_scenarios` rows and four `--self-test` arms, each mutation-checked: dropping the demand, accepting a reasonless waiver, and not counting a file lock as decided each fail their own arm.

**`SKILL.md`**: the unplaced ladder's rung 2 leads with what the seeder is for and what it cannot choose, and names both halves of P1's contract. **`seeder.py`**: the module docstring says it, where anyone reading the code to decide whether to call it will see it.

**Measured end to end:** run 27's own zone plan, the one that produced ten unbuildable seeds, is now refused by name — `2 declared edge connector(s) are not locked: CON1, CON2`.

**Gates green:** `placement_driver --self-test`, `--dump-all`, `--dump-refusals` (49 of 49 refusal texts rendered, 59 distinct refusals from 51 scenarios), test_run8_skill_drivers, test_431_skill_commands, test_923_output_key_claims, test_904_closeout_order, the mutation-anchor preflight.

Independent of #954, which fixes the collision itself at the seat; this one keeps the decision out of the seeder's hands in the first place.


edgehero:
Pre-push review found a real hole in this gate, fixed in a follow-up commit: it accepted an intent `must_lock` as "this connector's pose is decided", and `must_lock` does not pin a pose. It is a claim about the FILE that `place_seed` honours by stamping `(locked yes)` into its OUTPUT, after stage 1 has already seated the part. Only a file lock puts a ref in `placed` before stage 1 runs.

Measured on the seeder directly, with `must_lock: ['J1']` and no file lock: J1 goes (15, 10, 0) -> (15.0, 18.85, 0), the band midpoint at its incoming angle — byte for byte what it does when nothing is declared at all. So the gate would have passed such a plan, and SKILL.md's own "Board features that live ON the outline" sends authors down exactly that route, which made the documented path the hole.

What changed: the connector check reads the FILE lock; `locked` keeps the union only for the zone-coverage question, where must_lock legitimately means "not the seed's to arrange" (stage 1.5 seats it in place where it can). The refusal and SKILL.md both say why must_lock will not do. The census line said "placed and locked already", which overstated — a must_lock part IS re-seated — so it now separates the two and names which kind the seeder cannot move. The self-test arm that asserted must_lock suffices now asserts it does not.

Also fixed: the refusal printed `place_pose.py BOARD BOARD set <REF> <X> <Y> <ROT>`, and the real verb takes `--rot` — as printed it exited 2 with "unrecognized arguments".

Applied to run 27's real zone plan the guard now names three connectors rather than two: USB1 joins CON1 and CON2, because it carries must_lock and no file lock. On that board it only escaped the seeder because `place_seed` scoped the seed to the stacked pile, not because anything pinned it.

Issue #711, opened 2026-08-22
Placement: there is no channel for design intent; board_brief --requirements is carried verbatim and read by nothing


edgehero:
Placement infers everything about a board from the board. There is no channel through which a human or an LLM can state what the board is supposed to be, and the one channel that exists is read by nothing.

`py_tools/board_brief.py:857-861` accepts `--requirements` / `--requirements-file`, documented as "the constraints that live nowhere in the board file (enclosure, connector edges, thermal, EMI), carried VERBATIM". It is stored at `:673-678` and printed at `:829`. Grepping the tree, no other module reads that key. The same file states the design honestly at `:398-400`: "INFERENCE, NEVER AUTHORITY -- `--requirements` is the channel for fact."

It is also never invoked. `board_brief.py` appears in exactly one stage text, and only on the branch reached when a board is unplaced. On a real closed run of 153 tool invocations across 30 distinct tools, `board_brief.py` ran **zero** times.

What fills the gap instead is inference. The emitted intent is a mirror by construction, and says so at `floorplan.py:1185-1196`: "A starter intent READ OFF the board... Everything here describes what the board already is... The emitted intent grades CLEAN by construction." Each connector's edge is guessed from its current pose by proximity (`_nearest_edge`, `:1416` and `:1468`), which is the only source of an edge anywhere in the toolchain.

The cost is visible in the same run. A 7-pin 0.1 inch header spanning 48 percent of the board length, with six of seven pins 5 to 6.3 mm interior, was classed `connector_affinity` with the note "no edge claim; measured 1.56mm from the nearest edge", and the intent graded `pass: true` on 3 rules of 9. Separately, the single most consequential design decision on that board, which end of a stick-shaped board the USB connector protrudes from, was reverse-engineered at repair time from partner-pin distances.

The strongest evidence is from outside this repo. A project that uses this toolchain as its router hand-wrote three workarounds because there was nowhere to put a spec:

- a 10-entry placement dict, each line citing its requirement clause, because nothing generates a plan from a spec even though `place_fixed` is exactly the op for it;
- a length checker whose header says "the constraints that NO KRT check expresses... this is written by hand, and that it had to be is the finding";
- an intent file whose crystal block carries the note "NOTE the schema cannot express any of those three numbers".

What I think this wants, filed as a finding rather than a finished design:

[...]

PR #970, opened 2026-09-14
#960: record place_pose provenance at the promote, against the delivered path

### LINKED ISSUE #960: place_pose writes escape the unaided-regime ledger AND its refusal: staging through %TEMP% short-circuits regime_for
**Finding.** `place_pose.py` is in `provenance.LEVER_REGISTRY` (`provenance.py:54`, #892, *"the
sanctioned pose SETTER"*) and records **no provenance row** — and, the serious half, its
`UnaidedViolation` **refusal never fires either**.

## Measured (run 29, work dir staged by `stage_unaided.py`, manifest armed)

| | |
|---|---|
| `.pose-provenance.jsonl` | **120 rows**, all `place_seed.py` or `place_portfolio.py`; **0** from `place_pose.py` across 38 invocations |
| `provenance_audit.py` on the delivered board | `VERDICT: UNAIDED VIOLATION`, exit 4, **8 unclaimed**: 3 logo blocks, `CON1`, `CON2`, `Ref*`, `U1`, `USB1` |
| those 8 refs | every one has a logged `place_pose.py set` in `cmd_timing.jsonl` |
| the audit's own `levers` field | `["place_portfolio.py", "place_seed.py"]` — three registered levers moved parts, two recorded, one did not |

So the audit reports a violation **against parts seated correctly with the sanctioned tool**.

## Cause

`pose_ops.py:638` stages into `tempfile.TemporaryDirectory(prefix='place_pose_')` and calls
`write_placed_output` with that path. `record_write` then:

```python
root = regime_for(output_file)      # walks up from %TEMP%, finds no manifest
lever = active_lever()
if root is None:
    return None                     # <-- returns HERE (provenance.py:227)
if lever is None:
    raise UnaidedViolation(...)     # <-- never reached
```

`_promote()` (`pose_ops.py#939`) then `shutil.copyfile` + `os.replace`s the candidate in,
bypassing the writer. **Because the `return None` precedes the `raise`, the regime's refusal is
disarmed for any caller that stages through a temp dir** — a hand-rolled writer would be
accepted identically. `place_seed` is the positive control: it stages next to its output
(`args.output_file + '.reseat'`) and its rows are in the ledger.

## Fix

Stage the candidate next to `output_file`, as `place_seed` does — or have `_promote()` call
`record_write` after the copy.

## Acceptance

- `place_pose` rows appear in `.pose-provenance.jsonl` under an armed regime.
- `provenance_audit.py` returns CLEAN on a run whose poses were all set by registered levers.
- A writer that stages through a temp dir and promotes into an armed regime **raises**.
- The regression test covers the **promote** path, not only the writer call.

## Evidence

Reached independently by three channels before any read the others: `place_return.md` T2,
`JOURNAL.md` §11, `watch/engine_defects.md` 17 (its headline). `watch/tool_usage.md` makes it
its headline too.

edgehero:
Closes #960

`place_pose` stages its candidate in a temp dir. The writer's provenance call was made for that temp path, so `regime_for` found no manifest and `record_write` returned `None` before its refusal; `_promote` then copied the board into the armed work dir with no row. As the review suggested, the fix is now at the promote boundary: `_promote` records against `out_path` before the first copy, commits after the replace inside its existing rollback `try`, and discards the pending row in `finally` (`provenance.discard_write`, new).

**Re-scoped after review.** Everything else from the previous head is withdrawn: `placement/publication.py` (locks, journal, ledger rewrite, baseline preservation), the `ContextVar`, `decision_source`, the `regime_for` change, the audit changes, and the edits to `place_seed` / `place_reconstruct` / `place_route_loop` / `writer.py`. The evidence dump is gone; the old head stays readable at [`bae72dee`](https://github.com/edgehero/KiCadRoutingTools/tree/bae72deee7ef90a5e27618c23424c27293c88b72). Rebuilt on `main` (`a902fd7d`), 5 files.

### Measured

esp_prog staged by `tests/stress/stage_unaided.py`, then `place_pose.py board.kicad_pcb out.kicad_pcb set R1 136.4 98.8 --rot 270`:

| | `main` | this PR |
|---|---|---|
| `place_pose.py` rows in `.pose-provenance.jsonl` | 0 | 1, naming `out.kicad_pcb`, hash matches |
| `provenance_audit.py --delivered out.kicad_pcb` | UNAIDED VIOLATION, exit 4, R1 unclaimed | CLEAN, exit 0 |

### Tests

- `test_892_place_pose.py`: the armed-regime block asserted exit 0 and a file at OUT, both of which held with the ledger empty. It now asserts the row (lever, path, `refs_moved`, hash), the audit verdict with and without `--delivered`, and an in-place write. A new block promotes a temp-staged candidate three ways: undeclared (raises `UnaidedViolation`, destination byte-identical, no row), declared with a failing replace, and declared with a failing ledger append (both refused, destination restored, nothing pending). Run against `main`, the eight row and audit checks fail on the missing row; the promote-path checks stop at a `TypeError` on the new keywords there, so the mutation rows below are what show they can fail. With the fix: 152 passed, 0 failed.
- `mutate_892.py`, 5 new rows: record removed; record made for the staged path (the #960 shape); row computed after the replace; discard removed; ledger error swallowed. Battery at the commit carrying them: 32 rows, 32 killed, 0 survived, 0 broken.
- `test_411_placement_siblings` and `test_run8_write_order` are green, and the files they pin are untouched. `test_711_sibling_lists` is green; its source scan covers `pose_ops.py`, whose edit adds no sibling literal. Also green: `test_892_registries`, `test_provenance_audit`, `test_903_stagers_arm_the_regime`, `test_714_perturb_layer_flip`, `test_967_edge_floor`, `test_877_mutation_anchors`.
- Full `tests/run_all.py`, locally, on `main` (`a902fd7d`) and on this branch: 614 passed, 1 failed, 2 self-skipped on both, including `test_placement_ab.py`. The one failure is the same on both: `test_713_wallclock_census.py` flags an unregistered wall-clock site in `py_tools/repo_metrics.py`, from today's metrics commits on `main`.

### Behaviour change

An undeclared in-process `apply_poses` into an armed work dir, a lock-only call included, now raises `UnaidedViolation`, as an undeclared `write_placed_output` there already does. The only in-process caller outside tests, `place_pose.py`, declares its lever.

### Known limit, not changed here

When any row names the delivered file, `provenance_audit` takes pose claims only from rows that wrote that file. A lever that writes a NEW path from a hand-edited input therefore leaves the hand edit `unverifiable` rather than `drifted`, and the verdict is CLEAN. Measured on esp_prog: `place_optimize` moves C3 into `A`, C3 is hand-edited in `A` (audit of `A`: UNAIDED VIOLATION, drifted C3), then a declared write `A -> final` moving only R1 audits CLEAN with C3 unverifiable. That happens on `main` for a declared `write_placed_output`. With this PR, `place_pose` to a new path does the same, where on `main` its missing row made the audit fall back to the whole ledger. Closing it needs the audit to chain rows through `parent_sha256`: filed as #972.

### Not in this PR

- Pre-existing, untouched: `place_seed --repair/--reseat` delivers its temp-staged board with `write_placed_output(cur, args.output_file, [])`, so its row claims nothing; `place_route_loop --work-dir <outside the regime>` records nothing before its final `shutil.copy`; `place_seed`'s polish and re-seat passes write `<out>.polish` / `<out>.reseat` and `os.replace` them in, so their rows name the temp file and the audit's per-file pose check cannot use them. Filed as #973.
- #971 is stacked on the old head and imports `placement.publication`; it will be rebased and split on its own.


drandyhaas:
Reviewed. **#960 is a real bug and the diagnosis is exactly right** — `pose_ops.py` stages into `tempfile.TemporaryDirectory(prefix='place_pose_')`, `regime_for` walks up from `%TEMP%` and finds no manifest, and `record_write`'s `if root is None: return None` precedes the `raise`, so the refusal is disarmed for any caller that stages through a temp dir. That needs fixing.

**The patch is not mergeable as it stands: it breaks three tests, and the mechanism it introduces is much larger than the bug.**

## Measured

Full suite on Modal, whole stack tip (`f0a3cc90`): 598 passed, **16 failed**. `main` at `5a7fbcb6` has 2 pre-existing failures (`test_703_predictor_regen`, `test_run8_skills_generic`, both confirmed red on a clean `main` locally). Running every failing test against each branch in the stack attributes them:

```
TEST                                main  968  969  970  971
test_411_placement_siblings           0    0    0    1    1
test_711_sibling_lists                0    0    0    1    1
test_run8_write_order                 0    0    0    1    1
```

Reasons, on `pr970` alone:

- `test_411_placement_siblings` — `FAIL test_route_loop_round0_copy_carries_siblings: substring not found`
- `test_711_sibling_lists` — `AttributeError: module 'place_reconstruct' has no attribute '_SIBLING_EXTS'` (the module-level import was replaced by the `publish_board` delegation, and #711's "one list" gate reads that name)
- `test_run8_write_order` — `FAIL: 1 check(s): no staging board survives a completed run`

`tests/test_placement_ab.py` PASSes on this branch, so the placement objective itself is unmoved. Good.

## Two design concerns beyond the failures

**1. A killed run bricks the output path.** `publish_board` acquires `<output>.krt-publish-lock` with `os.mkdir` and only removes it in `finally` when `retain` is false. A `SIGKILL`, an OOM, or a hard `Ctrl-C` leaves the directory behind, and every later write to that path then raises `PublicationError('publication busy or recovery required: ...')`. The module docstring says this is deliberate ("subsequent writers refuse until an operator reconciles that evidence"), and for an armed benchmark regime that is defensible — but `write_placed_output` is the writer for *every* placement tool, including interactive ones, so this is now the failure mode for an ordinary interrupted `place_optimize`. If the transaction stays, please make the lock stale-detectable (pid + mtime) or scope it to an armed regime rather than to every write.

**2. `provenance._active` list → `ContextVar` changes thread inheritance.** A new `threading.Thread` starts with an *empty* context, so a lever declared on the main thread is no longer visible inside a worker. With the old module-level list it was. Nothing in `py_placer/` spawns threads today, so this is latent rather than live — but the GUI does run work off the main thread, and the failure mode is a silent `UnaidedViolation` (or, outside a regime, a silently unrecorded row). If the goal is the contention fix in `37799d98`, a `threading.local` with an explicit inherit, or keeping the list and locking around it, preserves the old semantics.

## Suggestion

Re-scope to the minimal #960 fix: record the provenance row at the **promote** boundary, where the destination path is the real one, so `regime_for` sees the manifest and the refusal fires before the copy. That is one of the two options the issue itself proposes and it fixes the disarmed gate without the lock/journal/backup/ledger-rewrite layer. The recoverable-publication work can then be judged on its own merits, with the three tests above green and `_SIBLING_EXTS` either kept or #711's gate updated deliberately.

Separately: 44,903 lines across 150 files of `docs/issue-960-evidence/` should come out before merge. #968 and #969 were merged today with their evidence stripped (the dumps stay on the PRs); `main` carries no `docs/*evidence*` convention, and on #969 the dump was not inert — five of its files were runnable scripts and `tests/test_937_tool_registry.py` failed because of them.

edgehero:
Re-scoped as you suggested, force-pushed, and the body is rewritten.

- The fix is now at the promote boundary: `_promote` records against the real `out_path` before the first copy, commits after the replace inside its rollback, and discards the pending row in `finally`. 5 files, most of it tests.
- `test_411_placement_siblings`, `test_711_sibling_lists`, `test_run8_write_order` pass again; the code the first and last pin is untouched.
- The publish lock is gone, so the stale-lock refusal you described cannot occur.
- `provenance._active` is the list again; no `ContextVar`.
- `docs/issue-960-evidence/` is gone; the old head stays readable at `bae72dee`.
- The transaction layer is withdrawn rather than folded in. If it comes back, it will be its own issue, with a stale-detectable or regime-scoped lock.

Two things surfaced while re-scoping. Both are filed, and neither is fixed here:
1. #972: the audit takes pose claims only from rows that name the delivered file, so a declared write to a new path turns an earlier hand edit into `unverifiable`, which grades CLEAN. That already happens on `main` for `write_placed_output`, and now that `place_pose` writes a row, it does the same.
2. #973: `place_seed --repair/--reseat`, its polish and re-seat writes, and `place_route_loop --work-dir` outside the regime have related delivery gaps.

PR #971, opened 2026-09-14
#961: grade a declared connector overhang on the drawn body, not a clearance shortfall

### LINKED ISSUE #961: rule_edge_connector grades declared overhang_mm with a clearance-shortfall magnitude, so a board can pass the clause BY violating the edge floor
**Finding.** `rule_edge_connector` grades the design brief's declared `overhang_mm` band against
`gate.rect_outside_amount(part.rect)` (`floorplan.py#2640`). That function measures a
**clearance shortfall**, not an overhang — so a board can satisfy a declared overhang clause
**by violating the edge-clearance floor**.

## Measured

`rect_outside_amount` against run 29's outline (`114.0, 91.0, 145.75, 105.5`):

| rect | margin 0.00 | margin 0.25 | margin 0.55 |
|---|---|---|---|
| ends **0.15 mm inside** the east edge | 0.0000 | **0.1000** | 0.4000 |
| **overhangs** the east edge by 0.50 | 0.5000 | **0.7500** | — |

It returns `margin - gap` inside the board and `overhang + margin` outside. The docstring at
`legality.py#835` (*"0 iff fully legal"*) holds only at margin 0.

**Live consequence.** `USB1`'s pad bbox ended 0.15 mm **inside** the east edge. At the board's
own `min_copper_edge_clearance` 0.25 that reads **0.10**, inside the declared
`overhang_mm {min: 0.0, max: 0.65}` — so `check_floorplan` scored the clause **satisfied**,
while `check_drc --check-pad-edge --clearance-margin 0` scored the same geometry as **2
violations** (`Pad:GND (USB1.0) too close to right board edge, Overlap: 0.100mm`,
`logs/p-lap6-pe.log`, exit 1). `check_drc` is right: the pad box is inside the board, so
physical overhang is 0.00.

**Second consequence.** The `min` conjunct (`amount < lo`, `lo = 0.0`) can never fire against a
non-negative magnitude.

**Related.** `edge_seating` rows publish `along_edge_fraction`, `along_edge_offset_mm`, `basis`,
`edge`, `span_mm` — and **no overhang field**, though the same pass grades the declared band. The
number the clause turns on is absent from its own evidence.

## Fix

Measure the conjunct on the **drawn body** against the outline as a **signed distance**, in its
own key; do not reuse the legality magnitude. Minimal mitigation if that is too large now:
compute the amount at **margin 0**, so it at least means *"outside the board"*. Add
`overhang_mm` and `overhang_basis` to each declared `edge_seating` row.

## Acceptance

- A part wholly inside the board reports overhang **0** at every margin; one overhanging by `x`
  reports `x` at every margin.
- `check_floorplan` and `check_drc --check-pad-edge` agree on the same geometry.
- The `edge_seating` row carries the number its clause was graded on.

## Note

Run 29's placement half filed this as *"the clause reads 0.0000 at every pose and cannot fire"* —
correct in direction, **understated**: the value tracks the margin and read 0.10 at lap 6, which
is what let the clause pass on a board `check_drc` was failing. The shipped board is clean on
that channel (worst gap 1.0500 mm) because `USB1` was re-seated later; the tool defect is
independent.

Found by `watch/engine_defects.md` 15; verified against source and re-derived.

edgehero:
`rule_edge_connector` graded a declared `overhang_mm` band on `rect_outside_amount(courtyard)`. That is a clearance shortfall at the gate's margin: `margin − gap` inside the board, `overhang + margin` outside. It is not an overhang. This PR grades the band on the **drawn body at zero margin**, in its own key, and changes nothing else.

Closes #961.

### Measured

esp_prog, USB1 moved `dx` mm along x from its flush pose (negative is west). Its courtyard is its pad box, 1.6 mm inboard of a Fab body that sits flush with the west edge. From `tests/test_961_body_overhang.py::test_margin_matrix`:

| dx | drawn body (graded now) | old reading @ margin .25 | old reading @ margin .55 |
|---|---|---|---|
| +0.00 | 0.0000 | 0.0000 | 0.0000 |
| −1.45 | **1.4500** | 0.1000 | 0.4000 |
| −2.10 | 2.1000 | 0.7500 | 1.0500 |
| +0.40 | 0.0000 | 0.0000 | 0.0000 |

At −1.45 the old reading satisfied `{min .05, max .20}` at margin .25 and failed it at .55, for identical geometry. The body reading fails it at both.

### What changed

- **The band's currency.** When the part's body can be measured, the band is graded on the body's overhang past the outline, summed over the sides it crosses. That is the same form as the occupancy reading it replaces, at zero margin. A body counts as measurable when the entry declares an `edge`, the outline is a rectangular Edge.Cuts, and the part draws a closed convex Fab outline out of straight segments, enclosing its own pads' centroid (an arc, a circle or a rounded `fp_rect` is not measured, and a lone pin-1 marker is not a body; SilkS is read when the part draws nothing on Fab). The geometry lives in `py_placer/placement/connector_geometry.py`, and it matches pcbnew on every measured body in the tracked corpus.
- **Everything else keeps its old reading, named.** Every call site keeps its own pre-#961 value verbatim. This covers courtyard-only parts, arcs or circles in the Fab drawing, non-rectangular outlines and entries with no `edge`. `overhang_basis` says which reading was used (`body:F.Fab`, or `legacy_occupancy@margin=0.55`). Nothing is refused, withheld or abstained.
- **The same currency where it has to agree with the rule:**
  - `_Ctx.oob_exempt`;
  - `seeder.edge_seat_ok`;
  - `_edge_correct`: it keeps the upstream walk and moves a pose on the body only when the walk's result would be refused;
  - the emitter's edged band `max`: widened to the body only when the body reads more, which never happens on the tracked corpus.
- **A band licenses the body, never copper.** The old reading carried any pad copper in front of the body, because the courtyard it measured is often the pad box itself. The body reading does not, so on the body path a declared connector whose pad copper leaves the outline (zero margin, castellated pads excepted) is a violation, and `oob_exempt` will not exempt it. Without this, a part with its pads off the board graded clean where it used to fail its band and keep its `oob_count` charge — found by review on tigard J7 moved flush with its edge. On the 22 tracked boards this adds no violation to any emitted intent, in either emit mode.
- **A marker is not a body.** A closed convex Fab drawing that does not enclose the centroid of its own pads is unmeasured: a pin-1 triangle read as the whole body graded a part over its edge as flush. All 1078 measurable envelopes on parts with pads in the corpus enclose their centroid, so no real body is refused.
- **The seat predicate sees copper too.** `seeder.edge_seat_ok` tested pad *centres*, so once the band read the body the seeder could seat a pose the grade then refused — the committed fixture puts a pad 0.75 mm past the edge under a body flush with it. It now refuses those, at zero margin, from pad boxes rebuilt in the footprint's local frame — no file reads and no `grade_pad_edge_clearance` in the candidate loop.
- **Evidence** (#961 acceptance 2 and 3). `GradeResult.edge_connector_evidence` (`--json` and the text report) has one row per declared connector, passing ones included. Each row carries:
  - `overhang_mm`, `overhang_basis`, the limit and a disposition;
  - the body's per-edge measurements;
  - `pad_copper_edge`: the part's pad-copper edge clearance from `grade_pad_edge_clearance`, at the floor `grade_pad_legality` resolves. This is evidence only, and the test checks it equals `check_drc --check-pad-edge` on the −1.45 board.

  The same keys are added to the entry's `edge_seating` row.

### Deliberately not in this PR

- **No publication gate** (review finding 1). No CLI `main()` changes; an unmeasured body only names its basis. Connector evidence in `place_seed`'s summary is issue #974.
- **No per-candidate `grade_pad_edge_clearance`, and no file reads, in the seat ladder** (finding 2). The seat does now test pad copper on every trial pose, but as containment only, off pad boxes cached per part. The clearance grading, with the board constants hoisted, is issue #975. Body envelopes are parsed once per board file content (the file is re-read and hashed per grade/state). Pad-copper clearance in the seat ladder, with the board constants hoisted, is issue #975.
- **No retired behaviours** (finding 3). These are unchanged:
  - the nearest-edge identity (run 27);
  - the `edge_receptacle` class setback and its no-overhang gate (run 4 A), which still reads the occupancy value;
  - the `connector_affinity` warning (run 23);
  - reconstruct/seeder `edge_bands` (run 4 F2).

  The repair census code is unchanged, but it charges refs from grade errors, and those are now in the body currency where a body is measured.

### What moves, and what does not

- **Emitted intents are inert.** They declare `min: 0`, and on the 22 tracked boards the emitted documents are byte-identical to `main` in both emit modes, their grades are identical across four shapes (emitted and band-widened, each at two margins), and the seeder's placements are byte-identical.
- **An intent authored against the old reading can change verdict.** Bands written as `[amt−0.05, amt+0.05]` from `main`'s own occupancy number gain `edge_connector` errors on 7 boards (splitflap and tigard flip from passing to failing). That is the issue's point: a positive minimum now fires on a body that is actually inboard.
- **The seat ladder accepts a different set of poses** (measured over a fixed pose grid: 219 lost, 1649 gained), because the band and the copper check both changed currency. End to end this is invisible: `seed_from_intent` produces byte-identical placements on both arms.

### Verification

- **Suite (Windows, local):** `upstream/main` 393eff83 gives 614 passed, 1 failed (`test_713_wallclock_census`, pre-existing: `py_tools/repo_metrics.py` unregistered). This branch gives 615 passed, 1 failed, 2 self-skipped — the same single failure, with `test_placement_ab` and `tests/test_961_body_overhang.py` both passing.
- **`tests/test_placement_ab.py`** (measured at `0b323a81`, re-run in the suite above): passes with no DRIFT, INVERTED, ORPHAN or MALFORMED line anywhere in its full log; 14 of 14 rows match both their measured verdict and the recorded baseline, which is untouched in this diff. The 9 tests the review attributed to #971 pass.
- **`test_706_seat_edge_target`** (measured at `0b323a81`): median 32.45 s → 33.06 s, **+1.9%**, inside the +5% bar (5 interleaved pairs, idle machine). Reported as measured rather than as "noise": the branch is slower in 5 of 5 pairs, so it is a real ~2% overhead — against the +36% the review measured on the withdrawn per-candidate-file-read version, which is the regression this gate exists to catch.
- **Corpus (22 tracked boards):**
  - Emitted intents are identical to `upstream/main` (plain and `declare_classes`).
  - `edge_connector` violations are identical on emitted intents.
  - Differences appear only on stress-band intents, and only on body-measured entries.
  - Of the 61 declared connectors on those boards, 23 take the body path and the rest keep the old reading (open or concave Fab outlines, arc or circle primitives, non-rectangular boundaries, and one part that draws nothing on its face). For the latter, acceptance 1 holds only on the basis the row names.
- **Cost:** a grade with edge connectors computes the copper evidence once. That added ~0.1–0.18 s on the non-rectangular boards measured in review round 1 at `7d163442` (interf_u, watchy, which use the sampled perimeter) and is noise on rectangular ones. No `floorplan.grade` caller runs it per candidate.
- **Tests:** `tests/test_961_body_overhang.py` (47 tests), plus 59 rows in `tests/mutate_902.py` that name it (80 rows in that file, against 21 before), all KILLED.
- **How far the independent verification reaches.** Five rounds, each a fresh code review, reproduction run and fact-check against measurement, every agent in its own worktree. The last full pass covers the branch through `0b323a81` and reported 14 of 15 acceptance items PASS; its one failure was a wrong number, which lived in both a doc sentence and a code comment, fixed here. The three commits after it — the certification rule for pads the edge grader cannot model, the structural selector that replaced a prose one, and two corrections — rest on this branch's own tests (47), its 59 mutation rows (all killing), the seeder-sensitive suite subset and the full suite above, not on an independent review. What they found and this branch fixes:
  - a **regression**: with the band on the body, a part whose pad copper sat off the outline passed its band and lost its `oob_count` charge (the copper conjunct above);
  - its **mirror in the seeder**: the seat predicate accepted poses the grade refused (the seat check above);
  - a pin-1 marker read as a whole body (the enclosure rule above);
  - a second-edge conjunct that was a new rule rather than a currency change, replaced by summing the body over the sides it crosses, which is the form of the reading it replaces;
  - arc-in-poly and rounded-`fp_rect` bodies measured as if they had straight edges;
  - around 40 branches that no test killed, across the rounds' four batteries (13 in one round alone), and a mutation row that killed by crashing rather than by an assertion;
  - several false or stale comments, including two that survived a round.

The old head `f0a3cc90` (publication transaction, evidence bundle) is kept on [`archive/961-connector-publication`](https://github.com/edgehero/KiCadRoutingTools/tree/archive/961-connector-publication).

https://claude.ai/code/session_01BN1dUUiHCWDwUUGmQMjnvP


drandyhaas:
Reviewed. **The core finding in #961 is correct and worth fixing**: `rect_outside_amount` returns `margin - gap` inside the board and `overhang + margin` outside, so it is a clearance-shortfall magnitude, not an overhang, and a board really can satisfy a declared `overhang_mm` band by violating the edge floor. Measuring the drawn body as a **signed** distance in its own key is the right shape, and `connector_geometry.measure`'s `body_signed_position_mm` / `body_setback_mm` / `other_body_edge_overhang_mm` are a good answer to it.

**The patch as a whole should not merge.** Four findings, in severity order.

## 1. It refuses to write output on ordinary boards

`connector_publication.run_checked` now wraps `place_seed.main` and `place_reconstruct.main`, and treats **unmeasured as fail**:

```python
accepted = not errors and not unmeasured
...
if not accepted:
    summary.update(status='refused', ...)
    print('REFUSED before publication: ...')
    rc = 4
```

Any abstention blocks publication entirely — no output file is written at all. From `test_run27_seed_gate_shorts` on this branch, a single pad is enough:

```
"unmeasured": [{"pad_ref": "FIX1.1", "reason": "unequal-axis circle uses unsupported native geometry"}]
"placed": 3, "published": false, "refusal": "declared connector requirements failed or unmeasured", "status": "refused"
```

That cascades: four other tests then die on `FileNotFoundError: .../seed-out.kicad_pcb`. `place_seed --reseat` on a damaged board exits **4** where it used to exit 0. Anything driving `place_route_loop` stops here, because the next step has no board to route.

An unmeasured conjunct is a reason to **abstain and say so** — which the rest of this codebase already does well — not a reason to withhold the board. The publication gate and the measurement are separable; the measurement is the part worth keeping.

## 2. A file-re-reading grader lands in the seeder's per-candidate path

`edge_seat_ok` now calls `candidate_copper` → `grade_pad_edge_clearance(pcb, ..., state.pcb_file)`, and that function re-reads the whole `.kicad_pcb` (`_collect_edge_cuts_segments(stream.read())`), the `.kicad_dru`, and `read_design_rules` on the `.kicad_pro` — **on every trial pose**. `edge_seat_ok` is the `on_board` predicate inside the seat ladder, called per candidate per rotation.

Measured on this branch, single-footprint candidate grade:

| board | size | per candidate |
|---|---|---|
| esp_prog | 110 KB | 0.65 ms |
| tigard | 574 KB | **6.28 ms** |
| ulx3s | 1.2 MB | **6.49 ms** |

`tests/test_706_seat_edge_target.py` went **13.78 s → 18.76 s (+36%)** on splitflap_driver, a ~100 KB board. On a 1 MB board the per-call cost is ~10× that. The outline segments and the design rules are per-board constants; they belong in the `state._connector_geometry` cache that `state_measure` already builds, not in the inner loop.

## 3. It deletes four previously-measured behaviours

`rule_edge_connector` loses the nearest-edge identity conjunct (run 27: USB1's pad box reading "nearest south, declared west" on a socket flush with the west edge), the `edge_receptacle` class-default setback (run 4 A: a receptacle 15.8 mm interior passing), and the `connector_affinity` interior WARN (run 23). Separately, `reconstruct.pad_oob_amount`, `build_candidates` and `_exchange_solve` lose the `edge_bands` allowance (run 4 F2: without it the gate tuple reverts a correct edge-homecoming, because the move raises board-wide oob by exactly the legitimate overhang).

Each of those has a measured rationale in the comment block being deleted. The "body overhang is not copper containment" argument is a good one and may well justify some of them — but #961 asked for a signed body measurement in its own key, and four separate reversals of recorded findings each need their own evidence rather than riding along.

## 4. The placement A/B baseline moved and was not re-recorded

`tests/test_placement_ab.py` FAILs on this branch (it PASSes on `pr968` and `pr970`, so this is entirely #971):

```
DRIFT intent-rp2350.off/on.edge_facing_pads: baseline 0, measured 29
DRIFT intent-ulx3s.off.edge_facing_pads:     baseline 0, measured 49
DRIFT intent-ulx3s.on.edge_facing_pads:      baseline 0, measured 47
DRIFT intent-rp2350.off/on.intent_errors_by_rule[edge_connector]: baseline 0, measured 1
DRIFT intent-ulx3s.on.crossings: baseline 2438, measured 2436
DRIFT intent-ulx3s.on.hpwl:      baseline 7554.74, measured 7553.74
```

CLAUDE.md requires a placement-engine change that moves these to re-record the baseline with `--write-baseline` in the same commit, after reading the table. Beyond the process point, this is independent evidence that engine behaviour actually moved on two real boards.

## Full-suite numbers

Modal, stack tip `f0a3cc90`: 598 passed, **16 failed**. `main` at `5a7fbcb6` has 2 pre-existing (`test_703_predictor_regen`, `test_run8_skills_generic`). Attribution by running each against every branch:

```
TEST                                main  968  969  970  971
test_630_seeder_eviction              0    0    0    0    1
test_701_keepout_seating              0    0    0    0    1
test_713_abstention_drift             0    0    0    0    1
test_797_zone_exclusive_seating       0    0    0    0    1
test_878_side_rule_sites              0    0    0    0    1
test_place_reconstruct                0    0    0    0    1
test_place_seed                       0    0    0    0    1
test_run27_seed_gate_pinned           0    0    0    0    1
test_run27_seed_gate_shorts           0    0    0    0    1
test_placement_ab                   PASS PASS   -  PASS FAIL
```

Nine new failures from this PR on top of #970's three.

## Evidence

476,710 lines across 696 files, **18.4 MB**, including a 1.18 MB binary `native-fixtures.zip` and 32 copies of `.kicad_pcb` boards. The repository's entire `docs/` tree is 9.1 MB and `.git` is 53 MB. `main` carries no `docs/*evidence*` convention. #968 and #969 were merged today with their dumps stripped (they stay readable on the PRs); please do the same here.

## Suggested split

1. `connector_geometry.py` + the signed measurement + the new `overhang_mm` / `overhang_basis` keys on `edge_seating` rows — that is #961, and it looks good.
2. The publication gate — separately, with abstain-and-report rather than withhold-the-board.
3. The per-candidate copper check in `edge_seat_ok` — separately, with the board-constant reads hoisted out of the loop.
4. Each removed conjunct — separately, with the measurement that retires the recorded finding it replaces.

edgehero:
Thanks for the review. I re-scoped this to split 1 and force-pushed. The old head is kept on `archive/961-connector-publication`. Each finding in turn:

> **1. It refuses to write output on ordinary boards**

Gone. There is no publication gate and no CLI `main()` change. A body that cannot be measured keeps its old reading and names it in `overhang_basis`; nothing is withheld or abstained. `test_run27_seed_gate_shorts` and the four tests that died on the missing output pass again. Connector evidence in `place_seed`'s summary, as abstain-and-report, is #974.

> **2. A file-re-reading grader lands in the seeder's per-candidate path**

`edge_seat_ok` no longer calls `candidate_copper`/`grade_pad_edge_clearance`. It keeps its pad-centre test and reads the body from an envelope parsed once per board file content. `test_706_seat_edge_target` is median 32.45 s -> 33.06 s, +1.9%, inside the +5% bar (5 interleaved pairs, idle machine). The pad-copper seat check with the board constants hoisted is #975, gated by a per-candidate benchmark and a commit-vs-commit A/B.

> **3. It deletes four previously-measured behaviours**

All four are back, unchanged: the nearest-edge identity, the `edge_receptacle` class setback, the `connector_affinity` interior WARN, and the `edge_bands` allowances in reconstruct and the repair census. The setback conjunct's "no overhang" gate also stays on the occupancy reading, and a test and a mutation row pin that. On the 22 tracked boards, emitted intents and `edge_connector` violations on them are identical to `main`.

> **4. The placement A/B baseline moved and was not re-recorded**

`test_placement_ab` passes with no DRIFT, INVERTED, ORPHAN or MALFORMED line anywhere in its full log; 14 of 14 rows match both their measured verdict and the recorded baseline, which is untouched in this diff. The drift came from the emitter detecting edge parts on the body instead of the occupancy census. Detection is back on the census, and emitted intents are identical to `main` on all 22 tracked boards.

Three things the review rounds added on top of the currency swap, each because the body reading loses something the occupancy reading carried:

- a declared connector whose **pad copper** leaves the outline is a violation on the body path, and is not exempt from `oob_count` (without it, tigard J7 moved flush with its edge graded clean where `main` failed it twice);
- `edge_seat_ok` refuses such a pose too, so the seeder cannot hand the grade a seat it will reject — it tested pad centres only, which the old band reading covered implicitly;
- a drawn envelope that does not enclose its own pads' centroid is **unmeasured**, so a pin-1 triangle is not read as the body.

None of them changes any emitted intent's grade on the 22 tracked boards. The seat refusal is a seeder predicate rather than a rule, and the seeded placements are byte-identical on both arms.

> **Evidence**

Stripped. The diff is 8 text files, with no `docs/*evidence*`, binaries or board copies. The measured table in the PR body is produced by `tests/test_961_body_overhang.py::test_margin_matrix`; the corpus, pose-grid, timing and suite figures come from the review rounds' own scripts and runs.

The whole suite on this branch is 615 passed with 1 failure and 2 self-skips, against 614/1 on `main` 393eff83 on the same machine (the one failure is the pre-existing `test_713_wallclock_census`).

